PR #1, opened 2024-09-09
Update readme

staab:


README.md
```diff
@@ -9,4 +9,4 @@ White Noise is a secure & private messenger client for Nostr. It's built with Ta
 1. Make sure you have `rust` and `cargo` installed.
 2. Make sure you have [bun](https://bun.sh/) installed.
 3. Run `bun install` to install the JS dependencies.
-4. Run `cargo tauri dev` to start the Tauri dev server, this will open a native app window.
\ No newline at end of file
+4. Run `bun tauri dev` to start the Tauri dev server, this will open a native app window.
```
erskingardner, line 12: Interesting. I've been running the cargo version and it runs the `bun run` command as part of it. 

I'll try this and see if there is any difference.

staab: When I tried the cargo version it didn't work 🤷‍♂️

PR #2, opened 2024-09-10
Improve loading of contacts list

erskingardner:
- Add a loading indicator while we're fetching contacts
- Show an error message if we can't fetch contacts
- Show a message if there are no contacts
- Ensure that changing profiles updates the contacts list
- Make the loader a component and allow for different sizes
- Creates a Nostr backend client (as a singleton)
- Changes contacts to use the backend client and handles updating the client and refetching contacts on identity change
- Updates lots of identity methods to handle async work (most of the rust-nostr methods are async)

### DIFF
--- a/src-tauri/src/nostr.rs
+++ b/src-tauri/src/nostr.rs
@@ -64,10 +64,6 @@ pub async fn update_signer_with_keys(keys: Keys) -> Result<()> {
         *CLIENT.lock() = Some(client);
     }
 
-    println!(
-        "Updated signer for Nostr client {:?}",
-        keys.public_key().to_hex()
-    );
     Ok(())
 }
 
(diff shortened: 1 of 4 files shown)


src/components/SidebarHeader.svelte
```diff
@@ -1,12 +1,11 @@
 <script lang="ts">
     import { NotePencil, MagnifyingGlass } from "phosphor-svelte";
-    import type { Component } from "svelte";
+    import type { ComponentType } from "svelte";
```
erskingardner, line 3: I started with this but it's deprecated apparently. Let's leave it like this for now... 

CleanShot 2024-09-11 at 08 23 39

staab: Weird, I know I read that it was deprecated in the migration guide, but maybe the new type isn't implemented yet? At any rate, I was getting type errors, which is why I changed it too.

PR #5, opened 2024-09-12
Legacy chats

erskingardner:


.gitignore
```diff
@@ -9,3 +9,4 @@ node_modules
 vite.config.js.timestamp-*
 vite.config.ts.timestamp-*
 src-tauri/target
+src-tauri/db
```
erskingardner, line 12: nearly certain i need to move this in the future but don't want to accidentally commit it.

src-tauri/src/nostr.rs
```diff
@@ -80,3 +98,65 @@ pub async fn get_contacts() -> Vec<Contact> {
...
+    let current_pubkey = PublicKey::from_hex(&pubkey).unwrap();
+    let filter = Filter::new()
+        .kind(Kind::EncryptedDirectMessage)
+        .authors(vec![current_pubkey]);
+
+    let filter2 = Filter::new()
+        .kind(Kind::EncryptedDirectMessage)
+        .pubkeys(vec![current_pubkey]);
+    
+    let events = client.get_events_of(vec![filter, filter2], EventSource::both(Some(DEFAULT_TIMEOUT))).await.unwrap();
```
erskingardner: This is the sync / blocking call that will run until EOSE or until the timeout is hit. WE'll want to replace this with a proper subscription in the background somehow.

src/components/Avatar.svelte
```diff
@@ -4,29 +4,44 @@
...
+    }
 
-    let user: NDKUser;
-    let profile: NDKUserProfile | null = null;
-    
-    onMount(async () => {
-        user = $ndk.getUser({ pubkey });
-        profile = await user.fetchProfile();
-    });
+    let { pubkey, pxSize = 32, showRing = false }: Props = $props();
```
erskingardner: @staab this is the new props and derived stuff I mentioned. It's really nice.

src-tauri/src/nostr.rs
staab: How hard could concurrency be?

src-tauri/src/nostr.rs
```diff
@@ -80,3 +98,65 @@ pub async fn get_contacts() -> Vec<Contact> {
...
+    
+    let mut chats: HashMap<String, Conversation> = HashMap::new();
+    let signer = client.signer().await.unwrap();
+    let signer_pubkey = signer.public_key().await.unwrap();
+
+    for event in events {
+        let (other_party_pubkey, decrypt_pubkey) = if event.author() == signer_pubkey {
+            let other_pubkey = PublicKey::parse(event.get_tag_content(TagKind::SingleLetter(SingleLetterTag::lowercase(Alphabet::P)))
```
staab: So much ceremony lol

src/components/Avatar.svelte
staab: I don't understand why this is an improvement over the old `export let` stuff. What's the difference?

src/routes/(app)/legacy/+page.svelte
```diff
@@ -3,19 +3,109 @@
...
+        chats = [];
+        selectedChat = undefined;
+        try {
+            const fetchedChats = (await invoke("get_legacy_chats", {
+                pubkey: $currentIdentity,
+            })) as NChat;
+            const sortedChats = Object.entries(fetchedChats)
+                .sort(([, a], [, b]) => b.latest - a.latest)
+                .reduce((acc, [key, value]) => ({ ...acc, [key]: value }), {});
+            chats = sortedChats as NChat[];
```
staab: All the type coercion should be avoidable. Maybe wrapping the invoke call in a `fetchChats(pubkey): Promise<NChat[]>` function would work.

staab (approved):

src-tauri/src/nostr.rs
erskingardner: Seriously. This took me like an hour to work out. I was furious. 😂

src/components/Avatar.svelte
erskingardner: The props like this are automatically reactive and deeply reactive so you can pass objects through if you want as well as simple values. It's the combination of that plus the derived values below that means loads of stuff just stays in sync.

src/routes/(app)/legacy/+page.svelte
erskingardner: Yeah, this is very ugly and definitely super optimizable. I wanted to just get something basic going to see it working, but this ultimately will be replaced with the subscription based logic so I didn't spend much time optimizing.

src/components/Avatar.svelte
staab: Interesting, I don't feel like I ever had that problem. Is it because $state variables proxy array methods like `push`? I tried using push the other day but it didn't trigger reactivity.

erskingardner: I feel like fetching profile data especially has always been problematic for me. Might just be the way I'm using things in variables instead of const functions like you prefer. The state stuff is supposed to track changes to arrays, objects, etc. now. https://youtu.be/tErKyuUTzsM?si=1sVVoAmqPRrjqrss&t=611

PR #6, opened 2024-09-14
Small adj. to `nostr` module

yukibtc:
Hi, I've done some small adj. to the `nostr` module and removed some unnecessary deps.


erskingardner:
Thanks @yukibtc ! I'm still so new to Rust and have played with how to set this up a few times. I'm actually in the midst of changing things again so this might not get merged but the changes are super informative and helpful for me to see! 🙏

yukibtc:
> Thanks @yukibtc ! I'm still so new to Rust and have played with how to set this up a few times. I'm actually in the midst of changing things again so this might not get merged but the changes are super informative and helpful for me to see! 🙏

Feel free to ping me if you need help for the nostr-sdk, the documentation is pretty bad at the moment.

erskingardner:
Thanks @yukibtc - quick question, the timeouts that I see around a few methods. In the case that I'm using a database backend (nostrdb) the await seems to take the full timeout even if there are already values being returned by the database. Is that right? Is there a better way that I should be trying to quickly & incrementally load data from the cache but ensure that we're also fetching from relays at the same time?

yukibtc:
> Thanks @yukibtc - quick question, the timeouts that I see around a few methods. In the case that I'm using a database backend (nostrdb) the await seems to take the full timeout even if there are already values being returned by the database. Is that right? Is there a better way that I should be trying to quickly & incrementally load data from the cache but ensure that we're also fetching from relays at the same time? 

The timeouts are used only when fetching data from relays. If it wait the full timeout probably it's due to a very slow relay or a huge query.
In rust-nostr the disconnected relays are automatically skipped after some attempts so this shouldn't be the issue.

I think you can use the `subscribe` method and query directly from database.
The `subscribe` method send a `REQ` and keep it up (the `get_events_of` is an auto-closing subscription, close the `REQ` on `EOSE`). All events received are automatically saved into the database.

```rust
// Subscribe
client.subscribe(filters, None).await?;

// ...

// Query from database
let database = client.database():
let events = database.query(filters).await?;
```

PR #71, opened 2025-02-11
Feature/linux build

dmcarrington:
Added scripts and instructions for Linux builds. Tauri builds and dependencies vary greatly between distros, so done this using a Dockerfile to get a stable build environment.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ node_modules
 vite.config.js.timestamp-*
 vite.config.ts.timestamp-*
 upload-keystore.jks
+package-lock.json
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,43 @@
+FROM ubuntu:22.04
+
+# Install necessary dependencies
+RUN apt-get update && apt-get install -y \
+    curl \
+    wget \
+    build-essential \
+    libwebkit2gtk-4.0-dev \
+    libssl-dev \
+    libgtk-3-dev \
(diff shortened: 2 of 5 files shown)


erskingardner:
@dmcarrington Thanks for the PR! I left a few comments there, would love to use Bun instead of NPM so that we're consistent everywhere.

dmcarrington:
@erskingardner I've replaced npm with bun where possible. I don't see any further comments, did you possibly not save the review?

erskingardner:
> @erskingardner I've replaced npm with bun where possible. I don't see any further comments, did you possibly not save the review?

Yup - didn't save the review 🤦‍♂️

erskingardner:
Thanks @dmcarrington - just a quick question - this will always build the app on ubuntu but the builds will function for any flavor of linux, correct?

Also - out of curiosity, does running `bun tauri dev` work ok on linux?

dmcarrington:
The `.deb`  package will work on Ubuntu (and other Debian based distros). The `.rpm` package will work for Fedora and other Redhat derived distros. The `Appimage` should work on everything. Package management on Linux is much more diverse than on Apple and there isn't really a "universal" solution.
Haven't checked `bun tauri dev` yet, may need to run that through another Dockerfile for consistency. I'm using Fedora atm, and getting tauri to build natively is not exactly straightforward :confused:

erskingardner (approved):

PR #73, opened 2025-02-17
Default relay is localhost:8080 in dev

justinmoon:
The localhost:8080 thing can be confusing to new dev, this helps.

Are we using any linter / fixer on the typescript? These lines look pretty ugly to me lol but I always just depend on tooling to format.


erskingardner (approved):
Great catch.

justinmoon:
> Yes. I'm using Biomejs

Cool, will install

PR #74, opened 2025-02-17
TAURI_DEV_PORT env var

justinmoon:
This allows devs to run 2 instances of the app at once by setting different TAURI_DEV_PORT on each. This is useful when trying to chat back and forth between 2 clients when testing a new feature.

If you don't specify anything then it still defaults to 1420.

### DIFF
--- a/src-tauri/tauri.conf.json
+++ b/src-tauri/tauri.conf.json
@@ -5,7 +5,6 @@
     "identifier": "com.paper-robin.whitenoise",
     "build": {
         "beforeDevCommand": "bun run dev",
-        "devUrl": "http://localhost:1420",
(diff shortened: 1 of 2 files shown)


src-tauri/tauri.conf.json
```diff
@@ -5,7 +5,6 @@
     "identifier": "com.paper-robin.whitenoise",
     "build": {
         "beforeDevCommand": "bun run dev",
-        "devUrl": "http://localhost:1420",
```
justinmoon, line 8: If this is still in there you can't run on any other port

erskingardner: Have you verified that builds will still function with this removed?

PR #79, opened 2025-02-21
Basic docker-compose.yml that runs relay and blossom server

justinmoon:
Copied in example configs for both servers -- this can be nice to understand what options are available.


jgmontoya:
Related: https://github.com/jgmontoya/whitenoise/pull/1

PR #84, opened 2025-02-21
Only set debug window title on iOS

justinmoon:
Without this iOS build fails. I confirmed that iOS build no longer fails after this change.


justinmoon:
Not needed with https://github.com/erskingardner/whitenoise/pull/85

PR #94, opened 2025-02-22
feat(payments chat[id]+page RepliedTo): make invoice payment show as special reply

jgmontoya:
Makes invoice payments replies rather than reacts

image


a-mpch:
tAck 🌮

jgmontoya:
Closed in favor of #95

PR #97, opened 2025-02-25
feat(commands/groups chats/[id]/+page): add delete message functionality

jgmontoya:
# 🗑️ Message Deletion Feature

This PR adds the ability for users to delete their own messages in chats!

## Features

* Delete your own messages with an additional button on the message menu
* See a simple "Message deleted" indicator where deleted messages used to be (styled in a way that makes it obvious it's a "system message" so a user can't fake a deletion)
* Checks to ensure only message authors can delete their messages
* I made it so "Invoice paid" messages are not deletable (from the ui)

https://github.com/user-attachments/assets/79cd3e3d-6d0c-4275-88c7-b3605a26bbca

## Questions

Currently, messages are not being really deleted from the clients database. Actually deleting them would be possible, but could make it weird in the ui (having a message actually disappear). Perhaps a reasonable compromise could be to delete the message content, but leave the event_id and created_at. We could then easily render these 'empty' messages as deleted. It's a bit odd to mutate events in this way though, so I decided to just ask @erskingardner for what the recommended approach would be.
On the meantime, this is already usable as-is.


jgmontoya:
We're better than Signal now(?) 😂


https://github.com/user-attachments/assets/734698db-2874-4029-ac5f-4da5905a2411

erskingardner:
Amazing work @jgmontoya! 🙏

I think for now we can leave the message content alone in the database. I can't think of a really good reason to delete it. 

Just FYI. When we implement disappearing messages we'll be removing the entire event/message object from the database based on the disappearing policy.

src/routes/(app)/chats/[id]/+page.svelte
```diff
@@ -492,15 +566,19 @@ onDestroy(() => {
...
                             {#if doesMessageHaveQTag(message)}
-                                <RepliedTo messageId={findQTagReplyTo(message)} />
+                                <RepliedTo messageId={findQTagReplyTo(message)} messages={messages} />
                             {/if}
                             <div class="flex {message.content.trim().length < 50 && !isSingleEmoji(message.content) ? "flex-row gap-6" : "flex-col gap-2"} w-full {doesMessageHavePreimageTag(message) ? "items-center justify-center" : "items-end"}  {isSingleEmoji(message.content) ? 'mb-4 my-6' : ''}">
                                 <div class="break-words-smart w-full {doesMessageHavePreimageTag(message) ? 'flex justify-center' : ''} {isSingleEmoji(message.content) ? 'text-7xl leading-none' : ''}">
-                                    {#if message.content.trim().length > 0}
+                                    {#if isMessageDeleted(message)}
+                                        <div class="inline-flex flex-row items-center gap-2 bg-gray-200 rounded-full px-3 py-1 w-fit text-black">
+                                            <span>🗑️</span><span class="italic opacity-60">Message deleted</span>
```
Comment (erskingardner) [inline src/routes/(app)/chats/[id]/+page.svelte]: This is a nitpick but can we use the phosphor `Trash` icon instead of an emoji here?

src/lib/components/RepliedTo.svelte
```diff
@@ -29,7 +37,13 @@ onMount(() => {
                     <Name pubkey={message.pubkey} unstyled={true} />
                 </span>
             {/if}
-        <span class="break-words-smart">{message.content}</span>
+        {#if isDeleted}
+            <div class="inline-flex flex-row items-center gap-2 bg-gray-200 rounded-full px-3 py-1 w-fit text-black">
+                <span>🗑️</span><span class="italic opacity-60">Message deleted</span>
```
erskingardner: Same with the emoji here, change it for the icon.

src/lib/components/MessageBar.svelte
```diff
@@ -101,7 +113,13 @@ onMount(() => {
 <div class="messagebar sticky bottom-0 left-0 right-0 bg-gray-900 drop-shadow-message-bar">
     {#if replyToMessageEvent}
         <div class="w-full py-4 px-6 pl-8 bg-blue-700/50 text-white backdrop-blur-sm border-t border-gray-700 border-l-4 border-l-blue-500 flex flex-row gap-2 items-start justify-between rounded-t-xl">
-            <span>{replyToMessageEvent.content}</span>
+            {#if isReplyToMessageDeleted}
+                <div class="inline-flex flex-row items-center gap-2 bg-gray-200 rounded-full px-3 py-1 w-fit text-black">
+                    <span>🗑️</span><span class="italic opacity-60">Message deleted</span>
```
erskingardner: Same again - the icon

erskingardner (changes requested):
Couple of super small nitpicks. Otherwise looks ready to merge!

jgmontoya:
> I think for now we can leave the message content alone in the database. I can't think of a really good reason to delete it.
> 
> Just FYI. When we implement disappearing messages we'll be removing the entire event/message object from the database based on the disappearing policy.

I guess the main reason would be the expectation that retrieving the content from deleted messages shouldn't be possible. I think the communication being over nostr makes that kind of expectation hard to fulfill on the first place but we can still maybe come closer?

src/routes/(app)/chats/[id]/+page.svelte
Comment (jgmontoya) [reply src/routes/(app)/chats/[id]/+page.svelte]: done

src/lib/components/RepliedTo.svelte
jgmontoya: done

src/lib/components/MessageBar.svelte
jgmontoya: done

erskingardner (approved):

Issue #62, opened 2025-02-04
Onboarding process should be improved


vladimir-krstic:
The onboarding process should better explain Inbox relays, Key Package relays, and Key Packages. Moreover, how White Noise works should be improved in general.

alltheseas:
Agreed. I deem the current setup is for advanced nostr users only. 

Steps 1 - 3 through should be on auto-pilot for new users @erskingardner. 

If you want to configure advanced inbox relays and similar configs, this should be an advanced setup option.

erskingardner:
yup - I agree. was just doing a bit of user testing with a UX designer and she had the same feedback too. 

We can come up with something better. 

1. If we detect that you don't have any of the required things set up then we give you a single button to click and we do everything for you.
2. We have a button to expand the settings of what we're going to do so that power users can change things.

alltheseas:
> If we detect that you don't have any of the required things set up then we give you a single button to click and we do everything for you.

For the flow I tested, I created new keys. The above pre-condition is true initially for all new keys generated in app.

erskingardner:
Great point. As part of creating a new account, we should do all those things for you.

alltheseas:
YESTR

erskingardner:
For new users we're now doing all the things for you. I'm going to close this and open another new issue for how we manage the onboarding process for users that sign in to WN with an existing account.

PR #100, opened 2025-02-27
Update onboarding and Profile settings

### LINKED ISSUE #58: Account switching and Settings optimization
- Few people use multiple accounts, and of those that do, a small portion requires switching on the fly between accounts. UI will be much cleaner  with account switching in a dropdown. 
- Almost all settings are tied to a single account so each account should have independent Settings page. Icon for settings can be tied to the account in a account switch modal. 
- This would require to have manage and add accounts in the same modal.

Image

### LINKED ISSUE #61: No nsec when creating new account
When user creates new nostr account it doesn't get nsec nor it can back it up later. 

User should be instructed how nostr key pair works and have the option to backup nsec.

### LINKED ISSUE #63: 💁 Edit Profile Metadata
Users need the option to change nostr profile metadata right in the app without relying on the other nostr clients. At least profile picture and name for the start.

erskingardner:
This PR covers a few changes: 

1. For brand new nostr identities (create new account/keypair) we automatically complete the onboarding steps since we know the user won't have any events published already. (related to #62)
2. Redesign the account switcher on the settings page. (Fixes #58)
3. Add profile and keys settings sub-pages. 
4. Profile page allows users to view and update their profiles (Fixes #63)
5. Keys page allows users to view and export their public and private keys (Further fixes #61)
6. Network page now allows users to manage their relays (nostr (nip-65), inbox, and key package)


src/routes/(app)/settings/profile/+page.svelte
```diff
@@ -0,0 +1,371 @@
...
+            picture: picture,
+        };
+
+        // Publish metadata update
+        await invoke("publish_metadata", { metadata: newMetadata });
+        toastState.add("Success", "Profile updated successfully", "success");
+    } catch (error) {
+        console.error("Error updating profile:", error);
+        toastState.add("Error", `Failed to update profile: ${error}`, "error");
+    }
```
a-mpch, line 230: nit, i would scroll up to see the feedback

a-mpch (commented):
tACK, tested branch

I had an error when setting up an account from "main onboarding". The display name is not persisted somehow, showing the npub.
If you add an account through the modal works.

Debugging the error, once I received `Error publishing metadata event for new user:` but couldn't replicate it.
The other times that I've tried metadata was published but still not persisted in the account.

Maybe a race condition(?)


Not functional but of design, all text on inputs are white and inputs are white 🤷🏼, I can't read whatever it is changed in all changed inputs

a-mpch:
just for references images of the inputs
image

src-tauri/src/commands/accounts.rs
```diff
@@ -41,7 +43,33 @@ pub async fn create_identity(
...
+    publish_relay_list(
+        vec![
+            "wss://relay.damus.io".to_string(),
+            "wss://relay.primal.net".to_string(),
+            "wss://nos.lol".to_string(),
+        ],
+        10051,
+        wn.clone(),
+    )
+    .await?;
```
jgmontoya: Maybe instead of having these relays hard-coded we can read them off a file? That way we can later on modify the default relays without needing to make changes in the code.

src-tauri/src/commands/accounts.rs
```diff
@@ -41,7 +43,33 @@ pub async fn create_identity(
...
+    publish_relay_list(vec!["wss://auth.nostr1.com".to_string()], 10050, wn.clone()).await?;
+
+    // Publish a key package relay list
+    publish_relay_list(
+        vec![
+            "wss://relay.damus.io".to_string(),
+            "wss://relay.primal.net".to_string(),
+            "wss://nos.lol".to_string(),
+        ],
+        10051,
```
jgmontoya: What is kind 10051? Couldn't find the nip for it 😬

src-tauri/src/accounts.rs
```diff
@@ -110,6 +115,19 @@ impl Account {
...
+
+        match event_result {
+            Ok(event) => {
+                tracing::debug!(target: "whitenoise::accounts", "Published metadata event for new user: {}", event.id().to_hex());
+            }
+            Err(e) => {
+                tracing::error!(target: "whitenoise::accounts", "Error publishing metadata event for new user: {}", e);
+            }
+        }
+
```
jgmontoya, line 130: Nit: extract to a method

src-tauri/Cargo.toml
```diff
@@ -18,6 +18,7 @@ crate-type = ["staticlib", "cdylib", "rlib"]
 tauri-build = { version = "2", features = [] }
 
 [dependencies]
+bitcoin_hashes = { version = "0.14", default-features = false, features = ["serde"] }
```
jgmontoya, line 21: Latest version is `0.16.0`, is there any reason not to use that one?

src-tauri/src/commands/nostr.rs
```diff
@@ -602,3 +604,68 @@ pub async fn export_nsec(
...
+
+    tracing::debug!(
+        target: "whitenoise::commands::nostr::generate_nip98_auth_token",
+        "Generated auth token for URL: {}, method: {}, event: {:?}",
+        url,
+        method,
+        event
+    );
+
+    Ok(event.as_json())
```
jgmontoya, line 682: This part is not using the state directly (`wn`) so maybe we could extract it to a new method that could be tested

src/routes/(app)/settings/profile/+page.svelte
```diff
@@ -23,24 +37,197 @@ $effect(() => {
...
+            lud16 = metadata.lud16 ?? "";
+            banner = metadata.banner ?? "";
+            picture = metadata.picture ?? "";
         });
     }
 });
 
+$inspect(metadata);
+
+async function uploadImageToNip96(file: File): Promise<string> {
```
jgmontoya, line 54: I think this function is too large, and also doesn't seem to be directly related to the component. Is there any reason why this should be handled by the client instead of in rust? Regardless of that, I'd move it to something like a "Nip96Client" class with an "upload_image" method. What do you think?

src/routes/(app)/settings/profile/+page.svelte
```diff
@@ -23,24 +37,197 @@ $effect(() => {
...
 async function handleImageUpload(event: Event, field: "picture" | "banner") {
     const input = event.target as HTMLInputElement;
     if (input.files?.[0]) {
         const file = input.files[0];
-        // TODO: Implement image upload to a hosting service
-        // For now, we'll just use a local URL
-        metadata[field] = URL.createObjectURL(file);
+        try {
+            // Set loading state
+            if (field === "banner") {
```
jgmontoya, line 174: Nit: Perhaps define

```ts
const isBanner = field === "banner";
```

and then check `isBanner` rather than making the string comparison each time

jgmontoya (commented):
Nice!! Leaving some comments :D

src-tauri/src/commands/accounts.rs
jgmontoya: oh nvm I see it's KeyPackage Relays Event from nip 104

src/routes/(app)/settings/profile/+page.svelte
erskingardner: You probably caught it in a broken state _ this wasn't ready to review, I just keep pushing when I get to a good checkpoint.

src-tauri/Cargo.toml
erskingardner: yes, it's on v.14 in nostr-sdk and if we use a different version we get errors.

PR #101, opened 2025-02-28
Refactor: split groups commands

jgmontoya:
# Refactor Groups Commands Module

## What’s This PR About?

This PR breaks up the commands in `groups.rs` into new groups module, with each command getting its own file. 
Why? It’s simpler to navigate (just pick a file named after the command) and makes dependencies super clear, now you can see exactly what each command relies on, including other commands.

## Why it’s better

**Cleaner Code:** Splits a chunky 1.1k+ line file into bite-sized pieces, making everything easier to manage and aligning with the single responsibility principle.

**Easier Maintenance:** Smaller files mean quicker understanding, simpler reviews, and less risk of breaking stuff when you tweak one command.

**Easier Testing:** Each command can now (more easily) be tested on its own (check out `delete_message`).

**Open-close principle:** Adding new commands means just creating a new file and exporting it. It’s predictable and doesn't require changes on previously existing files/commands.

The second commit shows off the perks by refactoring `delete_message` validation logic. It’s now more testable and less tied to global state. This setup, besides organizing code, naturally leads to better design patterns throughout the codebase.


erskingardner (approved):
🔥

PR #103, opened 2025-02-28
Refactor/split commands

jgmontoya:
Follows the refactor proposed in https://github.com/erskingardner/whitenoise/pull/101 for other commands.


src-tauri/src/commands/payments/pay_invoice.rs
```diff
@@ -0,0 +1,262 @@
...
+        wn,
+        app_handle,
+    )
+    .await
+    .map_err(|_| CommandError::MessageError)?;
+
+    Ok(unsigned_message)
+}
+
+struct MlsMessageParams {
```
erskingardner, line 63: This stuff should probably be moved out of the command and into another file, what do you think?

jgmontoya: I agree, although maybe we can hold on until we find another candidate for reusing it. For the moment it's just internal to the functioning of this particular command

PR #105, opened 2025-03-01
Refactor (onboarding updates): split accounts and nostr commands

jgmontoya:
Splits the accounts and nostr commands like we did in #101 

This is an alternative to #104, but pointing to the onboarding-updates branch to avoid conflicts.


jgmontoya:
> I'm guessing that I don't need to merge this now that I've merged 104 right @jgmontoya? I am happy to deal with merge conflicts when I hit them.

Correct, closing!

Issue #20, opened 2025-01-11
⬆️ Media Uploads


erskingardner:
This has to be carefully implemented. Media is a vector used by many exploits. 

A few initial thoughts: 

- Blossom should be used if at all possible.
- Pre-rendering/processing of media (while it breaks e2ee) is how the big tech cos handle keeping user devices safe.
- We want to initially handle images, videos, and pdfs.

![CleanShot 2025-01-11 at 13 21 37@2x](https://github.com/user-attachments/assets/87fa60d9-b9b4-4984-99d5-9460cb651ce9)

erskingardner:
BTW, I'm totally open to the fact that this might be overkill for now and we should just upload the full file, encrypted to blossom and use the normal NIP-92 `imeta` tag to give the client some info while we're downloading the full file in one shot. This would likely work great for audio, images, and docs. And only break down when it comes to video.

josefinalliende:
I really like this idea—it seems like a smart solution. That said, I’m not familiar with this NIPs or Blossom, so feel free to correct me if I’m missing something.

My main question is: Would starting with full file uploads make it harder to transition to a chunked structure later? I think that uploading the whole file - after compressing it on the client side by default—would be easier to implement and would likely still be sufficient for users. Later, we could introduce an HD feature—similar to what WhatsApp does—where users can explicitly choose to upload files in full quality, and at that point, we could use chunked uploads.

If starting with full file uploads would make switching to a chunk-based structure more difficult in the future, then maybe we could implement this solution from the start, but initially using just one chunk that contains the entire file (with reduced quality on the client side) and later we could extend it to support actual chunking for larger files. Does an approach like this make sense? Or, if we already have the chunking structure, the effort to process multiple chunks instead of one would be small?

jgmontoya:
When in doubt, I'd lean toward simplicity.
The chunked approach feels a bit overly custom and complex when range requests are a standard option for partial downloads. 

I like the idea of having a metadata-first message, it tells us what we’re downloading and adds useful info. Could we skip the .json step though? I think we could just add the required information directly on the first message if we're not doing chunking on our end. It'd cut a layer of indirection. Not sure if we'd be losing too much flexibility though, thoughts?

Regarding this:

> This would likely work great for audio, images, and docs. And only break down when it comes to video.

Why does it break for videos? Is it about streaming, wanting playback before the full download?

[...]

PR #108, opened 2025-03-13
Media Uploads

erskingardner:
This implements #20 

- [x] Create media module to handle encrypting, building imeta tags, and uploading.
- [x] Create blossom module to handle upload and download of raw data to blossom servers.
- [x] Create basic uploading UI showing preview of images, allowing removal of images before message send.
- [x] Create `media_files` table to store data about media uploads (including group and message ids, and one-time nostr key used to upload to blossom)
- [x] Basic file upload sanitizer
- [ ] Delete files from cache and blossom when they're removed from the new message UI
- [ ] Create UI to replace URL of media in messages with actual media files. (image gallery, videos, file icons)

## Summary by CodeRabbit

- **New Features**
  - Enabled file attachments and previews in chat, improving media sharing.
  - Upgraded media handling with secure uploads, metadata sanitization, and caching.
  - Enhanced desktop interactions via updated file system and dialog capabilities.
  - Updated the application homepage to [https://parres.org](https://parres.org).
  - Adjusted upload security settings to require authentication while disabling unused endpoints.
  - Introduced functionality for uploading media files alongside messages in the chat.
  - Expanded relay connectivity during development with additional WebSocket URLs.
  - Introduced a new module for handling media operations, including file uploads and encryption.
  - Enhanced the `MessageBar` component for improved file upload management.

- **Chores & Enhancements**
  - Increased CI build timeout and refined deployment configurations.
  - Introduced improved error handling and logging in event processing.
  - Enhanced Android build process with additional cleanup and configuration steps.
  - Added new workflows for building Android split and universal binaries.


erskingardner:
@jgmontoya @a-mpch I've implemented quite a lot of the changes to add the metadata table and local cache. I'm expecting this build to fail (and the front-end won't work yet) but I'd love your quick thoughts on the changes to make media it's own big module and break apart all the sub-jobs into their own sub modules.

src-tauri/db_migrations/0002_add_media_files.sql
```diff
@@ -0,0 +1,16 @@
+CREATE TABLE media_files (
+    id INTEGER PRIMARY KEY AUTOINCREMENT,
+    mls_group_id BLOB NOT NULL,
+    file_path TEXT NOT NULL,
+    blossom_url TEXT,
+    file_hash TEXT NOT NULL,
+    nostr_key TEXT,
+    created_at INTEGER NOT NULL,
+    file_metadata TEXT,  -- JSON string for file metadata
```
jgmontoya: If we want to use regular json this is fine, if we want jsonb I think this should be BLOB

src-tauri/db_migrations/0002_add_media_files.sql
```diff
@@ -0,0 +1,16 @@
+CREATE TABLE media_files (
```
jgmontoya, line 1: Should we have a reference to the message (message_id)?

src-tauri/db_migrations/0002_add_media_files.sql
```diff
@@ -0,0 +1,16 @@
...
+    id INTEGER PRIMARY KEY AUTOINCREMENT,
+    mls_group_id BLOB NOT NULL,
+    file_path TEXT NOT NULL,
```
jgmontoya, line 5: Instead of making it `NOT NULL` we could use null values to signify "not yet downloaded" files

src-tauri/src/commands/groups/send_mls_message.rs
```diff
@@ -50,6 +44,77 @@ pub async fn send_mls_message(
 
     let export_nostr_keys = Keys::parse(&export_secret_hex).map_err(|e| e.to_string())?;
 
+    // Process media files if present
```
jgmontoya, line 52: We should try to extract this to functions as much as possible as it's already too large

src-tauri/src/media/mod.rs
```diff
@@ -0,0 +1,307 @@
...
+/// * `exporter_secret_hex` - The 32-byte exporter secret in hex format, used for encryption
+/// * `data_dir` - The directory to cache the file in
+/// * `db` - The database connection for storing file metadata
+/// * `blossom_client` - The client for interacting with the Blossom server
+///
+/// # Returns
+///
+/// * `Ok(UploadedMedia)` - The uploaded media descriptor and IMETA tag
+/// * `Err(MediaError)` - Error if any step of the process fails
+pub async fn add_media_file(
```
jgmontoya, line 76: I don't think we should have any of these functions directly on the mod.rs file, I'd much rather them be on their own separate files

jgmontoya (commented):
Some comments!

src-tauri/src/accounts.rs
```diff
@@ -686,17 +686,28 @@ impl Account {
     }
 
     /// Stores a Nostr Wallet Connect URI for this account
-    pub fn store_nostr_wallet_connect_uri(&self, nostr_wallet_connect_uri: &str, wn: tauri::State<'_, Whitenoise>) -> Result<()> {
-        secrets_store::store_nostr_wallet_connect_uri(&self.pubkey.to_hex(), nostr_wallet_connect_uri, &wn.data_dir)
-            .map_err(AccountError::SecretsStoreError)
+    pub fn store_nostr_wallet_connect_uri(
```
a-mpch: nit: maybe a good thing to do is adding clippy and rustfmt files so we can code in same standards!

src-tauri/src/commands/groups/send_mls_message.rs
a-mpch: +1 and would be nice to add some test for this branch

src-tauri/db_migrations/0002_add_media_files.sql
a-mpch: it's down in the foreign key or am I missing something?

a-mpch: oh, now it's gone +1

a-mpch: I guess because the message will have attached the sha of the file. So if seen we just need to fetch that one. If we bound 2 columns, we would have double reference.
I don't know if we would like to do the join to fetch messages. Maybe just _fetch_ from the cache when displayed

a-mpch (commented):
keep the good work!

I just reviewed the code and modules, didn't try locally the changes. But for me seems really logical and approachable separating _media_ with _nostr_manager_.

I saw a lot of _formatting_ commits that probably would be solved if we had a config file at the root or we could have clippy running in the CI with stricter rules.

src-tauri/db_migrations/0002_add_media_files.sql
erskingardner: Yeah - we have a chicken and egg issue here. 

We want to upload the file before we have the message_id (for better ux) but we don't have the message_id value at that stage. 

I thought about two ways to handle: 

1. we can remove it altogether. this is way simpler and what I went for. this means that all cached files for a group are stored in a directory of the group_id. I think this is likely fine. It also means that if someone uploads exactly the same file twice to a group we'll see that it has the same hash and we won't bother uploading the file to Blossom or re-caching it. We'll just use the previously cached version (I _think_ this is ok. I'm sure it might cause some weirdness at the edges but I'm happy to take that risk right now unless someone comes up with a blocker). 
2. or we can upload without the message_id value and then only cache and create the media_file database row entry once the user sends the message. This is definitely possible, just means re-ordering things and putting more processing on the moment of message sending (yes, of course we can create a background job for this - but it's more complexity).

src-tauri/db_migrations/0002_add_media_files.sql
erskingardner: Ok - so I WAS going to leave this as text but I did some research this morning and decided that if we're going to change to JSONB, now is the time. As part of moving to a new apple developer account, I have change the bundle identifier of the app which is going to nuke the database of everyone using the app on the next release anyway. There is no better time to make fundamental changes to the schema than now.

I'm going to change the fields related to media here in this PR. 

I'll create other PRs to handle 1) the bundle identifier change and 2) the changes needed to the current schema (all the stuff in the 0001 migration). 2 is a large set of changes unfortunately - it'll probably be a good chance to add some tests along the way though.

erskingardner: @jgmontoya @a-mpch @josefinalliende FYI 👆

src-tauri/db_migrations/0002_add_media_files.sql
a-mpch: I'm totally fine with 1

src-tauri/db_migrations/0002_add_media_files.sql
a-mpch: What kind of tests are you thinking?

erskingardner: Just all the rust testing that I've skipped so far. 

Changing to JSON on the other tables will cause a lot of code churn - so since we'll be there doing those changes, we might as well add some basic unit tests.

src-tauri/src/commands/groups/send_mls_message.rs
erskingardner: Yes, I've been adding media related tests as I go. Since this PR is already getting big, I don't want to add more refactoring to this method here. Let's do that separately once we have this out there and working.

src-tauri/src/accounts.rs
erskingardner: I think I'm running the absolute defaults but I'll look into it.

src-tauri/db_migrations/0002_add_media_files.sql
erskingardner: I'm not sure where this would be helpful. in the sense that we'll probably only create these rows when we've downloaded the file and we have all the data needed to create a row. There's no benefit to trying to create a media_files row in the db before we have the file.

src-tauri/src/media/mod.rs
erskingardner: Yeah I went back and forth on it. I guess single method files aren't the worst thing...

src-tauri/db_migrations/0002_add_media_files.sql
jgmontoya: Ok, I'm fine with 1, I hadn't thought of an n-to-n relationship. If needed we can have an additional table to associate media_files to messages.

This can be useful btw for example to allow the user to jump to the message were an image was sent and get the conversation context.

src-tauri/db_migrations/0002_add_media_files.sql
jgmontoya: I guess it's an order of operations thing. I was thinking first create the media_file record and then download the file and populate the file_path field. That way, if download fails for example, we can easily detect the empty file_path and re-download.

If we do it the other way around (first download, then create the row) and something fails after the download (or the download is incomplete or something), I thought it could be a bit harder to identify as the changes would only be on the filesystem and not on the db.

But tbf it doesn't seem like a a big deal anyways so 🤷

src-tauri/db_migrations/0002_add_media_files.sql
a-mpch: I can help you on doing test for 2. Helps me learning bit more of rust.
Maybe start out the PR and I'll continue or we merge both changes and I start over a new branch for testing cases

jgmontoya: It could be useful to have some issues maybe to track that. I might be able to help with some tests as well

erskingardner:
I think this is ready for review.

a-mpch:
seems good, I'll test it out locally during the weekend

src-tauri/src/media/mod.rs
erskingardner: Ahh - I should probably fix this before merging.

erskingardner:
CleanShot 2025-03-21 at 15 26 42

The dreaded "No space left on device"...

docker-compose.yml
```diff
@@ -1,28 +1,31 @@
...
-    ports:
-      - 3000:3000
     volumes:
       - ./dev/data/blossom-ts-server:/app/data
       - ./dev/blossom-server-ts-config.yml:/app/config.yml
+    ports:
+      - 3000:3000
+    restart: unless-stopped
+    environment:
+      - DEBUG=*
```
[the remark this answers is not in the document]
jgmontoya: I think this is not relevant because it's only for local dev environment

josefinalliende:
I was trying to test this branch in local but had an issue with account switching. It seems to be related to PR #113  so I'll leave more details there

erskingardner:
@coderabbitai - ignore this PR from now on

Issue #117, opened 2025-03-21
🖼️ Show Media in Messages


erskingardner:
We have implemented media uploading (with encryption/decryption, caching, and using blossom as the backend) in #108. This issue is about displaying media in conversations. In the newest version of rust-nostr (0.40), there is a `NostrParser` module that will help us parse message content for URLs, mentions, etc. This issue just deals with parsing out Blossom urls and displaying the media that they relate to in a good way. We'll create other issues for mentions, etc. 

## The flow

We currently allow uploading of images (`image/*`), videos (`video/*`), audio (`audio/*`), and pdf (`application/pdf`) files. We'll need to handle each slightly differently.

### Images
1. We should use the `NostrParser` module to parse out URLs.
2. Once we have a URL, we can check for a corresponding `imeta` tag.
3. This gives us dimensions (aspect ratio) and the blurhash that we can show immediately in the message. We might optionally want to show a loading spinner over the blurhash while we're performing the next step.
4. We should download and save the file locally. Then replace the blurhash image with the real image.

### Video
1. Same as above
2. Same as above
3. We get the dimensions of the video in the `imeta` tag and we can create a placeholder div with a camera icon and a loading spinner. We don't support any sort of streaming at this point so we just have to wait for the full download.
4. We'll display the video with a standard html video player

### Audio
Basically the same as video. 

### PDFs
Basically the same except that we should use the `phosphor-icons` PDF icon as a link anchor for downloading the PDF file. Ideally, we'd show the original file name under the icon.

Ideally, we shouldn't show the content of the message (with the URLs) at all while we're doing the initial processing (we can render the message as soon as we have the blurhash ready).

erskingardner:
@josefinalliende this might be a good one to work on next if you're up for it. We're parsing tokens from messages already but the styling of the different token types (other than text) is not done at all. We need to style URLs, Nostr references (mentions etc), images, etc. 

if you think you might have time and you're interested we can catch up this week to strategize more on how to get it done.

PR #127, opened 2025-03-25
Display media

erskingardner:
Implements #117 

- [x] Created two new migrations to add new fields to Messages table
- [x] Updated many commands to return `Message` struct instead of `UnsignedEvent` so that we can get more data about events (including the parsed content tokens)
- [x] Created a new `Parser` module that wraps `NostrParser` and implements serialization and deserialization
- [x] Updated the front-end Chat store to handle `Message` structs (called `CachedMessage` on the front-end)
- [x] Add basic token display in the front-end chats view.
- [x] Add lots of tests ❤️‍🔥 

## Summary by CodeRabbit

- **New Features**
  - Introduced a new component for displaying message tokens, enhancing the readability of chat messages.
  - Added support for cached messages, allowing for improved message handling and performance.

- **Improvements**
  - Refined chat interactions with updated message handling, deletion, and payment processes for a smoother user experience.
  - Enhanced backend indexing and caching to support more consistent messaging operations.
  - Updated configuration for Vite to streamline development and testing processes within the Tauri environment.
  - Introduced new fields in message structures to accommodate additional data, improving message representation.


josefinalliende (commented):
The new `CachedMessage` type makes sense to me, but I did find it a little confusing at first that it doesn’t only transform into a `Message`. Maybe renaming `Message` to `TextMessage`, Reaction to `ReactionMessage`, and Deletion to `DeletionMessage` could help make it clearer that a `CachedMessage` can become any of those types...

erskingardner:
yeah, i agree. maybe we should see about coming up with a better name? 

I like you're pattern. What about the following: 
- `Message` is what comes back from the backend (matching the naming there)
- `ChatMessage` is a kind 9 chat message.
- `ReactionMessage` is the reaction kind
- `DeletionMessage` is the deletion kind

This is extensible as we support more and more message kinds, it matches the backend and makes it a little more clear that we have a generic message that then becomes something more specific once processed.

erskingardner:
@coderabbitai full review

erskingardner:
Going to go ahead and merge this. We're doing the most basic processing of tokens accurately and I've added lots of tests. But all the front-end stuff will come with the UI redesign.

josefinalliende:
> yeah, i agree. maybe we should see about coming up with a better name?
> 
> I like you're pattern. What about the following:
> 
> * `Message` is what comes back from the backend (matching the naming there)
> * `ChatMessage` is a kind 9 chat message.
> * `ReactionMessage` is the reaction kind
> * `DeletionMessage` is the deletion kind
> 
> This is extensible as we support more and more message kinds, it matches the backend and makes it a little more clear that we have a generic message that then becomes something more specific once processed.

Like you pattern!

PR #128, opened 2025-03-27
UI layout

erskingardner:
## Summary by CodeRabbit

- **New Features**
  - Introduced enhanced theme support and mobile functionality (including barcode scanning) for a more adaptive experience.
  - Launched an updated wallet settings interface for seamless connection management.
  - Added a new `FormattedNpub` component for displaying Nostr public keys.
  - Implemented a new `getNostrWalletConnectBalance` function to retrieve wallet balances efficiently.

- **UI Enhancements**
  - Redesigned the login screen with a full-screen background, gradient overlay, and modern typography.
  - Updated header, chats, and settings layouts for improved navigation and clarity.
  - Enlarged the desktop app window for a more spacious view.
  - Simplified the settings page layout with new sections for profile, privacy, and developer settings.
  - Enhanced the user interface for displaying messages when no chats are found.

- **Technical Improvements**
  - Upgraded dependencies and refined error handling for a more stable application.
  - Enhanced clipboard utility with new functionality to read from the clipboard.
  - Improved the organization and documentation of various utility functions for better maintainability.

CleanShot 2025-03-28 at 17 48 41
CleanShot 2025-03-28 at 17 50 37
CleanShot 2025-03-28 at 17 49 43
CleanShot 2025-03-28 at 17 50 22
CleanShot 2025-03-28 at 17 49 56
CleanShot 2025-03-28 at 17 50 11


static/styles/colors.css
```diff
@@ -0,0 +1,87 @@
+@theme {
+    /* Glitch */
```
erskingardner: @josefinalliende Would love your thoughts on how I'm organizing things here. 

I like the named colors but in practice it's actually pretty messy to try and keep track of what everything, especially when they're almost all `glitch` colors.

src/routes/(app)/settings/+page.svelte
```diff
@@ -306,199 +301,152 @@ function toastCopyErrorMessage(errorMessage: string) {
...
+    }} tabindex="0" role="button" class="section-title-button">
+        <h2 class="section-title">Profile</h2>
+        {#if showProfileSection}
+            <button class="text-muted-foreground-light dark:text-muted-foreground-dark" onclick={() => console.log("open new profile sheet")} aria-label="Add a profile">
+                <AddLarge size={24} />
+            </button>
+        {/if}
+    </div>
+    {#if showProfileSection}
+        <div class="overflow-hidden p-0 m-0" transition:slide={slideParams}>
```
erskingardner, line 311: There's something janky going on with these accordion transitions. there is a small vertical jump at the end of a collapse that you can see and it's gross. I played around a little but didn't find the cause. This is mostly a reminder to fix it before merging.

src/lib/components/MessageTokens.svelte
```diff
@@ -0,0 +1,67 @@
...
+</div>
+
+<style>
+    .message-tokens {
+        display: inline;
+        white-space: pre-wrap;
+        word-break: break-word;
+    }
+
+    .url {
```
pepina-dev: It's a very personal opinion, but I prefer to use inline styles with Tailwind classes and leave CSS only for styles that can't be done with Tailwind.

src/app.css
```diff
@@ -22,3 +25,15 @@
     --padding-safe-left: env(safe-area-inset-left);
     --padding-safe-right: env(safe-area-inset-right);
 }
+
+button {
+    cursor: pointer;
+}
+
+.header-title {
```
pepina-dev: Following up on the previous comment — I prefer keeping Tailwind classes inline where they’re used, rather than creating custom classes in a global styles file. If we want everything to look consistent, it makes sense to extract a small component, but I’d still rather do that than rely on a separate style definition — especially when the custom class is just wrapping Tailwind utilities.

I’d leave custom CSS for things that can’t be easily done with Tailwind — otherwise it just adds an extra layer to maintain, and can introduce conflicts like this:
`class="header-title text-blue-500"` — where the Tailwind class might override the text color from the custom class. Without testing, it’s hard to tell which one will actually apply, since I don’t have Tailwind’s precedence rules memorized. And if I’m not mistaken, changing the order to `text-blue-500 header-title `could produce a different result.

src/lib/components/Button.svelte
```diff
@@ -0,0 +1,66 @@
+<script lang="ts">
+import type { Snippet } from "svelte";
+import Loader from "./Loader.svelte";
+
+type ButtonProps = {
+    variant?: "default" | "secondary" | "destructive" | "outline" | "ghost" | "link";
+    size?: "default" | "icon" | "sm" | "lg";
```
pepina-dev: [sug] The names "default" and "icon" under size don’t give me a clear idea of whether they’re larger or smaller than sm or lg. I think it would be easier to understand if these options followed a consistent size scale

static/styles/colors.css
pepina-dev: The truth is that I'm used to the following: primary and secondary are called that way and there’s not much explanation about what color they are, but that’s because in practice you usually know what the two main colors of the app are. For the rest of the colors, I just give them a name that tells me what color it is, not so much what it’s used for. Also, I find that use cases tend to change: for example, the color we define for the background might later be used in a button icon… or the color of a card might end up being used elsewhere too. So you end up with weird names like bg-primary-foreground-dark

pepina-dev: In this particular case, it's even trickier because we are using black and white as main colors, so in dark mode everything, even primary color changes.  So if we end up with primary-dark, primary-light,  secondary-dark and  secondary-light already feels confusing to me — I’d honestly rather just use black and white directly.

static/styles/colors.css
```diff
@@ -0,0 +1,87 @@
...
+    --color-glitch-400: oklch(79.9% 0.012 95.7); /* #afb1af */
+    --color-glitch-500: oklch(70.9% 0.017 95.7); /* #8c908c */
+    --color-glitch-600: oklch(57.7% 0.023 95.7); /* #727772 */
+    --color-glitch-700: oklch(48.2% 0.027 95.7); /* #5a605a */
+    --color-glitch-800: oklch(39.9% 0.031 95.7); /* #474c47 */
+    --color-glitch-900: oklch(27.1% 0.036 95.7); /* #2d312d */
+    --color-glitch-950: oklch(21.0% 0.037 95.7); /* #202320 */
+
+    /* Primary */
+    --color-primary-dark: oklch(0.987 0.024 297.388deg); /* #FAFBFA */
```
pepina-dev: Or I am getting confused and with primary-dark it is not refered to the primary color in dark mode and primary-light the primary color in light mode? Just the primary color a bit lighter or darker? In that case  I would prefer the use of tailwind scales from 100 to 900

josefinalliende (commented):
Great job! Just a few thoughts regarding the use of CSS classes that apply Tailwind styles. Specifically about the color definitions — for colors beyond primary and secondary, I’d personally prefer names that describe the actual color (like pink, red, etc.) rather than their intended use (like foreground or card). That said, it’s probably something we should align with the designers on, to make it easier to match Figma styles if the naming is consistent across both.

src/app.css
pepina-dev: That said, totally okay if you see it differently — maybe I’m just being a bit opinionated here 😊

josefinalliende:
PS: Something else that’s worked for me is holding off on adding colors until they’re actually being used — especially when there are a lot of them. I’ve seen Figma design systems with 30 colors where only 5 end up getting used in practice  in the components 😅 That said, if I do add a color, I like to include the full scale — that way, if a dev sees bg-rose-500 and needs rose-400, they’ll know it’s already available and won’t need to go configure it themselves.

static/styles/colors.css
pepina-dev: PS: Something else that’s worked for me is holding off on adding colors until they’re actually being used — especially when there are a lot of them. I’ve seen Figma design systems with 30 colors where only 5 end up getting used in practice in the components 😅 That said, if I do add a color, I like to include the full scale — that way, if a dev sees bg-rose-500 and needs rose-400, they’ll know it’s already available and won’t need to go configure it themselves.

erskingardner:
> Great job! Just a few thoughts regarding the use of CSS classes that apply Tailwind styles. Specifically about the color definitions — for colors beyond primary and secondary, I’d personally prefer names that describe the actual color (like pink, red, etc.) rather than their intended use (like foreground or card). That said, it’s probably something we should align with the designers on, to make it easier to match Figma styles if the naming is consistent across both.

The reason for these names is that they're the default shadcn colors, which means that we have to change way fewer styles in the components when we create them. I guess that's a one-time cost but keeping with the convention will probably make our lives easier when it comes to upgrading, etc.

src/lib/utils.ts
```diff
@@ -0,0 +1,6 @@
+import { type ClassValue, clsx } from "clsx";
+import { twMerge } from "tailwind-merge";
+
+export function cn(...inputs: ClassValue[]) {
```
pepina-dev, line 4: I think `cn `isn’t a very descriptive name. Maybe something like `mergeClasses`, `classNames`, or `twMergeClasses` would be clearer?

tailwind.config.ts
```diff
@@ -0,0 +1,116 @@
...
+                },
+                accent: {
+                    DEFAULT: "hsl(var(--accent) / <alpha-value>)",
+                    foreground: "hsl(var(--accent-foreground) / <alpha-value>)",
+                },
+                popover: {
+                    DEFAULT: "hsl(var(--popover) / <alpha-value>)",
+                    foreground: "hsl(var(--popover-foreground) / <alpha-value>)",
+                },
+                card: {
```
pepina-dev, line 67: [sug] I feel that names like card, popover, and sidebar are very tied to where the color is used, rather than what the color actually is. I'd prefer names that describe the color itself instead of its usage—it makes the palette easier to reason about and reuse without needing to dig into where each color is applied.

josefinalliende (commented):
Huge job! I'm really impressed by how quickly it's starting to look like the new design—those screenshots look great. I haven't tested it locally yet, but I left a couple of comments in the meantime 😉

josefinalliende:
> > Great job! Just a few thoughts regarding the use of CSS classes that apply Tailwind styles. Specifically about the color definitions — for colors beyond primary and secondary, I’d personally prefer names that describe the actual color (like pink, red, etc.) rather than their intended use (like foreground or card). That said, it’s probably something we should align with the designers on, to make it easier to match Figma styles if the naming is consistent across both.
> 
> The reason for these names is that they're the default shadcn colors, which means that we have to change way fewer styles in the components when we create them. I guess that's a one-time cost but keeping with the convention will probably make our lives easier when it comes to upgrading, etc.

Oh sorry I hadnt realized that. So you need this names in order the library works right?

tailwind.config.ts
pepina-dev: Sorry, I just saw the comment regarding class names related to shadcn, so I think this does not apply anymore 😅

src/lib/utils.ts
erskingardner: it was autogenerated by shadcn-svelte. I just left it alone.

tailwind.config.ts
erskingardner: Yeah, it's. touch confusing but i think for now we should just leave things as is. So far it's been really really easy to modify one or two things in a component and it "just works". I would rather have some slightly odd names instead of renaming and creating our own set of names, even if they were slightly easier. Also @stackingsaunter is using the standard names in figma so things have been really easy to translate over to code because of that.

src/app.css
```diff
@@ -44,48 +17,99 @@ html {
...
+        --radius: 0.5rem;
+        --sidebar-background: 0 0% 98%;
+        --sidebar-foreground: 240 5.3% 26.1%;
+        --sidebar-primary: 240 5.9% 10%;
+        --sidebar-primary-foreground: 0 0% 98%;
+        --sidebar-accent: 240 4.8% 95.9%;
+        --sidebar-accent-foreground: 240 5.9% 10%;
+        --sidebar-border: 220 13% 91%;
+        --sidebar-ring: 217.2 91.2% 59.8%;
+    }
```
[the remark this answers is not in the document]
erskingardner: what are you!? an EU bureaucrat? 😂

erskingardner:
Merging this so that more of us can keep going on the UI changes.

PR #163, opened 2025-04-25
Nostr mls

erskingardner:
This PR moves us to the new Noster MLS library which handles storing lots of data that we used to storing White Noise. This is a major change which vastly simplifies the White Noise code base and should give us a much easier platform to build on going forward.

## Summary by CodeRabbit

- **New Features**
  - Introduced comprehensive documentation for Svelte 5 and SvelteKit 2 migration and usage.
  - Added support for managing group welcomes, including accepting, declining, and listing welcome events.
  - Added commands to fetch active groups and group relays.

- **Bug Fixes**
  - Improved error handling and state management for Nostr MLS initialization and media uploads.
  - Reduced navigation delay after Nostr initialization for a faster user experience.

- **Refactor**
  - Migrated group, message, and invite management to a new architecture using the Nostr MLS library with SQLite storage.
  - Simplified and streamlined event processing, group, message, and media handling.
  - Updated database schema to remove group and messaging tables, focusing on account and relay management.
  - Removed invite-related commands and replaced with welcomes-related commands.
  - Centralized state management and removed legacy secrets storage for MLS export secrets.
  - Updated media caching and encryption to use typed group IDs and simplified key handling.
  - Updated frontend components and stores to use new welcome and message types with tokenization.
  - Replaced legacy group and invite types with structured Nostr MLS group and welcome types.

- **Chores**
  - Updated dependencies and configuration files for improved compatibility and maintainability.
  - Enhanced .gitignore to exclude additional AI-related files.
  - Added automation tasks and scripts for publishing NIP-89 handlers.

- **Documentation**
  - Added detailed migration and usage guides for developers.


erskingardner:
The checks are all failing because it's still pointing to my local version of rust-nostr in the dependencies. I'll fix that once I've checked through more of the app

src/lib/types/nostr.ts
```diff
@@ -70,7 +70,7 @@ export type NGroup = {
 };
 
 export type MlsGroupId = {
-    value: Uint8Array;
+    value: { vec: Uint8Array };
```
pepina-dev, line 75: I got a bit confused with this.  When I read the  name of  `MlsGroupId` I think that it is a single id, not an array of ids. For me it would be more intuitive to represent an array of ids like  `MlsGroupId[]`

josefinalliende (commented):
When I tried to run this branch locally  I got some errors:
1. It appears to me that now rustr-nostr  needs to be cloned in the local machine for this to work.After running bun install and then running bun tauri dev  I got the error of the screenshot below. After cloning the rust-nostr repo (`git clone https://github.com/rust-nostr/nostr.git rust-nostr`)it does not fail with that error (it fails with another one, but later after the building logs).  
Captura de pantalla 2025-04-28 a la(s) 6 12 45 p m


2. The second error I get is related with migrations. I tried it on my existing db (with messages) and it failed with the error in the screenshot below. 
Captura de pantalla 2025-04-28 a la(s) 6 25 19 p m

Then I deleted the whitenoise sqlite file but I still get the same  error: "Failed to create database: Migrate(ExecuteMigration(Database(SqliteError { code: 1, message: "no such table: messages" }), 3))"

Maybe am I just doing something wrong for the second one?

src/lib/types/nostr.ts
erskingardner: This is confusing but it's because I'm following a Rust type that I have don't have control over. In this case, it's the GroupId value from the OpenMLS library. Because it's being serialized from Rust by Tauri we get this strange shape and have to just deal with it. :/ I'm going to hopefully write some simple methods that will help with extracting things but I think overall, on the front-end, we'll just want to pass around full objects (groups, welcomes, etc) in params and allow Rust to manage all the serialization and deserialization for us.

erskingardner:
@josefinalliende 

> The second error I get is related with migrations. I tried it on my existing db (with messages) and it failed with the error in the screenshot below.

That second error is because there are old migration files hiding in your `src-tauri/target` directory. You can safely just delete the entire directory and then it will be rebuilt on the next run. This fixed it for me.

PR #170, opened 2025-05-13
Goodbye Tauri 👋

erskingardner:
This PR will completely remove Tauri and leave behind only a rust crate containing core White Noise functionality that is independent of the user interface used. 

The immediate goal of this is to allow us to move to Flutter but do so in a way that leaves the app open to function completely independently of the front-end presentation layer chosen. Want to run White Noise from a CLI or a REST API? Totally possible.

This will be a long running branch/draft PR with several people working on it. If you're going to contribute on this branch please make sure to: 

1. Check out the branch. 
2. Make commits as you're working. 
3. Make sure to rebase (not merge) any changes that are pushed to the branch as you're working.
4. **IMPORTANT:** Once you have something finished and the app is in a working state, **squash your commits into a single commit** and push that to the branch. 

This ensures that we have a simple, linear changelog of working commits.


src-tauri/Cargo.toml
```diff
@@ -18,11 +18,13 @@ crate-type = ["staticlib", "cdylib", "rlib"]
 tauri-build = { version = "2", features = [] }
 
 [dependencies]
+anyhow = { version = "1.0.98", features = ["backtrace"] }
```
erskingardner: @jmcorgan curious about this one. we were using `thiserror` everywhere (and my errors were very badly organized), but wondering if anyhow has extra functionality that you like?

jmcorgan: The short answer is `thiserror` is for libraries, `anyhow` is for applications. It's used in the `test-backend` binary I introduced, and it's main feature is that it can convert any kind of error that implements the `std::error::Error` trait into something that can be dealt with at the top-level. No plans to use it in the actual module itself. If it turns out to be an extra crate that we could do without we can remove it.

jmcorgan:
The current code base follows a pattern where the front end application makes a call to the back end via the function attribute `#[tauri::command]`, with a tauri state object for the `Whitenoise` state and a tauri `app_handle` for fishing out parameters or emitting events back to the front end. These arguments are passed through sometimes several function calls deep into the backend code. The tauri bootstrapping and existing `Whitenoise` struct manage getting all this into place at startup so the front end can issue commands and receive events emitted by Rust code.

The first step is to create/modify the `Whitenoise` struct to be tauri independent, and accept a new struct `WhitenoiseConfig` that contains any configuration or startup state needed to be passed in externally. Then the startup steps of initializing directories, the database, nostr and mls underlying objects can occur. This part is done.

The goal is to turn all the existing commands and related code into impl functions of the `Whitenoise` struct. To accomplish this, we strip out all the tauri related function arguments and code from the command and function hierarchy, while leaving in the the parts of the code that were calling functions on or dereferencing the `wn` parameter. This part is done.

At this point several things need to happen:

* Move commands into impl functions for `Whitenoise`
* Replace `wn` usage with `self` where possible and added data members to `Whitenoise` if needed
* In some cases it might be possible to pass down the specific data needed through the function hierarchy instead of using `self`.

The result of all this will be a top-level `Whitenoise` struct that holds all application state and has functions that can be called to manipulate that state. How that changed state or events originating from relays gets passed back to the front end is still being considered.

To test all this, I added a placeholder `test-backend.rs` command line program that instantiates the `Whitenoise` struct and will be able to call functions on it, and can evolve as needed to drive the API independent of the front end.

Currently the code is code is completely broken from all the tauri removal. The next batch of code in this refactor will be to get above steps completed, and at that point a consolidation and reorganization of the public functions and state exposed to the front end can occur.

PR #174, opened 2025-05-19
Feat/media images

josefinalliende:
# Context
This PR adds replaces the media urls with images.  More context in this issue: https://github.com/parres-hq/whitenoise/issues/117. It needs media download logic done by @jgmontoya  in #173 

### What has been done
Image media files are now shown in the conversation

**Commit by commit:**
- Adds a method to generate a svg from blurhash string
- Adds tauri command to fetch all media files of a group
- Adds media tyoes
- Adds method to find imeta tags
- Adds utils related to media: get mimetype, attachments, read files (needed to add permissions to read file sin data dir)
- Process nostr event tags into media attachments
- Adds media store to map already cached media files 
- Uses media utils in the message bar
- Adds components to show images
- Changes chat page to use media store and new components to show images

**Screen recording**

https://github.com/user-attachments/assets/e479a546-7616-4b5f-88d7-6de492b9c856

### What's next 
Showing audios, videos and docs. 

## Summary by CodeRabbit

- **New Features**
  - Added support for media attachments in chat messages, including display, download, and error handling.
  - Introduced new components for displaying media attachments in messages.
  - Implemented a media store for managing media files and downloads.
  - Enhanced chat UI to show media grids and handle media-related events.
  - Added localization for media download errors in multiple languages.

- **Bug Fixes**
  - Improved handling of message tokens to remove trailing whitespace and filter out media URLs.

- **Tests**
  - Added comprehensive tests for media utilities, blurhash, message processing, and chat store behavior with media attachments.

- **Chores**
  - Updated dependencies to include the blurhash library.
  - Extended filesystem permissions for media file access.

PR #175, opened 2025-05-20
fix(build): try to fix blurhash build issue

josefinalliende:
## Context
The build is failing in master due to my last PR #174 
Captura de pantalla 2025-05-20 a la(s) 1 00 59 p m

What’s strange is that the build works fine on my Mac. Here’s the error message from CI:

Captura de pantalla 2025-05-20 a la(s) 1 01 43 p m

## What has been done
Based on the error, it seems we need to explicitly treat blurhash as an external dependency. I’m not entirely sure why this is necessary, but I’ve applied the suggested fix to see if it resolves the issue.

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -26,6 +26,12 @@ const excludes = [
 export default defineConfig(() => ({
     plugins: [sveltekit()],
 
+    build: {
+        rollupOptions: {
+            external: ["blurhash"],
+        },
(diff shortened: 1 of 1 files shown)


josefinalliende:
github actions not triggered? That's weird

josefinalliende:
It's no just because being a draft...

PR #178, opened 2025-06-03
Beginning of Account methods

erskingardner:
This updates the structure of how we're initializing the app and changes the Account struct and database Accounts some. 

First public crate method `create_identity` added. 🎉


jmcorgan:
> @jmcorgan I'm going to merge this now but would still love your thoughts.

Sent you a bunch of comments in DM. Overall the new structure is following our discussion; there are a few items I found that were likely in the old code before this reorg.

PR #181, opened 2025-06-06
Refactor nostr_manager and event_processor

erskingardner:
This PR simplifies the nostr_manager significantly (there is still some dead code to remove there) and makes it account aware. 

it also removes the event_processor module and instead moves the event queue and processing logic to the Whitenoise impl level so that we have access to everything we need (Accounts, Database, etc.) to process events correctly.


scripts/check-clippy.sh
```diff
@@ -11,5 +11,5 @@ cargo +$version --version || rustup install $version
 cargo +$version clippy --version || rustup component add clippy --toolchain $version
 
 echo "Checking clippy"
-cargo +$version clippy --all-targets --all-features --no-deps -- -D warnings
+cargo +$version clippy --all-targets --all-features --no-deps
```
erskingardner: We'll put this back later.

src/accounts.rs
```diff
@@ -121,10 +121,16 @@ impl Account {
             let nostr_mls_guard = self.nostr_mls.lock().unwrap();
```
jmcorgan: The unwrap() here will panic if the lock() fails, you can propagate this upward with something like:

```
let nostr_mls_guard = self.nostr_mls.lock()
            .map_err(|_| AccountError::NostrMlsNotInitialized)?;
```

src/accounts.rs
```diff
@@ -121,10 +121,16 @@ impl Account {
...
             if let Some(nostr_mls) = nostr_mls_guard.as_ref() {
                 let groups = nostr_mls.get_groups()?;
-                group_ids = groups.iter().map(|g| g.nostr_group_id).collect::<Vec<[u8; 32]>>();
+                group_ids = groups
```
jmcorgan: this could be made more efficient/readable:

```
if let Some(nostr_mls) = nostr_mls_guard.as_ref() {
        let groups = nostr_mls.get_groups()?;
        Ok(groups
            .iter()
            .map(|g| hex::encode(g.nostr_group_id))
            .collect())
    } else {
        Ok(Vec::new())
    }
```

src/accounts.rs
```diff
@@ -148,10 +154,7 @@ impl Whitenoise {
     ///
     /// Returns a `WhitenoiseError` if the account is not found, if deserialization fails,
     /// or if initialization of the NostrManager or NostrMls fails.
-    pub(crate) async fn find_account_by_pubkey(
-        &self,
-        pubkey: &PublicKey,
-    ) -> Result<Account> {
+    pub(crate) async fn find_account_by_pubkey(&self, pubkey: &PublicKey) -> Result<Account> {
         let mut txn = self.database.pool.begin().await?;
```
jmcorgan: I'm not familiar with sqlx crate. Does a read-only SELECT require a txn at all? Even if so, it's not being committed, though I don't know if that becomes automatic when it goes out of scope.

src/accounts.rs
```diff
@@ -189,10 +192,7 @@ impl Whitenoise {
     ///
     /// Returns a `WhitenoiseError` if any database operation, event fetching, serialization,
     /// or key storage fails.
-    pub(crate) async fn add_account_from_keys(
-        &self,
-        keys: &Keys,
-    ) -> Result<Account> {
+    pub(crate) async fn add_account_from_keys(&self, keys: &Keys) -> Result<Account> {
```
jmcorgan, line 203: It seems these are all or nothing operations that should be done within a single transaction context. If the account gets saved but the key storage fails, you'll have an account without keys.

src/accounts.rs
```diff
@@ -359,20 +358,14 @@ impl Whitenoise {
...
     /// Returns a `WhitenoiseError` if any database or Nostr operation fails.
-    pub(crate) async fn onboard_new_account(
-        &self,
-        account: &mut Account,
-    ) -> Result<Account> {
+    pub(crate) async fn onboard_new_account(&self, account: &mut Account) -> Result<Account> {
         tracing::debug!(target: "whitenoise::accounts::onboard_new_account", "Starting onboarding process");
 
         // Set onboarding flags
         account.onboarding.inbox_relays = true;
```
jmcorgan: Should probably set these to false, then only set to true when the op succeeds.

src/accounts.rs
erskingardner: Great catch - you don't need a txn for `Select` statements. You can just use the database.pool directly.

src/accounts.rs
erskingardner: Refactored the method to do things in a better order and try and clean up after itself.

src/accounts.rs
erskingardner: Yes, definitely.

PR #182, opened 2025-06-07
Refactor library code for simplicity

erskingardner:
This refactors all whitenoise methods into a single `impl Whiltenoise` in the `whitenoise.rs` file. This makes things easier to understand and ensures that all other modules are standalone and not mixing whitenoise code. 

I've also moved the old code we'll eventually delete to an `_OLD` dir. 

This also adds tests and expands what the integration_test is checking.


.cursor/TEST_CLEANUP_SUMMARY.md
```diff
@@ -0,0 +1,130 @@
+# Test Cleanup Summary
```
delcin-raj: Good initiative

src/nostr_manager/mod.rs
```diff
@@ -102,22 +101,49 @@ impl NostrManager {
...
+            target: "whitenoise::nostr_manager::new",
+            "Attempting to connect to relays with {}s timeout...",
+            connection_timeout.as_secs()
+        );
 
-            // Connect to the default relays
-            client.connect().await;
+        // Use timeout for connection to prevent indefinite blocking
+        match tokio::time::timeout(connection_timeout, client.connect()).await {
+            Ok(_) => {
```
delcin-raj: If I'm not wrong nostr client on connection might return an error. 
That is not handled here.

src/nostr_manager/mod.rs
```diff
@@ -102,22 +101,49 @@ impl NostrManager {
...
+                tracing::debug!(
+                    target: "whitenoise::nostr_manager::new",
+                    "Successfully connected to relays"
+                );
+            }
+            Err(_) => {
```
delcin-raj: What if all the clients failed to connect?
Should this function succeed in that case?

erskingardner: Yeah, tbh, I'm going to work on this. It was added as part of trying to make the tests work ok with relays that failed but I don't like that it just fails and moves on.

src/accounts.rs
```diff
@@ -54,7 +51,7 @@ pub struct OnboardingState {
 
 /// This is an intermediate struct representing an account in the database
 #[derive(Serialize, Deserialize, Debug, Clone, sqlx::FromRow)]
-struct AccountRow {
+pub(crate) struct AccountRow {
```
delcin-raj, line 55: For the Account struct nostr_mls field,
The Mutex used should be from tokio.
Else it will block on acquiring lock.

PR #189, opened 2025-06-08
Load accounts from database on initialize

### LINKED ISSUE #188: Ensure that we're loading previously created/logged in accounts on initialize
Right now we don't load Accounts from the database that were previously stored there. We need to do that on initialize of the Whitenoise struct

erskingardner:
Fixes #188

### DIFF
--- a/src/whitenoise.rs
+++ b/src/whitenoise.rs
@@ -139,9 +139,7 @@ impl Whitenoise {
         // Create SecretsStore
         let secrets_store = SecretsStore::new(data_dir);
 
-        // TODO: Load accounts from database
-
-        // Create Whitenoise instance
+        // Load all accounts from database
         let mut whitenoise = Self {
             config,
             database,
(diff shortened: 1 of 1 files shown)


src/whitenoise.rs
```diff
@@ -398,6 +404,94 @@ impl Whitenoise {
...
+                );
+                // Continue loading other accounts even if one fails
+                continue;
+            }
+
+            // Add the account to the HashMap first, then trigger background fetch
+            accounts.insert(pubkey, account.clone());
+
+            // Trigger background data fetch for each account (non-critical)
+            if let Err(e) = self.background_fetch_account_data(&account).await {
```
jmcorgan, line 469: This looks like it works but you're serializing all your calls to this background task with the await.

delcin-raj: One way to resolve this is to collect these futures in a vec,
`vec.push(self.background_fetch_account_data(&account))`

Then await them outside the for loop.

erskingardner: I don't think any of this matters. We don't care about the returns. The events that come back from these fetches will be processed by the event professor loop and also stored on the nostr client cache. This is a quick way simply trigger a fetch.

PR #193, opened 2025-06-11
Add contact management methods and integration tests

### LINKED ISSUE #186: Create `whitenoise.update_account_contacts` method
This method should update the contact list of a given account and published the new contact list to the Nostr network based on the users nostr relays.

erskingardner:
Closes #186

### DIFF
--- a/src/bin/integration_test.rs
+++ b/src/bin/integration_test.rs
@@ -310,6 +310,8 @@ async fn main() -> Result<(), WhitenoiseError> {
         initial_contacts.len()
     );
 
+    assert_eq!(initial_contacts.len(), 0);
+
     // Add a contact and publish to relays
     match whitenoise
         .add_contact(&active_account, test_contact_pubkey)
--- a/src/whitenoise.rs
+++ b/src/whitenoise.rs
@@ -1102,51 +1102,6 @@ impl Whitenoise {
         Ok(relays)
     }
 
-    /// Loads a user's contact list from the Nostr network.
-    ///
-    /// This method retrieves the user's contact list, which contains the public keys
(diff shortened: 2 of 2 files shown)


src/bin/integration_test.rs
```diff
@@ -285,5 +285,91 @@ async fn main() -> Result<(), WhitenoiseError> {
...
+        "Testing add contact logic for: {}",
+        test_contact_pubkey.to_hex()
+    );
+
+    // Load current contact list to verify it's empty initially
+    match whitenoise.load_contact_list(active_account.pubkey).await {
+        Ok(current_contacts) => {
+            tracing::info!(
+                "Current contact list has {} contacts",
+                current_contacts.len()
```
delcin-raj: Length should be asserted to 0.

src/bin/integration_test.rs
```diff
@@ -285,5 +285,91 @@ async fn main() -> Result<(), WhitenoiseError> {
...
+            );
+
+            // Verify that adding a new contact would be valid (contact doesn't exist)
+            let can_add = !current_contacts.contains_key(&test_contact_pubkey);
```
delcin-raj: It's better to call the `add_contact` and `remove_contact` functions and do the assertions.

erskingardner:
> So the contact lists should always be a vec and the tests should call the apis and assert the results.

So you're saying that we should always keep the contact list as a vec (whether we're querying in the nostr_manager or doing any other ops). Just want to make sure I understand your point before I make any changes. The idea being that we're basically just converting to a hashmap, then scanning/searching, then converting back to a vec before we publish and that's most wasteful than a slightly slower search through a plan vec but without the conversions in and out of the hashmap?

PR #195, opened 2025-06-12
Implement sqlx::FromRow for Account

### LINKED ISSUE #194: Work on smooth conversions between database types and rust types.
Implement `sqlx::FromRow` rather than manual conversions.

delcin-raj:
* For our use case it doesn't make sense to complicate the db with multiple tables and foreign keys
* This approach requires the least changes while improving the ergonomics of sqlx queries

closes #194

### DIFF
--- a/src/accounts.rs
+++ b/src/accounts.rs
@@ -26,7 +26,7 @@ pub enum AccountError {
     NostrMlsNotInitialized,
 }
 
-#[derive(Serialize, Deserialize, Debug, Clone, sqlx::FromRow)]
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
 pub struct AccountSettings {
     pub dark_theme: bool,
     pub dev_mode: bool,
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -16,8 +16,8 @@ pub mod whitenoise;
 
 // Re-export main types for library users
 pub use accounts::{Account, AccountSettings, OnboardingState};
-pub use relays::RelayType;
 pub use error::WhitenoiseError;
+pub use relays::RelayType;
```
delcin-raj, line 20: This change is due to `just check-fmt`

src/accounts.rs
```diff
@@ -107,7 +145,7 @@ impl Account {
             settings: AccountSettings::default(),
             onboarding: OnboardingState::default(),
             last_synced: Timestamp::zero(),
-            nostr_mls: Arc::new(tokio::sync::Mutex::new(None)),
+            nostr_mls: Arc::new(Mutex::new(None)),
```
delcin-raj, line 148: This was already imported.

PR #197, opened 2025-06-12
Update account settings

### LINKED ISSUE #185: Create `whitenoise.update_account_settings` method
This method allows us to update the account_settings member of an account. This should update the account in the `whitenoise struct` and return the Account with the new settings.

### LINKED ISSUE #196: Implement whitenoise.fetch_account_settings
If the user is able to update, they should be able to fetch it as well.

This also helps in integration tests.

delcin-raj:
closes #185
closes #196

### DIFF
--- a/src/accounts.rs
+++ b/src/accounts.rs
@@ -26,7 +26,7 @@ pub enum AccountError {
     NostrMlsNotInitialized,
 }
 
-#[derive(Serialize, Deserialize, Debug, Clone, sqlx::FromRow)]
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
 pub struct AccountSettings {
     pub dark_theme: bool,
     pub dev_mode: bool,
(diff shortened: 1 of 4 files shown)


src/accounts.rs
```diff
@@ -227,6 +265,57 @@ mod tests {
...
 
+    #[tokio::test]
+    async fn test_from_row_implementation() {
+        use sqlx::SqlitePool;
+
+        // Create an in-memory database for testing
+        let pool = SqlitePool::connect(":memory:").await.unwrap();
+
+        // Apply the accounts table schema
+        sqlx::query(
```
erskingardner: Should we use the actual migrations here to make sure that we don't drift in the tests from what the real database state is?

src/whitenoise.rs
```diff
@@ -700,6 +670,66 @@ impl Whitenoise {
...
+    /// # Arguments
+    /// * pubkey
+    ///
+    /// # Returns
+    /// Returns `AccountSettings` on success
+    ///
+    /// # Errors
+    ///
+    /// Returns a `WhitenoiseError` if account does not exist or database operation fails or if serialization fails
+    pub async fn fetch_account_settings(&self, pubkey: &PublicKey) -> Result<AccountSettings> {
```
erskingardner, line 723: We use `load_*` for the other methods. I don't care which term we go with `fetch` or `load`, but I think we should decide on a pattern and stick to it for consistency.

erskingardner: We're going to use `fetch_`

PR #207, opened 2025-06-17
Create subscriptions

erskingardner:
This adds subscription bootstrapping on `create_identity` and `login`. Subscriptions have a custom sub_id that allows the event_processor to process the events for the right account as they come in. 

TODO: We still need to implement code that updates the subscriptions when we're creating new groups, when a user accepts an invite to a new group, or when their contact list changes. These will be handled in other PRs.


delcin-raj (approved):
Have read through the changes. LGTM.

Haven't reviewed the tests thoroughly

PR #212, opened 2025-06-18
Fix parsing of subscription_id to get accounts

erskingardner:
This fixes a bit of the new subscription logic that I overlooked. How we parse the account pubkey out from the subscription ID value.

### DIFF
--- a/src/whitenoise.rs
+++ b/src/whitenoise.rs
@@ -18,6 +18,8 @@ use crate::relays::RelayType;
 use crate::secrets_store::SecretsStore;
 use crate::types::ProcessableEvent;
 
+use sha2::{Digest, Sha256};
+
 #[derive(Clone, Debug)]
 pub struct WhitenoiseConfig {
     /// Directory for application data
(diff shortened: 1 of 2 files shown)


src/whitenoise.rs
```diff
@@ -1714,14 +1716,28 @@ impl Whitenoise {
     }
 
     /// Extract the account pubkey from a subscription_id
-    /// Subscription IDs follow the format: {pubkey}_{subscription_type}
-    fn extract_pubkey_from_subscription_id(subscription_id: &str) -> Option<PublicKey> {
+    /// Subscription IDs follow the format: {hashed_pubkey}_{subscription_type}
+    /// where hashed_pubkey is a 12-character SHA256 hash of the pubkey + session salt
+    async fn extract_pubkey_from_subscription_id(
+        &self,
+        subscription_id: &str,
```
delcin-raj, line 1723: This should be bytes &[u8]?

Not all hashes are valid string in rust

src/whitenoise.rs
```diff
@@ -1714,14 +1716,28 @@ impl Whitenoise {
...
+    ) -> Option<PublicKey> {
         if let Some(underscore_pos) = subscription_id.find('_') {
-            let pubkey_str = &subscription_id[..underscore_pos];
-            PublicKey::parse(pubkey_str).ok()
-        } else {
-            None
+            let hash_str = &subscription_id[..underscore_pos];
+            // Get all accounts and find the one whose hash matches
+            let accounts = self.accounts.read().await;
+            for (pubkey, _) in accounts.iter() {
```
delcin-raj: accounts.keys().iter will be better

src/whitenoise.rs
```diff
@@ -1714,14 +1716,28 @@ impl Whitenoise {
     }
 
     /// Extract the account pubkey from a subscription_id
-    /// Subscription IDs follow the format: {pubkey}_{subscription_type}
-    fn extract_pubkey_from_subscription_id(subscription_id: &str) -> Option<PublicKey> {
+    /// Subscription IDs follow the format: {hashed_pubkey}_{subscription_type}
+    /// where hashed_pubkey is a 12-character SHA256 hash of the pubkey + session salt
```
delcin-raj: hashed_pubkey = SHA256(session salt || accouny_pubkey)[..12].

The above line is easy to read in my opinion.

src/whitenoise.rs
erskingardner: Hmm, I think this might be something we have to handle in the creation of the hashes then. we have to send a valid string as the subscription_id to relays as they'll only take valid utf-8 strings. I'll look into and make sure.

erskingardner: I double checked. I'm actually hex encoding the hash on the front-end so this string coming back is going to be valid.

PR #215, opened 2025-06-18
Implement create group method

delcin-raj:


src/whitenoise.rs
```diff
@@ -1972,6 +1972,163 @@ impl Whitenoise {
...
+    /// * `Ok(Group)` - The newly created group
+    /// * `Err(String)` - Error message if group creation fails
+    ///
+    /// # Flow
+    /// 1. Validates that active account is the creator and signer
+    /// 2. Validates member and admin lists
+    /// 3. Fetches key packages for all members
+    /// 4. Creates MLS group with NostrMls
+    /// 5. Sends welcome messages to all members via Nostr
+    /// 6. Adds group to GroupManager database
```
erskingardner: What's this? Doesn't seem to be in the code below?

src/whitenoise.rs
```diff
@@ -1972,6 +1972,163 @@ impl Whitenoise {
...
+            .map_err(WhitenoiseError::from)?;
+
+            self.nostr
+                .client
+                .send_event_to(group_relays.clone(), &wrapped_event)
+                .await
+                .map_err(WhitenoiseError::from)?;
+        }
+
+        self.nostr
```
erskingardner, line 2103: We should create a more specific method that just updates the group subscription like I did with the contact_list_metadata subscriptions

src/whitenoise.rs
```diff
@@ -1972,6 +1972,163 @@ impl Whitenoise {
...
+                member_key_package_events.push((event.id, key_package));
+            }
+
+            let create_group_result = nostr_mls
+                .create_group(
+                    group_name,
+                    description,
+                    &creator_pubkey,
+                    &member_pubkeys,
+                    member_key_package_events
```
erskingardner: You need to get the serialized keypackages from the content of these messages, deserialize it and include the actual KeyPackage objects. not the events. You can see the code for this in the OLD_key_packages.rs file.

src/whitenoise.rs
```diff
@@ -1972,6 +1972,163 @@ impl Whitenoise {
...
+                "Welcome rumor: {:?}",
+                welcome_rumor
+            );
+
+            // Create a timestamp 1 month in the future
+            use std::ops::Add;
+            let one_month_future = Timestamp::now().add(30 * 24 * 60 * 60);
+
+            // sign and publish welcome rumor as a gift wrap
+            let signer = &self
```
erskingardner: We won't have a signer set at this point. The new pattern, since we might have multiple accounts running at any given time, is that we have a `publish_event_builder_with_signer` method that sets the signer, publishes, and then immediately unsets the signer. You might need to create a new method specifically for giftwrap publishing but it will be very similar.

src/nostr_manager/mod.rs
```diff
@@ -48,7 +48,7 @@ pub struct NostrManagerSettings {
 #[derive(Debug, Clone)]
 pub struct NostrManager {
     pub settings: Arc<Mutex<NostrManagerSettings>>,
-    client: Client,
+    pub(crate) client: Client,
```
erskingardner: I don't think we need this to be pub(crate) if we make that signer/publisher method a method on the nostr_manager instead.

src/whitenoise.rs
delcin-raj: I'm doing that, the naming of the variable is confusing I guess

src/whitenoise.rs
delcin-raj: Does that mean active_account will be deprecated?

erskingardner: No. Think of active_account as the account that is currently being used in the UI. The subscriptions and event processor are running in the background for all logged in accounts though.

delcin-raj: Realized that

PR #216, opened 2025-06-19
Implement fetch groups for whitenoise

### LINKED ISSUE #202: Add `whitenoise.fetch_groups` method
You can look at the `_OLD/OLD_commands/groups/get_active_groups.rs` file to see the logic that we need to implement/improve. 

I think we actually probably want to give the `fetch_group` method a param to fetch either active or non-active groups.

delcin-raj:
closes #202

### DIFF
--- a/src/whitenoise.rs
+++ b/src/whitenoise.rs
@@ -2114,22 +2114,18 @@ impl Whitenoise {
     }
 
     pub async fn fetch_groups(&self, active_filter: bool) -> Result<Vec<group_types::Group>> {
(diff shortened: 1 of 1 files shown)


delcin-raj:
Docs and tests will be done later. @erskingardner, we need to refactor into multiple modules. Sorry, my earlier suggestion is not good.

erskingardner:
> we need to refactor into multiple modules. Sorry, my earlier suggestion is not good.

I was wondering how long you'd hold out. 😂 All good - let's get started on that today.

PR #218, opened 2025-06-19
Fetch group members

### LINKED ISSUE #203: Add `whitenoise.fetch_group_members` method
You can look at the `_OLD/OLD_commands/groups/get_group_members.rs` file to see the logic that we need to implement/improve.

delcin-raj:
closes #203

### DIFF
--- a/src/whitenoise.rs
+++ b/src/whitenoise.rs
@@ -2114,22 +2114,18 @@ impl Whitenoise {
     }
 
     pub async fn fetch_groups(&self, active_filter: bool) -> Result<Vec<group_types::Group>> {
-        tracing::debug!(target: "whitenoise::get_groups", "Attempting to acquire nostr_mls lock");
         let active_account = self
             .get_active_account()
             .await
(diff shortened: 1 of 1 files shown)


src/whitenoise.rs
```diff
@@ -2112,6 +2112,42 @@ impl Whitenoise {
 
         Ok(group)
     }
+
+    pub async fn fetch_groups(&self, active_filter: bool) -> Result<Vec<group_types::Group>> {
+        let active_account = self
```
erskingardner, line 2117: This active_account thing was an older pattern that I used a lot in the old commands. I am a little undecided, but I think that we should be moving to methods that ask for the account or pubkey as a param from the user instead of relying on the internal state. 

What do you think? I mean, one one hand. we have the state, let's use it. On the other, i like the idea that consumers have to be explicit about the data they're expecting.

delcin-raj: Yes, the state requires a lot of error handling, more lines of code and testing.
It's always better for the user to be explicit, that's the functional programming way.

I would like explicit account in params.

PR #220, opened 2025-06-19
Active account state deprecated in whitenoise

### LINKED ISSUE #219: Remove active_account from Whitenoise struct
This requires doing a few things: 

- [x] Remove the member from the struct
- [x] Remove the getters/setters in the whitenoise impl
- [x] Remove any calls to those getters/setters
- [x] Update any public API method that isn't already requiring a PublicKey or Account param to require one.
- [x] Add validation to all methods that take in PublicKey or Account params to check that it's a valid, "logged-in" account

delcin-raj:
closes #219


erskingardner (approved):
I didn't go over it with a fine tooth comb (reviewing on the phone) but everything I was was good.

PR #229, opened 2025-06-24
Group tests

delcin-raj:


src/nostr_manager/query.rs
```diff
@@ -54,12 +55,19 @@ impl NostrManager {
         Ok(contacts_metadata)
     }
 
-    pub(crate) async fn query_user_key_package(&self, pubkey: PublicKey) -> Result<Option<Event>> {
+    pub(crate) async fn query_user_key_package(
```
erskingardner, line 58: I think we should keep the pattern where `query_` methods are always only querying the database instead of hitting relays. 

If we don't use this method anywhere else, let's just remove it.

delcin-raj: It's being used, we can change it to fetch_*?

erskingardner: Sure, but why have two methods then? one referencing the other?

delcin-raj: Got it

delcin-raj: But the `client` field is private in NostrManager, so we need that hop

PR #232, opened 2025-06-26
Point to the latest nostr mls changes

delcin-raj:


delcin-raj:
We can close this then

PR #235, opened 2025-06-29
Delete all does not delete the default relays for nostr client

delcin-raj:
* This enables the user to use the app without requiring them to restarting it.

### DIFF
--- a/src/whitenoise/accounts/mod.rs
+++ b/src/whitenoise/accounts/mod.rs
@@ -1249,7 +1249,7 @@ mod tests {
     use std::sync::Arc;
 
     #[tokio::test]
-    #[ignore]
+    #[ignore] // To avoid race conditions on `whitenoise` object when tests are run in parallel
     async fn test_login_after_delete_all_data() {
         let whitenoise = test_get_whitenoise().await;
 


src/whitenoise/accounts/mod.rs
```diff
@@ -1248,9 +1248,22 @@ mod tests {
     use crate::whitenoise::test_utils::*;
     use std::sync::Arc;
 
+    #[tokio::test]
+    #[ignore]
```
delcin-raj: Ignoring this as it deletes the data used by other tests.
This is happening because `Whitnoise` is a global singleton.
Individually running this test succeeds.

src/whitenoise/accounts/mod.rs
```diff
@@ -149,7 +149,7 @@ impl Account {
     ///
     /// This function does not currently return any errors, but it is fallible to allow for
     /// future error handling and to match the expected signature for account creation.
-    pub(crate) async fn new() -> core::result::Result<(Account, Keys), AccountError> {
+    pub(crate) fn new() -> core::result::Result<(Account, Keys), AccountError> {
```
delcin-raj, line 152: This function need not be async, so removing that overhead

src/whitenoise/accounts/mod.rs
erskingardner: Maybe add a short comment here explaining this in the code so we know why it's ignored in the future.

PR #237, opened 2025-06-30
Implement upload profile picture

delcin-raj:


src/bin/integration_test.rs
```diff
@@ -171,6 +171,7 @@ async fn main() -> Result<(), WhitenoiseError> {
         dark_theme: false,
         dev_mode: true,
         lockdown_mode: true,
+        blossom_profile: None,
```
erskingardner: what's that for? we shouldn't need to do this. we just want to use kind:0 metadata events to store the URL that we get back from the blossom response. 

1. user uploads image. 
2. we store on blossom.
3. we get the url back from blossom
4. we create a new metadata event for them
5. we publish that (once they hit Save in the UI).

src/nostr_manager/mod.rs
```diff
@@ -297,6 +299,18 @@ impl NostrManager {
         Ok(self.client.send_event_to(relays, &event).await?)
     }
 
+    pub(crate) async fn publish_metadata_for_account(
```
erskingardner: I thought we already had a method for updating metadata? It's in the `relay.rs` file (odd place for it).

delcin-raj: 1. Missed it because `onboard_new_account` did not use it, so I created this new one
2. I did not want to create a new file for those 2 metadata functions

Fixed both 1 and 2 in the following commit.

PR #241, opened 2025-07-02
Message aggregator

erskingardner:


src/whitenoise/message_aggregator/mod.rs
```diff
@@ -0,0 +1,137 @@
...
+            tracing::debug!(
+                "Processing {} messages for group {}",
+                messages.len(),
+                hex::encode(group_id.as_slice())
+            );
+        }
+
+        // Use the processor module to handle the actual processing
+        processor::process_messages(messages, parser, &self.config).await
+    }
```
delcin-raj, line 80: The purpose of `aggregate_messages_for_group` and `process_messages_internal` seems to add logging and at the end calls `processor::process_messages`.

Are they necessary?

delcin-raj: The logs could be added in `accounts::messages.rs` file itself

src/whitenoise/message_aggregator/processor.rs
```diff
@@ -0,0 +1,377 @@
...
+        is_reply,
+        reply_to_id,
+        is_deleted: false,
+        content_tokens,
+        reactions: Default::default(),
+    })
+}
+
+/// Extract reply information from message tags
+fn extract_reply_info(tags: &Tags) -> (bool, Option<String>) {
```
delcin-raj: The return type could simply be `Option<String>` as it's always returned as `(false, None)` and `(true, Some(_)`

delcin-raj (commented):
Nothing major, some types maybe be simplified in future.

src/whitenoise/message_aggregator/mod.rs
erskingardner: yeah, I'll remove the internal method

PR #245, opened 2025-07-03
Testing in progress

delcin-raj:


src/whitenoise/accounts/welcomes.rs
```diff
@@ -239,3 +239,53 @@ impl Whitenoise {
         Ok(())
     }
 }
+
+#[cfg(test)]
```
delcin-raj, line 243: @erskingardner run the below test and observe the logs.

PR #249, opened 2025-07-05
Tests for group related functions

delcin-raj:


erskingardner:
I think if this is about testing the nostr_group_id updates we're going to have to directly check the subscriptions that we have running on the nostr client once we perform actions on our side that should update these things. For now, we can just manually check all the spots but having tests would also be great.

erskingardner:
Is this PR still active?

PR #256, opened 2025-07-07
implement nip04 dm

delcin-raj:


src/whitenoise/accounts/messages.rs
```diff
@@ -238,4 +239,119 @@ impl Whitenoise {
...
+        let relays = self
+            .fetch_relays_with_fallback(*sender_pubkey, RelayType::Inbox)
+            .await?;
+
+        let _event_id = self
+            .nostr
+            .publish_event_builder_with_signer(dm_event_builder, &relays, sender_keys)
+            .await?;
+
+        Ok(self.nostr.parse(&content))
```
erskingardner: I think this should just return an unit `()`

PR #258, opened 2025-07-08
Test fetch welcomes

delcin-raj:


erskingardner (approved):

src/whitenoise/accounts/welcomes.rs
```diff
@@ -277,31 +273,33 @@ mod tests {
...
+        assert!(!result.is_empty(), "{:?}", result);
+        whitenoise.logout(&admin_key.public_key).await.unwrap();
 
+        tracing::info!("Logging into account {}", regular_key.public_key.to_hex());
         let account = whitenoise
             .login(regular_key.secret_key().to_secret_hex())
             .await
             .unwrap();
         // Give some time for the event processor to process welcome messages
-        // sleep(Duration::from_secs(10));
```
jgmontoya, line 304: The sleep is deleted but the comment is still there. We should delete as well to avoid confusion.

PR #266, opened 2025-07-10
Add query and fetch methods for contacts, thoroughly test query and fetch

erskingardner:
This adds distinct API method for using query (for local database cache querying) and fetch (for hitting relays) for contacts and their metadata. I've also added extensive testing and an example you can use to test real world usage with larger accounts. 

I'm quite sure that the API is returning the right metadata for the right for pubkeys.

### DIFF
--- a/src/nostr_manager/query.rs
+++ b/src/nostr_manager/query.rs
@@ -90,7 +90,10 @@ impl NostrManager {
             .author(pubkey)
             .limit(1);
 
-        let events = self.client.fetch_events(filter, self.timeout().await?).await?;
+        let events = self
+            .client
+            .fetch_events(filter, self.timeout().await?)
+            .await?;
 
         let contacts_pubkeys = if let Some(event) = events.first() {
             event
--- a/src/nostr_manager/query.rs
+++ b/src/nostr_manager/query.rs
@@ -90,7 +90,7 @@ impl NostrManager {
             .author(pubkey)
             .limit(1);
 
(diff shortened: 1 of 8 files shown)


delcin-raj:
I will test these changes later today

src/nostr_manager/query.rs
```diff
@@ -59,10 +59,73 @@ impl NostrManager {
...
+            .client
+            .fetch_events(filter, self.timeout().await?)
+            .await?;
+
+        let contacts_pubkeys = if let Some(event) = events.first() {
+            event
+                .tags
+                .iter()
+                .filter(|tag| tag.kind() == TagKind::p())
+                .filter_map(|tag| tag.content().map(|c| PublicKey::from_hex(c).unwrap()))
```
delcin-raj, line 103: This will crash the app when the public key parsing fails.

src/nostr_manager/query.rs
```diff
@@ -59,10 +59,73 @@ impl NostrManager {
...
+
+        let mut contacts_metadata = HashMap::new();
+        let meta_filter = Filter::new()
+            .kind(Kind::Metadata)
+            .authors(contacts_pubkeys.clone());
+        let meta_events = self
+            .client
+            .fetch_events(meta_filter, self.timeout().await?)
+            .await?;
+        for contact in contacts_pubkeys {
```
delcin-raj, line 121: This is running in O(n^2) and can be greatly simplified by iterating over metadata events.
There is not need to update the hashmap for contacts that don't have the metadata event, because `hashmap.get("missing_key")` returns None always.

PR #268, opened 2025-07-10
Optimise contacts metadata fetch and query

delcin-raj:


erskingardner:
I'm so curious! Is this that much faster?

erskingardner (approved):

PR #284, opened 2025-07-22
Account client

delcin-raj:
1. Remove unused structs OnboardingState, Relay, NostrSettings
2. Associate relays with accounts and contacts
3. Save the relays locally in accounts and contacts table
4. Make sure every fetch uses appropriate relays
5. `fetch_` implies from nostr, `load_` implies from sqlite, `get_` implies from memory
6. Remove dummy/logic tests that don't test a function

1. Process all subscription events.
2. Investigate subscription timeout warnings


src/whitenoise/accounts/messages.rs
jgmontoya, line 1: The ci/tests are failing because the doc examples are outdated. The fix is really small, you can cherrypick commit ef51cb7d2a5ded33f150656d7813313430cff5b1

db_migrations/0001_accounts.sql
jgmontoya, line 1: I think changing the schema like this, via modifying the migration will break for existing users (users who already had the previous 'version' of this migration applied).

This might be ok if we're fine with forcing everyone to 'start fresh' which given how early we're with everything might be good enough.

The backwards-compatible way to do it I believe would be to add a new migration file that drops the onboarding column and adds the new columns to the accounts table.

examples/onboarding_workflow.rs
jgmontoya, line 1: I'm a bit confused, why delete this file instead of update it to check the relays directly (instead of through `onboarding`)?? Or does it not make sense at all anymore?

src/nostr_manager/mod.rs
```diff
@@ -40,60 +39,43 @@ pub enum NostrManagerError {
...
         }
+        Ok(())
     }
-}
-pub type Result<T> = std::result::Result<T, NostrManagerError>;
 
-impl NostrManager {
+    pub(crate) fn default_timeout() -> Duration {
+        Duration::from_secs(5)
+    }
```
jgmontoya, line 67: Should we perhaps make this a constant rather than a function?

src/nostr_manager/mod.rs
```diff
@@ -40,60 +39,43 @@ pub enum NostrManagerError {
...
     /// Create a new Nostr manager
     ///
     /// # Arguments
     ///
     /// * `db_path` - The path to the nostr cache database
     /// * `event_sender` - Channel sender for forwarding events to Whitenoise for processing
     /// * `connect_to_relays` - Whether to attempt connecting to relays (false for testing)
```
jgmontoya: Please update the docstrings to match the new implementation

src/nostr_manager/mod.rs
```diff
@@ -860,195 +723,6 @@ impl NostrManager {
...
-        let (tx, _rx) = mpsc::channel(10);
-
-        let manager = NostrManager::new_without_connection(db_path, tx)
-            .await
-            .unwrap();
-
-        // Test that delete_all_data succeeds
-        let result = manager.delete_all_data().await;
-        assert!(result.is_ok());
-    }
```
jgmontoya, line 996: Are all of these tests really redundant/not useful now??

src/whitenoise/accounts/contacts.rs
```diff
@@ -122,20 +203,47 @@ impl Whitenoise {
...
+            key_package_relays = account.key_package_relays.clone();
+        }
+        let metadata = self
+            .fetch_metadata_from(account.discovery_relays.clone(), contact_pubkey)
+            .await?;
+
+        // save contact locally
+        let contact = Contact {
+            pubkey: contact_pubkey,
+            discovery_relays: account.discovery_relays.clone(), // The account discovered the contact through this relays
```
jgmontoya, line 206: Does this mean that all of our contacts will have discovery_relays equal to our own discovery_relays?

src/whitenoise/accounts/groups.rs
```diff
@@ -78,10 +77,16 @@ impl Whitenoise {
             .map(|g| hex::encode(g.nostr_group_id))
             .collect::<Vec<_>>();
 
+        if welcome_rumors.len() != contacts.len() {
+            return Err(WhitenoiseError::Other(anyhow::Error::msg(
+                "Welcome rumours are missing for some of the members",
+            )));
+        }
```
jgmontoya, line 77: Should this check be before the `let group_ids = nostr_mls` line? Just so we 'bail out' early. I'd move the group_ids thing towards the end of the function closer to where it's actually used.

src/whitenoise/accounts/groups.rs
```diff
@@ -100,10 +105,6 @@ impl Whitenoise {
                     "No public key found in key package event"
                 )))?;
 
-            let member_inbox_relays = self
-                .fetch_relays_with_fallback(member_pubkey, RelayType::Inbox)
-                .await?;
```
jgmontoya, line 105: I might be getting something wrong but I think before, this guaranteed that we were using the most up-to-date inbox relay for our contacts, but now we're going to use whatever inbox relay we have stored.

Is there a mechanism by which we keep that information relatively updated?

src/whitenoise/accounts/mod.rs
```diff
@@ -113,10 +111,34 @@ where
...
+            .map_err(|e| sqlx::Error::ColumnDecode {
+                index: "inbox_relays".to_owned(),
+                source: Box::new(e),
+            })?;
+        let key_package_relays: Vec<RelayUrl> = serde_json::from_str(&key_package_relays)
+            .map(|urls: Vec<String>| {
+                urls.iter()
+                    .filter_map(|url| RelayUrl::parse(url).ok())
+                    .collect::<Vec<_>>()
+            })
```
jgmontoya: We could probably extract the parsing logic to a function to avoid a bit of repetition

jgmontoya (commented):
Left a couple of comments, suggestions and questions!


> fetch_ implies from nostr, load_ implies from sqlite, get_ implies from memory

I really like this convention!

db_migrations/0001_accounts.sql
delcin-raj: Yeah, I did this under the assumption that we still don't have active users.

delcin-raj: Once we release a stable version, we have to be careful with migrations

examples/onboarding_workflow.rs
delcin-raj: OnboardingState struct did not make any sense, as we have to regularly refresh by publishing keypackage events.

When an account is created it's keypackage will be published. And has to be refreshed periodically.

src/nostr_manager/mod.rs
delcin-raj: I don't think so, because for different fetch events, different timeouts will be necessary in the future.
This is simply a placeholder for now.

src/nostr_manager/mod.rs
delcin-raj: I removed NostrManagerSettings struct as it did not make sense to have a global nostr relays set.
So corresponding tests are removed.

I shouldn't have removed test delete all data I guess. Let me look into this.

src/whitenoise/accounts/contacts.rs
delcin-raj: No, the idea here is that we discovered those contacts through our discovery relays.

So a subset of our discovery relays should be a subset of that contacts discovery relays.

Unless the contact discloses us the full set, we can never know.

src/whitenoise/accounts/groups.rs
delcin-raj: The functions to update the existing relays will be added.

src/nostr_manager/mod.rs
delcin-raj: So the delete all data was not doing anything after the changes made to how the nostr object was created from whitenoise.

db_migrations/0001_accounts.sql
erskingardner: I think this unfortunately wasn't a good assumption. We do have active users now. Blowing away their databases on a new version release isn't something we can do. 

Since this is already merged I think we're going to have to go back and make a change to migrations to make it upgrade for other users and all of us are going to have to start with fresh dev envs instead.

src/whitenoise/accounts/contacts.rs
```diff
@@ -34,13 +88,13 @@ impl Whitenoise {
     /// Returns a `WhitenoiseError` if the contact list query fails.
     pub async fn fetch_contacts(
         &self,
-        pubkey: PublicKey,
+        account: &Account,
```
erskingardner, line 91: what about when we want to fetch the contacts of a contact that isn't an account in our system. For example, when we want to get the contacts of one of your contacts so we can see who you both follow in common?

delcin-raj: That information should be private.
I think we need to publish the contact list in relays in an encrypted form.

erskingardner: No - contact lists are almost always published in public form. There is a way to do encrypted contacts (you just use NIP-44 encryption in the `content` field of the `kind:3` event with a json array of the private contacts) but we don't need to do that right now.

PR #287, opened 2025-07-28
Fix comments

delcin-raj:
In this PR following tasks are completed.

1. @jgmontoya 's concern about migrations deleting existing user data fixed.
2. @erskingardner comment about having the ability to find mutual contacts
3. rename discovery_relays as nip65_relays for clarity
4. When the user tries to login with their existing nostr account, their preferred relay list is used

### DIFF
--- a/db_migrations/0004_accounts_relays.sql
+++ b/db_migrations/0004_accounts_relays.sql
@@ -1,0 +1,35 @@
+-- Up Migration
+BEGIN;
+
+-- Step 1: Create a new temporary table with the updated schema
+CREATE TABLE accounts_new (
+    pubkey TEXT PRIMARY KEY,
+    settings JSONB NOT NULL,
+    discovery_relays TEXT NOT NULL,
+    inbox_relays TEXT NOT NULL,
+    key_package_relays TEXT NOT NULL,
+    last_synced INTEGER NOT NULL
+);
+
+-- Step 2: Copy data from old table to new table, with defaults for new columns
+INSERT INTO accounts_new (
+    pubkey, settings,
+    discovery_relays, inbox_relays, key_package_relays,
+    last_synced
+)
+SELECT
+    pubkey,
+    settings,
+    '[]',  -- default value for discovery_relays
+    '[]',  -- default value for inbox_relays
+    '[]',  -- default value for key_package_relays
+    last_synced
+FROM accounts;
+
+-- Step 3: Drop the old table
+DROP TABLE accounts;
+
+-- Step 4: Rename the new table to the original name
+ALTER TABLE accounts_new RENAME TO accounts;
+
+COMMIT;
--- a/db_migrations/0004_accounts_relays.sql
+++ b/db_migrations/0004_accounts_relays.sql
@@ -1,11 +1,8 @@
--- Up Migration
-BEGIN;
-
 -- Step 1: Create a new temporary table with the updated schema
 CREATE TABLE accounts_new (
     pubkey TEXT PRIMARY KEY,
     settings JSONB NOT NULL,
-    discovery_relays TEXT NOT NULL,
+    nip65_relays TEXT NOT NULL,
     inbox_relays TEXT NOT NULL,
     key_package_relays TEXT NOT NULL,
     last_synced INTEGER NOT NULL
@@ -14,13 +11,13 @@ CREATE TABLE accounts_new (
 -- Step 2: Copy data from old table to new table, with defaults for new columns
 INSERT INTO accounts_new (
     pubkey, settings,
-    discovery_relays, inbox_relays, key_package_relays,
+    nip65_relays, inbox_relays, key_package_relays,
     last_synced
 )
 SELECT
     pubkey,
     settings,
-    '[]',  -- default value for discovery_relays
+    '[]',  -- default value for nip65_relays
     '[]',  -- default value for inbox_relays
     '[]',  -- default value for key_package_relays
     last_synced
@@ -31,5 +28,3 @@ DROP TABLE accounts;
 
 -- Step 4: Rename the new table to the original name
 ALTER TABLE accounts_new RENAME TO accounts;
-
-COMMIT;
--- a/message-aggr-plan.md
+++ b/message-aggr-plan.md
@@ -497,7 +497,7 @@ impl Whitenoise {
     ) -> Result<Vec<ChatMessage>> {
         // Get account to access nostr_mls instance
         let account = self.fetch_account(pubkey).await?;
-        let nostr_mls_guard = account.nostr_mls.lock().await;
+        let nostr_mls_guard = account.nostr_mls.lock().unwrap();
         
         if let Some(nostr_mls) = nostr_mls_guard.as_ref() {
             // Use the aggregator to handle the complete pipeline
--- a/src/whitenoise/accounts/contacts.rs
+++ b/src/whitenoise/accounts/contacts.rs
@@ -90,10 +90,6 @@ impl Whitenoise {
         &self,
         pubkey: &PublicKey,
     ) -> Result<HashMap<PublicKey, Option<Metadata>>> {
-        if !self.logged_in(pubkey).await {
-            return Err(WhitenoiseError::AccountNotFound);
-        }
-
         let account = self.get_account(pubkey).await?;
         let contacts = self.nostr.fetch_user_contact_list(&account).await?;
         Ok(contacts)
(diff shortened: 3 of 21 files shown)


src/whitenoise/database.rs
jgmontoya, line 1: I think we can improve/simplify the way we're dealing with migrations, what do you think of something like the following:

```rs
use sqlx::{migrate::Migrator, sqlite::SqlitePoolOptions, MigrateDatabase, Sqlite, SqlitePool};
use std::path::PathBuf;
use std::time::Duration;
use thiserror::Error;
use once_cell::sync::Lazy;

pub static MIGRATOR: Lazy<Migrator> = Lazy::new(|| sqlx::migrate!("./db_migrations"));

#[derive(Error, Debug)]
pub enum DatabaseError {
    #[error("SQLx error: {0}")]
    Sqlx(#[from] sqlx::Error),
    #[error("Migration error: {0}")]
    Migration(#[from] sqlx::migrate::MigrateError),
    #[error("File system error: {0}")]
    FileSystem(#[from] std::io::Error),
}

#[derive(Clone, Debug)]
pub struct Database {
    pub pool: SqlitePool,
    pub path: PathBuf,
    pub last_connected: std::time::SystemTime,
}

impl Database {
    pub async fn new(db_path: PathBuf) -> Result<Self, DatabaseError> {
        if let Some(parent) = db_path.parent() {
            std::fs::create_dir_all(parent)?;
        }

        let db_url = format!("sqlite://{}", db_path.display());

        if !Sqlite::database_exists(&db_url).await.unwrap_or(false) {
            Sqlite::create_database(&db_url).await?;
            tracing::debug!("Database created at {}", db_url);
        }

        let pool = SqlitePoolOptions::new()
            .acquire_timeout(Duration::from_secs(5))
            .max_connections(10)
            .after_connect(|conn, _| {
                Box::pin(async move {
                    let conn = &mut *conn;
                    sqlx::query("PRAGMA journal_mode=WAL").execute(conn).await?;
                    sqlx::query("PRAGMA busy_timeout=5000").execute(conn).await?;
                    sqlx::query("PRAGMA foreign_keys=ON").execute(conn).await?;
                    sqlx::query("PRAGMA recursive_triggers=ON").execute(conn).await?;

                    Ok(())
                })
            })
            .connect(&format!("{}?mode=rwc", db_url))
            .await?;

        // Automatically run migrations
        MIGRATOR.run(&pool).await?;

        Ok(Self {
            pool,
            path: db_path,
            last_connected: std::time::SystemTime::now(),
        })
    }

    /// Run all pending migrations
    pub async fn migrate_up(&self) -> Result<(), DatabaseError> {
        MIGRATOR.run(&self.pool).await?;
        Ok(())
    }

    /// Revert the last migration (works only if `-- migrate:down` is defined)
    pub async fn migrate_down(&self) -> Result<(), DatabaseError> {
        MIGRATOR.revert(&self.pool).await?;
        Ok(())
    }

    /// Deletes all data while disabling foreign key checks
    pub async fn delete_all_data(&self) -> Result<(), DatabaseError> {
        let mut txn = self.pool.begin().await?;

        sqlx::query("PRAGMA foreign_keys=OFF").execute(&mut *txn).await?;
        sqlx::query("DELETE FROM media_files").execute(&mut *txn).await?;
        sqlx::query("DELETE FROM accounts").execute(&mut *txn).await?;
        sqlx::query("DELETE FROM contacts").execute(&mut *txn).await?;
        sqlx::query("PRAGMA foreign_keys=ON").execute(&mut *txn).await?;

        txn.commit().await?;
        Ok(())
    }
}
```

Also, I added the DELETE FROM contacts which I believe is missing

db_migrations/0004_accounts_relays.sql
```diff
@@ -0,0 +1,30 @@
...
+    '[]',  -- default value for inbox_relays
+    '[]',  -- default value for key_package_relays
+    last_synced
+FROM accounts;
+
+-- Step 3: Drop the old table
+DROP TABLE accounts;
+
+-- Step 4: Rename the new table to the original name
+ALTER TABLE accounts_new RENAME TO accounts;
```
jgmontoya, line 30: I think it's simpler to modify the table directly, also I believe it's a good idea to specify the down migrations for new migrations as well just in case we need them in the future:

```sql
-- migrate:up
ALTER TABLE accounts
  ADD COLUMN nip65_relays TEXT NOT NULL DEFAULT '[]';
ALTER TABLE accounts
  ADD COLUMN inbox_relays TEXT NOT NULL DEFAULT '[]';
ALTER TABLE accounts
  ADD COLUMN key_package_relays TEXT NOT NULL DEFAULT '[]';

ALTER TABLE accounts DROP COLUMN onboarding;

-- migrate:down
ALTER TABLE accounts DROP COLUMN nip65_relays;
ALTER TABLE accounts DROP COLUMN inbox_relays;
ALTER TABLE accounts DROP COLUMN key_package_relays;

ALTER TABLE accounts
  ADD COLUMN onboarding JSONB NOT NULL DEFAULT '{}';
```

delcin-raj: Thank you, proof that domain knowledge >> AI generated

erskingardner: Definitely agree with @jgmontoya here. ALTER is way better.

src/whitenoise/accounts/contacts.rs
```diff
@@ -88,13 +88,14 @@ impl Whitenoise {
     /// Returns a `WhitenoiseError` if the contact list query fails.
     pub async fn fetch_contacts(
         &self,
-        account: &Account,
+        pubkey: &PublicKey,
     ) -> Result<HashMap<PublicKey, Option<Metadata>>> {
-        if !self.logged_in(&account.pubkey).await {
+        if !self.logged_in(pubkey).await {
```
erskingardner: If we want to be able to use this method for fetching data about other users (other than the logged in account), then this conditional will have to be removed.

src/whitenoise/accounts/contacts.rs
```diff
@@ -197,13 +198,23 @@ impl Whitenoise {
...
+            .fetch_metadata_from(account.nip65_relays.clone(), contact_pubkey)
             .await?;
+        let mut nip65_relays = self
+            .fetch_relays_from(
+                account.nip65_relays.clone(),
+                contact_pubkey,
+                RelayType::Nostr,
+            )
+            .await?;
+        if nip65_relays.is_empty() {
```
erskingardner: I think we probably want to leave this blank if the user we're fetching for doesn't have a 10002 relay list. We should handle falling back to our connected relays in a different place instead of pretending that the user has a relay list. (we should only publish a relay list for a user when they explicitly ask us to by making changes)

src/whitenoise/accounts/mod.rs
```diff
@@ -574,12 +576,19 @@ impl Whitenoise {
         })?;
         tracing::debug!(target: "whitenoise::add_account_from_keys", "Keys stored in secret store");
 
+        let mut nip65_relays = self
+            .fetch_relays_from(Account::default_relays(), keys.public_key, RelayType::Nostr)
+            .await?;
+        if nip65_relays.is_empty() {
```
erskingardner, line 587: Again here. we shouldn't default relays onto a user that doesn't have them if they are a pre-existing user.

src/whitenoise/accounts/mod.rs
```diff
@@ -625,7 +634,7 @@ impl Whitenoise {
         let mut txn = self.database.pool.begin().await?;
 
         let discovery_urls: Vec<_> = account
```
erskingardner: variable name should be `nip65_urls`

src/whitenoise/accounts/relays.rs
```diff
@@ -132,4 +132,25 @@ impl Whitenoise {
 
         Ok(relay_statuses)
     }
+
+    pub(crate) async fn connect_account_relays(&self, account: &Account) -> Result<()> {
```
erskingardner, line 136: I think we need to spend some time to make sure that we're all on the same page about how relays are supposed to work and when you'd need to be connected to a given relay. 

In this case, we only need to connect to the user's nip65, and inbox relays in a continuous way (I think) - the key package relay we publish to but we only need to connect to it, publish, and then disconnect from it as the user themselves. For other user's key package and inbox relays, we'll need to connect fetch/publish, then disconnect to those relays while performing certain ops. 

The main gist is that we don't want to be maintaining connections to dozen's of relays constantly. So ensuring we come up with a system to improve this will be a big deal.

For this PR, let's just leave this bit of code alone but I wanted to make sure we're all clear that relay management is becoming top of mind for me and we need to comprehensively think it through and design a system that is going to do the right thing.

src/whitenoise/database.rs
erskingardner: really like this - we need to get super consistent with how we handle migrations.

erskingardner (changes requested):
Left some actionable comments and a few broader thoughts.

src/whitenoise/accounts/relays.rs
delcin-raj: Agree

src/whitenoise/accounts/mod.rs
delcin-raj: Yes, this is happening only during a fresh login by the user.
For pre-existing users, the account is directly loaded from the database.

src/whitenoise/database.rs
delcin-raj: @erskingardner I think this could be done in a different PR, after merging this one.

message-aggr-plan.md
```diff
@@ -497,7 +497,7 @@ impl Whitenoise {
     ) -> Result<Vec<ChatMessage>> {
         // Get account to access nostr_mls instance
         let account = self.fetch_account(pubkey).await?;
-        let nostr_mls_guard = account.nostr_mls.lock().await;
+        let nostr_mls_guard = account.nostr_mls.lock().unwrap();
```
erskingardner, line 500: We don't really want to use unwrap anywhere do we?

delcin-raj: https://doc.rust-lang.org/std/sync/struct.Mutex.html

In this case, it's okay. The justification i that, once the lock is acquired the thread is not expected to fail

erskingardner:
Go ahead and merge. I've got bad internet here and haven't had a chance to properly look over but most of my earlier issues were small.

delcin-raj:
I've fixed them

PR #289, opened 2025-08-01
Add functions to update relay connections

delcin-raj:
Functions to add and remove relays
    
    1. Update the client connections for nip65 and inbox relays
    2. Publish relay list for key package and inbox relays
    3. Save the account relays locally

### DIFF
--- a/src/whitenoise/accounts/relays.rs
+++ b/src/whitenoise/accounts/relays.rs
@@ -411,13 +411,13 @@ mod tests {
     use nostr::types::RelayUrl;
     use tokio::time::sleep;
 
-    use crate::whitenoise::test_utils::test_get_whitenoise;
+    use crate::whitenoise::test_utils::create_mock_whitenoise;
     use crate::{RelayType, Whitenoise};
 
     #[tokio::test]
     async fn test_add_remove_relay() {
         let l7777 = RelayUrl::parse("ws://localhost:7777").unwrap();
-        let whitenoise = test_get_whitenoise().await;
+        let (whitenoise, _, _) = create_mock_whitenoise().await;
         let account = whitenoise.create_identity().await.unwrap();
 
         tracing::info!("NOTE: Identitiy created");
@@ -449,7 +449,7 @@ mod tests {
             .unwrap();
         assert!(
             Whitenoise::relayurl_dashset_eq(relay_list.clone(), account.nip65_relays.clone()),
-            "{relay_list:?}"
+            "RELAY ======= {relay_list:?}"
         );
 
         tracing::info!("Checking add nip65 relay");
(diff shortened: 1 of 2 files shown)


delcin-raj:
Currently, nip65 relays are not published, because when a user adds a nip65 relay, they might've their nostr relay list stored in that relay.
How to handle that?
Case 1: Discard other relays that are currently connected and make a connection to that relay list
Case 2: Keep the current relays, include the new relays, and update the relay list by publishing the extra relays as well

We have to take one of the two approaches mentioned above implicitly.
Both have use cases according to the user.
We can take the correct approach only via the user's consent.

src/nostr_manager/mod.rs
```diff
@@ -62,6 +63,26 @@ impl NostrManager {
         Ok(())
     }
 
+    pub(crate) async fn connect_to_relay(&self, relay: RelayUrl) -> Result<()> {
+        let newly_added = self.client.add_relay(relay).await?;
```
erskingardner, line 67: We haven't done this so far - but we should respect the read/write settings of the relays. 

totally fine if we want to wait and do this in a follow up PR though

src/whitenoise/mod.rs
```diff
@@ -704,7 +704,7 @@ pub mod test_utils {
                 .nostr
                 .publish_event_builder_with_signer(
                     key_package_event_builder,
-                    &account.nip65_relays,
+                    account.nip65_relays.clone(),
```
erskingardner: I know this is a test but we should make sure that we're doing things accurately and publish key packages to the account's key package relays, not all their nip65 relays.

src/whitenoise/accounts/relays.rs
```diff
@@ -152,4 +153,131 @@ impl Whitenoise {
...
+            .collect();
+
+        let event = EventBuilder::new(relay_event_kind, "").tags(tags);
+        let keys = self
+            .secrets_store
+            .get_nostr_keys_for_pubkey(&account.pubkey)?;
+
+        let result = self
+            .nostr
+            .publish_event_builder_with_signer(event.clone(), relays_to_use, keys)
```
erskingardner, line 330: These lists need to be published always to the user's nip65 relays (that are marked as write or read/write - i.e. they don't have a third positional marker). 

inbox and key package relay lists need to be distributed widely, not just to themselves.

erskingardner (changes requested):

src/nostr_manager/mod.rs
delcin-raj: Will do that in the follow up

delcin-raj: https://github.com/parres-hq/whitenoise/issues/290

erskingardner (approved):

Issue #288, opened 2025-08-01
Enhance DB migration


delcin-raj:
Migration file 0004*.sql could be improved as @jgmontoya noted.
Add tests in contacts table, test_delete_all etc.,

PR #293, opened 2025-08-06
feat(src/whitenoise/database): improve the way migrations are handled

jgmontoya:
stop requiring migrations to be hard-coded on the database module

Closes https://github.com/parres-hq/whitenoise/issues/288


src/whitenoise/database.rs
```diff
@@ -1,49 +1,36 @@
...
-    ),
-    (
-        "0004_accounts_relays.sql",
-        include_bytes!("../../db_migrations/0004_accounts_relays.sql"),
-    ),
-    // Add new migrations here in order, for example:
-    // ("000X_something.sql", include_bytes!("../db_migrations/000X_something.sql")),
-    // ("000Y_another.sql", include_bytes!("../db_migrations/000Y_another.sql")),
-];
+pub static MIGRATOR: Lazy<Migrator> = Lazy::new(|| sqlx::migrate!("./db_migrations"));
```
delcin-raj: https://doc.rust-lang.org/std/cell/struct.LazyCell.html would solve the same without additional dependency.
I think `once_cell` was not maintained actively, as pointed out by @erskingardner on my earlier PR.

jgmontoya: I'm not sure about it not being actively maintained, however I agree that we shouldn't add dependencies for things we can do with the standard library. I'm changing it to use this: https://doc.rust-lang.org/std/sync/struct.LazyLock.html

PR #294, opened 2025-08-06
Re-add publish of metadata on account creation

erskingardner:


josefinalliende (approved):
✅  LGTM!

src/whitenoise/accounts/mod.rs
```diff
@@ -252,6 +252,33 @@ impl Whitenoise {
...
+                match chars.next() {
+                    None => String::new(),
+                    Some(first_char) => {
+                        let first_upper = first_char.to_uppercase().collect::<String>();
+                        first_upper + chars.as_str()
+                    }
+                }
+            })
+            .collect::<Vec<String>>()
+            .join(" ");
```
jgmontoya: How about this?

```rs
/// Capitalizes the first letter of a word, leaving the rest unchanged
fn capitalize_first_letter(word: &str) -> String {
    let mut chars = word.chars();
    match chars.next() {
        None => String::new(),
        Some(first) => first.to_uppercase().collect::<String>() + chars.as_str(),
    }
}
```

then here it becomes just:

```rs
// Generate a petname for the account (two words, separated by a space)
let petname = petname::petname(2, " ")
    .unwrap_or_else(|| "Anonymous User".to_string())
    .split_whitespace()
    .map(capitalize_first_letter)
    .collect::<Vec<_>>()
    .join(" ");
```

jgmontoya (approved):

PR #296, opened 2025-08-07
Refactor relay methods

erskingardner:
## Summary by CodeRabbit

* **New Features**
  * Introduced unified methods for adding and removing relays across all relay types, allowing more granular relay management for accounts.

* **Improvements**
  * Enhanced relay connection handling to ensure connections before publishing events.
  * Improved fallback logic for relay usage during group member addition and message publishing.
  * Streamlined account creation and relay assignment in test scenarios.
  * Simplified event fetching logic to improve performance and reliability.
  * Conditional publishing of relay info based on relay presence during login.
  * Unified relay list publishing to support updates on removed relays.
  * Updated test utilities to use consistent identity creation methods and public keys.

* **Bug Fixes**
  * Corrected relay URL formatting in integration tests.
  * Improved test reliability by adding delays and clearer logs during group creation.

* **Tests**
  * Expanded and refactored test coverage for relay management and account creation workflows.

### DIFF
--- a/src/nostr_manager/query.rs
+++ b/src/nostr_manager/query.rs
@@ -36,19 +36,19 @@ impl NostrManager {
         relay_type: RelayType,
         nip65_relays: DashSet<RelayUrl>,
     ) -> Result<DashSet<RelayUrl>> {
-        let filter = Filter::new().author(pubkey).kind(relay_type.into());
+        let filter = Filter::new()
+            .author(pubkey)
+            .kind(relay_type.into())
+            .limit(1);
         let relay_events = self
             .client
             .fetch_events_from(nip65_relays, filter.clone(), self.timeout)
             .await?;
         tracing::debug!("Fetched relay events {:?}", relay_events);
 
-        match relay_events
-            .into_iter()
-            .max_by_key(|event| event.created_at)
-        {
+        match relay_events.first() {
             None => Ok(DashSet::new()),
-            Some(event) => Ok(Self::relay_urls_from_event(event)),
+            Some(event) => Ok(Self::relay_urls_from_event(event.clone())),
         }
     }
 
--- a/src/whitenoise/accounts/groups.rs
+++ b/src/whitenoise/accounts/groups.rs
@@ -254,8 +254,13 @@ impl Whitenoise {
         // Fetch key packages for all members
         for pk in members.iter() {
             let contact = self.load_contact(pk).await?;
+            let relays_to_use = if contact.key_package_relays.is_empty() {
+                Account::default_relays()
+            } else {
+                contact.key_package_relays.clone()
+            };
             let some_event = self
-                .fetch_key_package_event_from(contact.key_package_relays.clone(), *pk)
+                .fetch_key_package_event_from(relays_to_use, *pk)
                 .await?;
             let event = some_event.ok_or(WhitenoiseError::NostrMlsError(
                 nostr_mls::Error::KeyPackage("Does not exist".to_owned()),
@@ -300,52 +305,70 @@ impl Whitenoise {
             )));
         }
 
-        let result = self
-            .nostr
-            .publish_event_to(evolution_event, &group_relays)
-            .await;
-
-        match result {
-            Ok(_event_id) => {
-                // Evolution event published successfully
-                // Fan out the welcome message to all members
-                for (welcome_rumor, contact) in welcome_rumors.iter().zip(contacts) {
-                    // Get the public key of the member from the key package event
-                    let key_package_event_id =
-                        welcome_rumor
-                            .tags
-                            .event_ids()
-                            .next()
-                            .ok_or(WhitenoiseError::Other(anyhow::anyhow!(
-                                "No event ID found in welcome rumor"
-                            )))?;
-
-                    let member_pubkey = key_package_events
-                        .iter()
-                        .find(|event| event.id == *key_package_event_id)
-                        .map(|event| event.pubkey)
-                        .ok_or(WhitenoiseError::Other(anyhow::anyhow!(
-                            "No public key found in key package event"
-                        )))?;
-
-                    // Create a timestamp 1 month in the future
-                    use std::ops::Add;
-                    let one_month_future = Timestamp::now().add(30 * 24 * 60 * 60);
-                    self.nostr
-                        .publish_gift_wrap_with_signer(
-                            &member_pubkey,
-                            welcome_rumor.clone(),
-                            vec![Tag::expiration(one_month_future)],
-                            contact.inbox_relays,
-                            keys.clone(),
-                        )
-                        .await
-                        .map_err(WhitenoiseError::from)?;
-                }
+        // Check if we have any relays to publish to and publish the evolution event
+        if group_relays.is_empty() {
+            tracing::warn!(
+                target: "whitenoise::add_members_to_group",
+                "Group has no relays configured, using account's default relays"
+            );
+            // Use the account's default relays as fallback
+            let fallback_relays: std::collections::BTreeSet<RelayUrl> = account.nip65_relays.iter().map(|relay_ref| relay_ref.clone()).collect();
+            if fallback_relays.is_empty() {
+                return Err(WhitenoiseError::Other(anyhow::anyhow!(
+                    "No relays available for publishing evolution event - both group relays and account relays are empty"
+                )));
             }
-            Err(e) => {
-                return Err(WhitenoiseError::NostrManager(e));
-            }
+            self.nostr
+                .publish_event_to(evolution_event, &fallback_relays)
+                .await?;
+        } else {
+            self.nostr
+                .publish_event_to(evolution_event, &group_relays)
+                .await?;
+        }
+
+        // Evolution event published successfully
+        // Fan out the welcome message to all members
+        for (welcome_rumor, contact) in welcome_rumors.iter().zip(contacts) {
+            // Get the public key of the member from the key package event
+            let key_package_event_id =
+                welcome_rumor
+                    .tags
+                    .event_ids()
+                    .next()
+                    .ok_or(WhitenoiseError::Other(anyhow::anyhow!(
+                        "No event ID found in welcome rumor"
+                    )))?;
+
+            let member_pubkey = key_package_events
+                .iter()
+                .find(|event| event.id == *key_package_event_id)
+                .map(|event| event.pubkey)
+                .ok_or(WhitenoiseError::Other(anyhow::anyhow!(
+                    "No public key found in key package event"
+                )))?;
+
+            // Create a timestamp 1 month in the future
+            use std::ops::Add;
+            let one_month_future = Timestamp::now().add(30 * 24 * 60 * 60);
+
+            // Use fallback relays if contact has no inbox relays configured
+            let relays_to_use = if contact.inbox_relays.is_empty() {
+                Account::default_relays()
+            } else {
+                contact.inbox_relays
+            };
+
+            self.nostr
+                .publish_gift_wrap_with_signer(
+                    &member_pubkey,
+                    welcome_rumor.clone(),
+                    vec![Tag::expiration(one_month_future)],
+                    relays_to_use,
+                    keys.clone(),
+                )
+                .await
+                .map_err(WhitenoiseError::from)?;
         }
 
         Ok(())
--- a/src/whitenoise/accounts/groups.rs
+++ b/src/whitenoise/accounts/groups.rs
@@ -420,22 +420,28 @@ mod tests {
 
     #[tokio::test]
     async fn test_create_group() {
-        let whitenoise = test_get_whitenoise().await;
+        let (whitenoise, _data_temp, _logs_temp) = create_mock_whitenoise().await;
 
         // Setup creator account
-        let (creator_account, _creator_keys) = setup_login_account(whitenoise).await;
+        let creator_account = whitenoise.create_identity().await.unwrap();
 
         // Setup member accounts
-        let member_accounts = setup_multiple_test_accounts(whitenoise, &creator_account, 2).await;
-        let member_pubkeys: Vec<PublicKey> =
-            member_accounts.iter().map(|(acc, _)| acc.pubkey).collect();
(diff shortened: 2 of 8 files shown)


src/whitenoise/accounts/relays.rs
```diff
@@ -154,152 +154,124 @@ impl Whitenoise {
...
-        )
-        .await?;
-        self.publish_relay_list_for_account(account, RelayType::Inbox)
+        // Update the in-memory account
+        {
+            let mut accounts = self.write_accounts().await;
+            accounts.insert(account.pubkey, account.clone());
+        }
+
+        // TODO: Do we need to manually disconnect or just wait for next session when we won't connect?
```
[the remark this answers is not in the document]
erskingardner: Not going to address this for now.

src/nostr_manager/query.rs
```diff
@@ -36,19 +36,19 @@ impl NostrManager {
...
         tracing::debug!("Fetched relay events {:?}", relay_events);
 
-        match relay_events
-            .into_iter()
-            .max_by_key(|event| event.created_at)
-        {
+        match relay_events.first() {
             None => Ok(DashSet::new()),
-            Some(event) => Ok(Self::relay_urls_from_event(event)),
+            Some(event) => Ok(Self::relay_urls_from_event(event.clone())),
```
jgmontoya, line 51: Before we got all the relay events, got the last one (`.max_by_key(|event| event.created_at)`) and returned based on that. Now it seems like we're getting just any event? Is that alright?

src/whitenoise/accounts/groups.rs
```diff
@@ -300,52 +305,74 @@ impl Whitenoise {
...
+                .await?;
+        } else {
+            self.nostr
+                .publish_event_to(evolution_event, &group_relays)
+                .await?;
+        }
+
+        // Evolution event published successfully
+        // Fan out the welcome message to all members
+        for (welcome_rumor, contact) in welcome_rumors.iter().zip(contacts) {
```
jgmontoya, line 337: Just to verify: we're sure that the welcome_rumors are in the same order as the contacts, correct?

src/whitenoise/accounts/groups.rs
```diff
@@ -300,52 +305,74 @@ impl Whitenoise {
...
+                .iter()
+                .find(|event| event.id == *key_package_event_id)
+                .map(|event| event.pubkey)
+                .ok_or(WhitenoiseError::Other(anyhow::anyhow!(
+                    "No public key found in key package event"
+                )))?;
+
+            // Create a timestamp 1 month in the future
+            use std::ops::Add;
+            let one_month_future = Timestamp::now().add(30 * 24 * 60 * 60);
```
jgmontoya: [Nit] I think using Duration would be a bit clearer/more explicit:

```rs
use std::time::Duration;
let one_month_future = Timestamp::now() + Duration::from_secs(30 * 24 * 60 * 60);
```

src/whitenoise/accounts/groups.rs
```diff
@@ -300,52 +305,74 @@ impl Whitenoise {
...
+
+            // Use fallback relays if contact has no inbox relays configured
+            let relays_to_use = if contact.inbox_relays.is_empty() {
+                Account::default_relays()
+            } else {
+                contact.inbox_relays
+            };
```
jgmontoya, line 364: For the future: I think this fallback logic is going to be used a lot in a bunch of places, perhaps we can encapsulate it into a method. Something like:

On `src/whitenoise/accounts/contacts.rs`:
```rs
/// Returns the relays to use for this contact of the specified type, falling back to default relays if none are set.
pub fn get_relays_of_type(&self, relay_type: RelayType) -> DashSet<RelayUrl> {
    let relays = match relay_type {
        RelayType::Nostr => &self.nip65_relays,
        RelayType::Inbox => &self.inbox_relays,
        RelayType::KeyPackage => &self.key_package_relays,
    };

    if relays.is_empty() {
        Account::default_relays()
    } else {
        relays.clone()
    }
}
```

src/whitenoise/accounts/mod.rs
```diff
@@ -634,11 +638,31 @@ impl Whitenoise {
         })?;
         tracing::debug!(target: "whitenoise::add_account_from_keys", "Keys stored in secret store");
 
+        let mut need_to_publish_nip65_relays = false;
+        let mut need_to_publish_inbox_relays = false;
+        let mut need_to_publish_key_package_relays = false;
```
jgmontoya: I think we can use a single flag `need_to_publish_relays`

josefinalliende (commented):
Cool! 🎉  I tested it locally from the  `431-relay-management-update` branch with the last commit of this PR and I can add local relay and it connects (previously it didn't). 

I still initially had no relays, I had to add it manually (for an existing account with data). But I'm not sure if maybe my relay list actually got screwed up and got published as empty so it is correct that now I don't have relays 🤔 

After adding a relay, if I close the emulator and open the app again, the added relay appears 🎉 

Captura de pantalla 2025-08-07 a la(s) 4 09 31 p m

src/whitenoise/accounts/mod.rs
```diff
@@ -1049,6 +1081,60 @@ impl Whitenoise {
...
+    pub async fn add_key_package_relay(&self, account: &Account, relay: RelayUrl) -> Result<()> {
+        self.add_relay_to_account(account.pubkey, relay, RelayType::KeyPackage)
+            .await
+    }
+
+    /// Convenience method to remove a key package relay
+    pub async fn remove_key_package_relay(&self, account: &Account, relay: RelayUrl) -> Result<()> {
+        self.remove_relay_from_account(account.pubkey, relay, RelayType::KeyPackage)
+            .await
+    }
```
jgmontoya: I think we should use the 'add_relay_to_account' and 'remove_relay_from_account' methods directly in tests

src/whitenoise/accounts/mod.rs
```diff
@@ -1049,6 +1081,60 @@ impl Whitenoise {
...
+        let account = self.load_account(pubkey).await?;
+
+        let relays = match relay_type {
+            RelayType::Nostr => account.nip65_relays,
+            RelayType::Inbox => account.inbox_relays,
+            RelayType::KeyPackage => account.key_package_relays,
+        };
+
+        Ok(relays)
+    }
```
jgmontoya, line 1096: Not a fan of these helper methods for tests. This one looks like it would be useful though in general, instead of using the `account.<relay_type>`, we can probably use the more general method and avoid code duplication (e.g.: iterating on relay types)

src/whitenoise/accounts/relays.rs
```diff
@@ -154,152 +154,124 @@ impl Whitenoise {
...
-            RelayType::KeyPackage,
-        )
-        .await?;
-        self.publish_relay_list_for_account(account, RelayType::KeyPackage)
-            .await
+        // Update the in-memory account
+        {
+            let mut accounts = self.write_accounts().await;
+            accounts.insert(account.pubkey, account.clone());
+        }
```
jgmontoya, line 189: Are we sure this is necessary? I mean we're already mutating the account relays (above on the `account.<relay_type>.insert...`), right?

src/whitenoise/accounts/relays.rs
```diff
@@ -154,152 +154,124 @@ impl Whitenoise {
...
+
+        // Ensure relays are connected
+        self.nostr
+            .ensure_relays_connected(DashSet::from_iter([relay]))
+            .await?;
+
+        self.publish_relay_list_for_account(&account, relay_type, &None)
+            .await?;
```
jgmontoya, line 197: Should we make sure that we're connected to all relays or would that overkill?

If so, should we make that always when calling `publish_relay_list_for_account`? So basically maybe have that function handle it within?

src/whitenoise/accounts/relays.rs
```diff
@@ -154,152 +154,124 @@ impl Whitenoise {
...
-            account.inbox_relays.clone().into_iter().collect(),
-            RelayType::Inbox,
-        )
-        .await?;
-        self.publish_relay_list_for_account(account, RelayType::Inbox)
+        // Update the in-memory account
+        {
+            let mut accounts = self.write_accounts().await;
+            accounts.insert(account.pubkey, account.clone());
+        }
```
jgmontoya, line 242: Same question

src/whitenoise/accounts/relays.rs
jgmontoya, line 244: I think we should manually disconnect after publishing the relay list, no?

src/whitenoise/accounts/relays.rs
```diff
@@ -154,152 +154,124 @@ impl Whitenoise {
...
-            .await
+        Ok(())
     }
 
-    pub(crate) async fn publish_account_relay_info(&self, account: &Account) -> Result<()> {
+    pub(crate) async fn publish_account_relay_info(
+        &self,
+        account: &Account,
+        prior_relays: &Option<DashSet<RelayUrl>>,
+    ) -> Result<()> {
```
jgmontoya: This one is only being used with `None` `prior_relays`, I'd remove the param

src/whitenoise/accounts/relays.rs
```diff
@@ -154,152 +154,124 @@ impl Whitenoise {
...
-        self.publish_relay_list_for_account(account, RelayType::KeyPackage)
+        self.publish_relay_list_for_account(account, RelayType::KeyPackage, prior_relays)
             .await
     }
 
     pub(crate) async fn publish_relay_list_for_account(
         &self,
         account: &Account,
         relay_type: RelayType,
+        prior_relays: &Option<DashSet<RelayUrl>>, // If provided, this means at least one relay was removed. We need to publish to the prior relays as well.
```
jgmontoya: I can see how this got here as removing a relay is the use case that made this required, however I'd rename it so it's more descriptive of what it does. What do you think about something like 'relays_to_use', or 'target_relays' or  something like that? So basically we know that whatever we pass here means we're publishing to those relays.

src/whitenoise/accounts/relays.rs
jgmontoya: I can answer this for myself: it is necessary (commented it out, saw it fail in the tests)

src/whitenoise/accounts/relays.rs
jgmontoya: same answer

jgmontoya (commented):
Some comments/questions/suggestions :D

src/nostr_manager/query.rs
delcin-raj: It's alright for nip65 relay list for sure, as it is replaceable in most nostr implementations.
I'm not sure about inbox and key package relays, so I used `max_by_key` to be on the safer side.

src/whitenoise/accounts/groups.rs
```diff
@@ -254,8 +254,13 @@ impl Whitenoise {
         // Fetch key packages for all members
         for pk in members.iter() {
             let contact = self.load_contact(pk).await?;
+            let relays_to_use = if contact.key_package_relays.is_empty() {
```
delcin-raj, line 258: Nice catch!

src/whitenoise/accounts/groups.rs
delcin-raj: The assumption is correct. Reference `build_welcome_rumors_for_key_packages` in nostr-mls crate.

src/whitenoise/accounts/groups.rs
delcin-raj: +1

src/whitenoise/accounts/mod.rs
```diff
@@ -329,11 +329,15 @@ impl Whitenoise {
             accounts.insert(account.pubkey, account.clone());
         }
 
-        self.publish_account_relay_info(&account).await?;
-
         self.setup_subscriptions(&account).await?;
 
-        self.publish_key_package_for_account(&account).await?;
+        // TODO: This should only query local nostr cached events, not fetch from relays
```
delcin-raj, line 334: Keypackage events are regularly updated. For this alone we have to fetch from the nostr relays, I think we agreed on this earlier.

delcin-raj (commented):
Left some comments.

src/nostr_manager/query.rs
erskingardner: Nostr SDK automatically orders by most recent created at so the max_by_key is unnecessary when you use `limit(1)`

src/whitenoise/accounts/groups.rs
erskingardner: Going to leave this for now, will create an issue to track though.

src/whitenoise/accounts/relays.rs
erskingardner: yeah, it tricked me too.

src/whitenoise/accounts/relays.rs
erskingardner: The issue with that is that we might be connected to that relay for other reasons, just because we remove it from a list doesn't mean that we might not have a valid reason to be connected to that relay (for a group, for fetching another user's data, etc.) I'm going to leave this for now but we should keep this in mind and see how things work in real world usage.

src/whitenoise/accounts/relays.rs
erskingardner: we use prior relays when we're removing a relay. I think the param is needed.

erskingardner: oh sorry, I think I see what you mean.

src/whitenoise/accounts/relays.rs
erskingardner: I think it's overkill. We don't, in fact, want to connect to all relays. Key package relays are only connected to when we want to publish to them for example.

src/whitenoise/accounts/relays.rs
jgmontoya: That's a very good point. We could have a 'reference counter' approach: Each time we find a reason to connect to a relay, increment its counter by 1, each time one of those reasons goes away, decrement the counter. If the counter is 0, disconnect.

Issue #297, opened 2025-08-07
Add a relay fallback method


erskingardner:
https://github.com/parres-hq/whitenoise/pull/296#discussion_r2260274314

jgmontoya:
I'm taking this

PR #301, opened 2025-08-08
Refactor/contact get relays of type

jgmontoya:
Closes https://github.com/parres-hq/whitenoise/issues/297


src/whitenoise/accounts/groups.rs
```diff
@@ -46,11 +47,7 @@ impl Whitenoise {
 
         for pk in member_pubkeys.iter() {
             let contact = self.load_contact(pk).await?;
-            let relays_to_use = if contact.key_package_relays.is_empty() {
-                Account::default_relays()
-            } else {
-                contact.key_package_relays.clone()
-            };
+            let relays_to_use = contact.get_relays_of_type(RelayType::KeyPackage);
```
erskingardner, line 50: We don't want to use the default relays in this situation do we? e.g. we're fetching a key package for a user, if we don't get a key package relay list then using the account default relays is just stabbing in the dark isn't it? I guess it's better than just failing in some ways but it's not really the right behavior.

jgmontoya: What do we want the behavior to be?? I think there are some options:
- We keep this as is, _stabbing in the dark_, using default relays
- We just fail
- We use other relays for the contact: nip65 relays

I think the last option is the one that makes more sense to me, I feel like that should be the fallback always for `RelayType::KeyPackage`, what do you think?

jgmontoya: As we talked in person, I went with the third option, implemented it for `RelayType::Inbox` and `RelayType::KeyPackage`, and goes to default relays as a last resort if the contact doesn't have nip65 relays

PR #311, opened 2025-08-11
Create new tables and basic database impls

erskingardner:


db_migrations/0005_new_tables.sql
```diff
@@ -0,0 +1,89 @@
...
+CREATE TABLE account_followers (
+    id INTEGER PRIMARY KEY AUTOINCREMENT,
+    account_id INTEGER NOT NULL,
+    user_id INTEGER NOT NULL,
+    created_at DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    updated_at DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP
+);
+
+-- New relays table
+CREATE TABLE relays (
```
delcin-raj, line 31: It's best to not have this table.
A single table `user_relays` with `user_id` being a foreign key referenceing `users` table with

```sql
UNIQUE (user_id, url)
```
constraint will be optimal

jgmontoya: a single relay can belong to multiple lists so we're using the relay table to avoid having to duplicate the relay information/url. It will also allow us to do some additional stuff in the future if we want, like this: https://github.com/parres-hq/whitenoise/pull/296#discussion_r2262172682

delcin-raj: Then we need a `Unique(user_id, relay_id, relay_type)` constraint for the `user_relays` table.
No need to have `id PRIMARY KEY`

PR #319, opened 2025-08-18
Feat/add group information table

jgmontoya:
## Summary by CodeRabbit

- New Features
  - Persistent group metadata with automatic typing (Direct Message vs Group) and timestamps.
  - Group creation accepts an explicit group type; type is inferred when omitted.
  - Fetch group information by one or multiple group IDs.

- Refactor
  - Centralized default relay handling and streamlined relay registration.
  - Newly created or followed accounts now refresh relay lists and user metadata automatically.

- Chores
  - Database migration added to create the group information table.

- Tests
  - Expanded tests for group info, relay sync, and user metadata/update flows.

### DIFF
--- a/src/whitenoise/users.rs
+++ b/src/whitenoise/users.rs
@@ -389,7 +389,6 @@ mod tests {
         assert_eq!(url1, url2);
         assert_ne!(url1, url3);
 
-        // Test HashSet behavior with RelayUrls
         let mut url_set = HashSet::new();
         url_set.insert(&url1);
         url_set.insert(&url2); // Should not increase size since url1 == url2
@@ -404,7 +403,6 @@ mod tests {
     async fn test_update_relay_lists_success() {
         let (whitenoise, _data_temp, _logs_temp) = create_mock_whitenoise().await;
 
-        // Create a test user
         let test_pubkey = nostr_sdk::Keys::generate().public_key();
         let user = User {
             id: None,
@@ -413,11 +411,7 @@ mod tests {
             created_at: Utc::now(),
             updated_at: Utc::now(),
         };
-
-        // Save the user
         let saved_user = user.save(&whitenoise.database).await.unwrap();
-
-        // Create some initial NIP-65 relays for the user
         let initial_relay_url = RelayUrl::parse("wss://initial.example.com").unwrap();
         let initial_relay = whitenoise
             .find_or_create_relay(&initial_relay_url)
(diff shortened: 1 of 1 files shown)


src/whitenoise/accounts/core.rs
```diff
@@ -928,7 +912,8 @@ mod tests {
 
         // Verify that all relay sets contain the same default relays
         // Convert DashSet to Vec to avoid iterator type issues
```
erskingardner: Out of date comment? I think we removed dashset?

src/whitenoise/follows.rs
```diff
@@ -37,6 +37,7 @@ impl Whitenoise {
     pub async fn follow_user(&self, account: &Account, pubkey: &PublicKey) -> Result<()> {
         let (user, newly_created) = User::find_or_create_by_pubkey(pubkey, &self.database).await?;
         if newly_created {
+            user.update_relay_lists(self).await?;
```
erskingardner, line 41: Why don't we also fetch metadata here? Just to make the time we block shorter?

src/whitenoise/accounts/groups.rs
```diff
@@ -484,6 +517,16 @@ mod tests {
         assert_eq!(group.description, config.description);
         assert!(group.admin_pubkeys.contains(&creator_account.pubkey));
         assert!(group.admin_pubkeys.contains(&member_pubkeys[0]));
+
+        if let Ok(group_info) =
```
delcin-raj: The result of `GroupInformation::get_by_mls_group_id` being Err should make this test fail?

src/whitenoise/accounts/core.rs
jgmontoya: yep, deleted

src/whitenoise/follows.rs
jgmontoya: I just didn't do it because it was a bit out-of-scope wrt this PR and wasn't sure if we wanted to block here for longer but taking a second look at it I think it's ok. Will add it

src/whitenoise/accounts/groups.rs
jgmontoya: well spotted, fixing!

src/whitenoise/follows.rs
jgmontoya: done

src/whitenoise/accounts/groups.rs
jgmontoya: done

delcin-raj (approved):

PR #321, opened 2025-08-20
Clean up for flutter

erskingardner:
## Summary by CodeRabbit

- New Features
  - Fetch relay connection status for an account.
  - Retrieve a user’s KeyPackage event.
  - Accounts can list relays by type and read current metadata.

- Improvements
  - Metadata, relay-list, and follow-list publishing now happen in the background for a more responsive experience.
  - Following/unfollowing users automatically triggers background follow-list publishing.
  - Relay handling unified by URL for more reliable setup.
  - Image uploads use standardized MIME types for better compatibility.

- Refactor
  - Streamlined internal sync and relay operations; some controls are now handled within the app for consistency.

### DIFF
--- a/src/nostr_manager/mod.rs
+++ b/src/nostr_manager/mod.rs
@@ -56,17 +56,7 @@ pub struct NostrManager {
 pub type Result<T> = std::result::Result<T, NostrManagerError>;
 
 impl NostrManager {
-    #[allow(dead_code)]
-    pub(crate) async fn add_relays<I>(&self, relays: I) -> Result<()>
-    where
-        I: IntoIterator<Item = RelayUrl>,
-    {
-        for relay in relays {
-            self.client.add_relay(relay).await?;
-        }
-        Ok(())
-    }
-
+    /// Default timeout for client requests
     pub(crate) fn default_timeout() -> Duration {
         Duration::from_secs(5)
     }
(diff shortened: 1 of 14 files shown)


erskingardner:
@coderabbitai - can you review?

src/nostr_manager/mod.rs
```diff
@@ -639,7 +558,8 @@ impl NostrManager {
         Ok(())
     }
 
-    pub async fn sync_all_user_data(
+    /// Syncs all user data to the Nostr client.
```
jgmontoya: This comment kinda threw me off a bit, it makes me think that we're publishing to nostr our user data, but we're doing the opposite.

src/nostr_manager/mod.rs
```diff
@@ -453,70 +370,72 @@ impl NostrManager {
...
-
-    /// Get the status of a specific relay
-    #[allow(dead_code)]
-    pub async fn get_relay_status(&self, relay_url: &RelayUrl) -> Result<RelayStatus> {
+    /// - This method only checks the status of relays that have been previously
+    ///   added to the client's relay pool
+    /// - The status reflects the real-time connection state and may change
+    ///   frequently as network conditions vary
+    /// - Use this method for monitoring relay health and connection diagnostics
+    pub(crate) async fn get_relay_status(&self, relay_url: &RelayUrl) -> Result<RelayStatus> {
```
delcin-raj, line 438: Do we need this detailed documentation for non pub functions?

src/whitenoise/database/users.rs
```diff
@@ -248,8 +247,7 @@ impl User {
         database: &Database,
     ) -> Result<(), WhitenoiseError> {
         let user_id = self.id.ok_or(WhitenoiseError::UserNotPersisted)?;
-        let saved_relay = relay.save(database).await?;
-        let relay_id = saved_relay.id.expect("Relay should have ID after save");
+        let relay_id = relay.id.expect("Relay should have ID after save");
```
jgmontoya, line 250: This makes it so we always have to make sure the relay we pass is already persisted.
Not sure if we call this with unpersisted relays tbh.
Before we were always calling the .save, but maybe what we want is to call the save if it's not persisted (id is None)?? Wdyt?

src/whitenoise/users.rs
```diff
@@ -371,6 +384,35 @@ impl Whitenoise {
...
+        let user_clone = user.clone();
+        let mut mut_user_clone = user.clone();
+
+        tokio::spawn(async move {
+            let whitenoise = Whitenoise::get_instance()?;
+            // Parallel network requests
+            let relay_task = user_clone.update_relay_lists(whitenoise);
+            let metadata_task = mut_user_clone.update_metadata(whitenoise);
+
+            let (relay_result, metadata_result) = tokio::join!(relay_task, metadata_task);
```
jgmontoya: Shouldn't we do this sequentially?? How do we know where to get the metadata from if we haven't updated the relays?

delcin-raj: Yes, this has to be done sequentially

src/nostr_manager/mod.rs
erskingardner: Yeah - i'll update it with an actual doc comment

src/whitenoise/database/users.rs
erskingardner: I think in most places we're calling `find_or_create_by_url` to get the relay that we pass in here, that method will call `save`. I don't think there is another way that we access relays and I think I would prefer that this would fail if a user tries to create a relay on the fly and pass it in.

src/whitenoise/users.rs
erskingardner: hmm, good point. Getting relays first is probably better. and since we're doing this in the background, the parallelism doesn't matter.

delcin-raj: I don't think we should be using the term "parallelism" anywhere in async code.

src/nostr_manager/mod.rs
erskingardner: 🤷‍♂️ I guess not but at the same time, having docs is probably better than not.

PR #326, opened 2025-08-24
Remove admins param from create_group, refactor tests a bit

erskingardner:
## Summary by CodeRabbit

- New Features
  - Update group data/metadata via a new action; updates are merged and propagated to relevant relays.
  - Enhanced Direct Message group handling with correct types, admins, and members.

- Refactor
  - Streamlined group creation: admin/creator keys are supplied inside group configuration rather than separate inputs.
  - Public exports adjusted to make group information and types more directly available.

### DIFF
--- a/src/whitenoise/groups.rs
+++ b/src/whitenoise/groups.rs
@@ -63,11 +63,8 @@ impl Whitenoise {
         let group_relays = config.relays.clone();
 
         let nostr_mls = Account::create_nostr_mls(creator_account.pubkey, &self.config.data_dir)?;
-        let create_group_result = nostr_mls.create_group(
-            &creator_account.pubkey,
-            key_package_events.clone(),
-            config
-        )?;
+        let create_group_result =
+            nostr_mls.create_group(&creator_account.pubkey, key_package_events.clone(), config)?;
 
         let group_ids = nostr_mls
             .get_groups()?
@@ -370,7 +367,12 @@ impl Whitenoise {
     /// * `account` - The account performing the group data update (must be group admin)
     /// * `group_id` - The ID of the group to update
     /// * `group_data` - The new group data to update
-    pub async fn update_group_data(&self, account: &Account, group_id: &GroupId, group_data: NostrGroupDataUpdate) -> Result<()> {
+    pub async fn update_group_data(
+        &self,
+        account: &Account,
+        group_id: &GroupId,
+        group_data: NostrGroupDataUpdate,
+    ) -> Result<()> {
         let nostr_mls = Account::create_nostr_mls(account.pubkey, &self.config.data_dir)?;
         let update_result = nostr_mls.update_group_data(group_id, group_data)?;
         nostr_mls.merge_pending_commit(group_id)?;
@@ -389,8 +391,6 @@ impl Whitenoise {
             .await?;
         Ok(())
     }
-
-
 }
 
 #[cfg(test)]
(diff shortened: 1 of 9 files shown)


src/whitenoise/groups.rs
```diff
@@ -364,6 +358,39 @@ impl Whitenoise {
...
+        nostr_mls.merge_pending_commit(group_id)?;
+        let group_relays = nostr_mls.get_relays(group_id)?;
+
+        let evolution_event = update_result.evolution_event;
+
+        let mut relays = HashSet::new();
+        for relay_url in group_relays {
+            let db_relay = self.find_or_create_relay_by_url(&relay_url).await?;
+            relays.insert(db_relay);
+        }
```
jgmontoya, line 387: I don't think we need to deduplicate here because `nostr_mls.get_relays` already returns a set (`BTreeSet<RelayUrl>`)

jgmontoya (approved):
Looks good, just a minor thing

src/whitenoise/groups.rs
erskingardner: This is more about making sure the relays are actually created in our db rather than deduplication.

src/whitenoise/groups.rs
```diff
@@ -311,7 +305,7 @@ impl Whitenoise {
             // Create a timestamp 1 month in the future
             let one_month_future = Timestamp::now() + Duration::from_secs(30 * 24 * 60 * 60);
 
-            // Use fallback relays if contact has no inbox relays configured
+            //  TODO: Maybe need to use fallback relays if contact has no inbox relays configured
```
delcin-raj, line 308: I can see that when a new user is created, their relays are not inserted by default.
Reference `fn find_or_create_by_pubkey`

src/whitenoise/groups.rs
```diff
@@ -364,6 +358,39 @@ impl Whitenoise {
...
+    /// * `group_data` - The new group data to update
+    pub async fn update_group_data(
+        &self,
+        account: &Account,
+        group_id: &GroupId,
+        group_data: NostrGroupDataUpdate,
+    ) -> Result<()> {
+        let nostr_mls = Account::create_nostr_mls(account.pubkey, &self.config.data_dir)?;
+        let update_result = nostr_mls.update_group_data(group_id, group_data)?;
+        nostr_mls.merge_pending_commit(group_id)?;
```
delcin-raj, line 378: This could be done in nostr-mls?

src/whitenoise/groups.rs
delcin-raj:
```rust
        let mut relays = Vec::new();
        for relay_url in group_relays {
            let db_relay = self.find_or_create_relay_by_url(&relay_url).await?;
            relays.push(db_relay);
        }

        self.nostr
            .publish_event_to(evolution_event, &relays)
            .await?;
   ```
   
   This version is cleaner.

delcin-raj (commented):
Left some comments and questions.
The tests look comprehensive!

src/whitenoise/groups.rs
erskingardner: No, needs to happen here. The client should be in control of when to call it because they have to be the ones to publish the actual event to relays

src/whitenoise/groups.rs
erskingardner: Yeah - actually - we need to be really careful about this don't we. When we're doing lots of different operations we first need to get user's and also make sure we have their relays etc. I think this deserves another issue that we tackle separately to ensure that we're checking every location where we find or create a user to ensure that we're also trying to fetch their relays and metadata.

PR #329, opened 2025-08-25
Create event publish and processor tracking

erskingardner:
## Summary by CodeRabbit

- New Features
  - Automatic contact-list syncing: follows/unfollows update from received events.
  - Persistent event tracking and deduplication to avoid reprocessing.

- Improvements
  - More reliable group creation and messaging with explicit relay validation and safer fallbacks.
  - Smarter relay synchronization to reduce redundant updates.
  - Message aggregation uses retry polling for stability.
  - Follow checks return false on lookup failures instead of errors.
  - Clear error for groups missing relays.

- Bug Fixes
  - Prevents duplicate processing and ignores self-published events where appropriate.


src/whitenoise/event_processor/event_handlers/handle_contact_list.rs
```diff
@@ -0,0 +1,107 @@
...
+        // Process unfollows (but don't publish follow list after each individual unfollow)
+        for pubkey in &users_to_unfollow {
+            let (user, _) =
+                crate::whitenoise::users::User::find_or_create_by_pubkey(pubkey, &self.database)
+                    .await?;
+            account.unfollow_user(&user, &self.database).await?;
+        }
+
+        // Only publish the follow list once after all changes are made
+        self.background_publish_account_follow_list(account).await?;
```
delcin-raj: This function is called when processing events from a nostr relay.
I'm assuming we are publishing again, because some of the nostr relays may not contain this updated event?

src/whitenoise/event_processor/event_handlers/handle_contact_list.rs
```diff
@@ -0,0 +1,107 @@
...
+use crate::{
+    nostr_manager::NostrManager,
+    whitenoise::{accounts::Account, error::Result, Whitenoise},
+};
+
+impl Whitenoise {
+    /// Contact list handler that performs the actual processing
+    /// Note: Event tracking (published/processed checks) is handled at the processor level
+    pub(crate) async fn handle_contact_list(&self, account: &Account, event: Event) -> Result<()> {
+        // Acquire per-account semaphore permit to serialize contact list processing for this account
```
delcin-raj, line 14: Excellent use of Semaphore. I've finally seen it in action today!

src/whitenoise/groups.rs
```diff
@@ -214,8 +214,21 @@ impl Whitenoise {
...
+            // Try and get user's key package relays, if they don't have any, use account's default relays
+            let mut relays_to_use = user.relays(RelayType::KeyPackage, &self.database).await?;
+            if relays_to_use.is_empty() {
+                tracing::warn!(
+                    target: "whitenoise::accounts::groups::add_members_to_group",
+                    "User {} has no relays configured, using account's default relays",
+                    user.pubkey
+                );
+                relays_to_use = account.nip65_relays(self).await?;
+            }
```
delcin-raj, line 247: I suggest changing the test for creating a group.
The members should be users, not accounts.
Currently, `create_identity` is used for members in that test.
That's not the expected behaviour.

delcin-raj (approved):
Left some comments

src/whitenoise/event_processor/event_handlers/handle_contact_list.rs
erskingardner: We are changing the contact list for the user in the lines above this, but that's only local changes in our own database. We still need to publish those changes to relays when we're finished.

src/whitenoise/groups.rs
erskingardner: I think it's ok. In this case we do this so that we get the automatic set (key packages, etc) that create_identity gives us. The behaviour for creating the group is basically the same because we're still going out and fetching the key packages for those member users we're trying to add.

src/integration_tests/test_cases/advanced_messaging/aggregate_messages.rs
```diff
@@ -46,12 +84,28 @@ impl TestCase for AggregateMessagesTestCase {
...
+                    msg.is_deleted,
+                    msg.kind
+                );
+            }
+
+            panic!(
+                "Expected at least {} messages, but got {} after retries. Check logs for details.",
+                self.expected_min_messages,
+                aggregated_messages.len()
+            );
```
jgmontoya: If at this point we return an error instead of crashing, it will work nicer with the integration test framework (allowing it to continue with the next scenario)

src/nostr_manager/mod.rs
jgmontoya, line 1: I'm not sure I like that we're adding this logic here tbh. I think ideally the nostr_manager should't really need to know about our higher-level structs, like Account. I know it was already the case, but this feels like we're moving in the opposite direction. I think the nostr_manager should only be concerned about nostr things, and not app-level or business logic.

I think for this kind of things we could have a Service object that helps us, something like this:

```rs
pub struct PublishingService {
    nostr_manager: NostrManager,
    whitenoise: Arc<Whitenoise>,
}

impl PublishingService {
    pub async fn publish_and_track(&self, ...) -> Result<Output<EventId>> {
        let result = self.nostr_manager.publish_event_builder_with_signer(...).await?;
        // Handle tracking here at the business logic layer
        Ok(result)
    }
}
```

What do you think??

src/nostr_manager/mod.rs
```diff
@@ -229,6 +236,47 @@ impl NostrManager {
...
+                }
+                Err(e) => {
+                    tracing::warn!(
+                        target: "whitenoise::nostr_manager::publish_event_builder_with_signer",
+                        "Published event {} but Whitenoise instance unavailable for tracking: {}",
+                        event_id.to_hex(),
+                        e
+                    );
+                }
+            }
```
jgmontoya: I think this should be extracted to a private method

src/nostr_manager/utils.rs
jgmontoya, line 1: Seems very straight forward but some simple test on this wouldn't hurt

src/whitenoise/database/published_events.rs
```diff
@@ -0,0 +1,411 @@
...
+        let result: (i64,) = sqlx::query_as(
+            "SELECT COUNT(*) FROM published_events WHERE event_id = ? AND account_id = ?",
+        )
+        .bind(event_id.to_hex())
+        .bind(account_id)
+        .fetch_one(&database.pool)
+        .await?;
+
+        Ok(result.0 > 0)
+    }
```
jgmontoya: I think this would be more idiomatic:

```rs
pub async fn exists_optional(/* ... */) -> Result<bool, DatabaseError> {
    let result = sqlx::query(
        "SELECT 1 FROM published_events WHERE event_id = ? AND account_id = ? LIMIT 1"
    )
    .bind(event_id.to_hex())
    .bind(account_id)
    .fetch_optional(&database.pool)  // Returns None if no rows found
    .await?;
    Ok(result.is_some())
}
```

src/whitenoise/database/published_events.rs
```diff
@@ -0,0 +1,411 @@
...
+             WHERE account_id = ? AND event_kind = ?
+             ORDER BY created_at DESC",
+        )
+        .bind(account_id)
+        .bind(event_kind.as_u16() as i64)
+        .fetch_all(&database.pool)
+        .await?;
+
+        Ok(rows)
+    }
```
jgmontoya: This method doesn't seem to be used

src/whitenoise/database/published_events.rs
```diff
@@ -0,0 +1,411 @@
...
+    pub async fn cleanup_old(database: &Database, days_old: i32) -> Result<u64, DatabaseError> {
+        let result = sqlx::query(
+            "DELETE FROM published_events WHERE created_at < datetime('now', '-' || ? || ' days')",
+        )
+        .bind(days_old)
+        .execute(&database.pool)
+        .await?;
+
+        Ok(result.rows_affected())
+    }
```
jgmontoya: This one also doesn't seem to be used at all (it's not even in tests 👀)

src/whitenoise/database/published_events.rs
jgmontoya, line 1: None of these methods should be public. They should at most be pub(crate)

src/whitenoise/database/published_events.rs
```diff
@@ -0,0 +1,411 @@
...
+    /// Checks if we already processed a specific event
+    pub async fn exists(database: &Database, event_id: &EventId) -> Result<bool, DatabaseError> {
+        let result: (i64,) =
+            sqlx::query_as("SELECT COUNT(*) FROM processed_events WHERE event_id = ?")
+                .bind(event_id.to_hex())
+                .fetch_one(&database.pool)
+                .await?;
+
+        Ok(result.0 > 0)
+    }
```
jgmontoya: Same idea with this one regarding using `SELECT 1` and `fetch_optional`

src/whitenoise/database/published_events.rs
```diff
@@ -0,0 +1,411 @@
...
+    pub async fn cleanup_old(database: &Database, days_old: i32) -> Result<u64, DatabaseError> {
+        let result = sqlx::query(
+            "DELETE FROM processed_events WHERE processed_at < datetime('now', '-' || ? || ' days')"
+        )
+        .bind(days_old)
+        .execute(&database.pool)
+        .await?;
+
+        Ok(result.rows_affected())
+    }
```
jgmontoya: These two also don't seem to be used at all

db_migrations/0011_published_events.sql
```diff
@@ -0,0 +1,28 @@
+-- Generic table to track events we've published to prevent infinite loops
+CREATE TABLE published_events (
+    id INTEGER PRIMARY KEY AUTOINCREMENT,
+    event_id TEXT NOT NULL UNIQUE
+        CHECK (length(event_id) = 64 AND event_id GLOB '[0-9a-fA-F]*'), -- 64-char hex
+    account_id INTEGER NOT NULL,          -- Who published it
+    event_kind INTEGER NOT NULL,          -- Nostr event kind (3, 0, 10002, etc.)
```
jgmontoya: I'm not sure I understand, why do we need these two attributes?

db_migrations/0011_published_events.sql
```diff
@@ -0,0 +1,28 @@
...
+-- Indexes for efficient lookups
+CREATE INDEX idx_published_events_lookup ON published_events(event_id);
+CREATE INDEX idx_published_events_account_kind ON published_events(account_id, event_kind);
+
+-- Table to track processed events to ensure idempotency
+CREATE TABLE processed_events (
+    id INTEGER PRIMARY KEY AUTOINCREMENT,
+    event_id TEXT NOT NULL UNIQUE
+        CHECK (length(event_id) = 64 AND event_id GLOB '[0-9a-fA-F]*'), -- 64-char hex
+    event_kind INTEGER NOT NULL,          -- Nostr event kind
```
jgmontoya: Why do we need the `event_kind` here?

src/whitenoise/event_processor/event_handlers/handle_contact_list.rs
jgmontoya, line 1: Should we call this follow_list instead of contact_list, everywhere?

src/whitenoise/event_processor/event_handlers/handle_contact_list.rs
```diff
@@ -0,0 +1,107 @@
...
+        let semaphore = self
```
jgmontoya, line 15: I'm not sure I understand why the semaphore is needed 😬 as I see it, the process_events run in a single tokio:spawn task with a sequential loop that processes one event at a time, so handle_contact_list wont ever be called concurrently, no?

src/whitenoise/event_processor/event_handlers/handle_contact_list.rs
jgmontoya: Why do we need to publish if in principle we got this changes from an event from elsewhere? So this would already be published, no?

src/whitenoise/event_processor/mod.rs
```diff
@@ -95,20 +91,76 @@ impl Whitenoise {
                     // Process the event
                     match event {
                         ProcessableEvent::NostrEvent { event, subscription_id, retry_info } => {
+                            // Check if we already processed this event - skip if so
+                            let already_processed = whitenoise.already_processed_event(&event.id).await.unwrap_or(false);
+                            if already_processed {
```
jgmontoya, line 119: Besides the case where we were the ones that published the event, can it happen that we receive the event more than once? Maybe because we get it from multiple different relays?

src/whitenoise/event_processor/mod.rs
```diff
@@ -95,20 +91,76 @@ impl Whitenoise {
...
+                                    "Already processed event: {} (kind {})",
+                                    event.id.to_hex(),
+                                    event.kind.as_u16()
+                                );
+                                continue; // Skip to next event
+                            }
+
+                            // For account-specific events, check if WE published this event
+                            // We don't skip giftwraps and MLS messages because we need them to process in nostr-mls
+                            let should_skip = match event.kind {
```
jgmontoya, line 131: I think I'm not quite understanding this. Isn't the whole purpose of the already_processed to handle this?

src/whitenoise/event_tracking.rs
```diff
@@ -0,0 +1,48 @@
...
+
+    /// Checks if we already processed a specific event
+    pub(crate) async fn already_processed_event(
+        &self,
+        event_id: &nostr_sdk::EventId,
+    ) -> Result<bool> {
+        let result = ProcessedEvent::exists(&self.database, event_id).await?;
+        Ok(result)
+    }
+}
```
jgmontoya: Having this pub(crate) methods hanging from Whitenoise feels off

src/nostr_manager/mod.rs
erskingardner: I feel like that is going to get very circular. `NostrManager` is a field on `Whitenoise` already. I'm open to the idea of trying to reduce the bleed of database stuff into these methods though. I'm going to start by at least breaking the publish related methods into a submodule in NostrManager so we can see them all more easily.

src/whitenoise/database/published_events.rs
erskingardner: Moved into the tests as a helper function

src/whitenoise/event_processor/event_handlers/handle_contact_list.rs
erskingardner: I used Contact List because that's the nostr convention. It's a kind 3 contact list event. It's a little confusing at first but feels accurate to me. What do you think?

src/whitenoise/event_processor/mod.rs
erskingardner: Yes, it's definitely possible to receive the same event from two relays separately. nostr_sdk does try hard to deduplicate but it's not perfect.

src/whitenoise/event_processor/event_handlers/handle_contact_list.rs
erskingardner: Yes, we need it (we might not have previously but I just added some other code that makes it necessary). Previously we weren't actually including the user's follow list (and follow syncing) in the `sync_all_user_data` method. I added that to the method so that on first load / re-load we will synchronize the user's follows list - both adding and removing users. This uses the handle_contact_list.rs handler. 

We are also using this handler in the event processor code. That means that there are multiple places where we might be trying to update a given account's follows at a given time. This makes sure that we don't try and do that at the same time.

> The semaphore uses contact_list_guards: `DashMap<PublicKey, Arc<Semaphore>>` which creates a separate semaphore for each account's public key.
>
> Cross-Task Protection: Since both call sites access the same Whitenoise instance (stored in a static OnceCell), they share the same contact_list_guards DashMap. The Arc<Semaphore> ensures the semaphore works across different tasks.
>
> Race Condition Prevention: Without this semaphore, you could have:
> -  Event processor handling a contact list event
> - Background sync simultaneously fetching and processing the same account's contact list
> - Both trying to modify the same account's follows/unfollows simultaneously
> - Database inconsistencies or lost updates

src/whitenoise/event_processor/event_handlers/handle_contact_list.rs
erskingardner: You're right - got confused here I think.

db_migrations/0011_published_events.sql
erskingardner: We actually don't need the kind, the account id is required though because we need to who published vs who is receiving so we can properly decide waht to do.

src/whitenoise/database/published_events.rs
erskingardner: removed.

src/whitenoise/database/published_events.rs
erskingardner: removed

db_migrations/0011_published_events.sql
erskingardner: we don't, removed.

src/nostr_manager/mod.rs
erskingardner: I've abstracted this into a set of trait methods instead. I really like how it all ended up. Curious what you think.

erskingardner: It's all in the `event_tracker.rs` file

src/whitenoise/event_processor/mod.rs
erskingardner: Another situation where this happens is that we're fetching directly to sync up, but then we could also get the same events via subscription (we try to avoid this with the `since` param on the filters but I'm dubious it's perfect)

src/whitenoise/event_processor/mod.rs
erskingardner: This is for checking whether the event was published by us, not processed already. I've changed the logic here to be more readable. We want to process giftwraps and MLS messages even if we published them so that we can allow nostr_mls to update some things in it's data store.

src/whitenoise/event_tracking.rs
erskingardner: Moved into trait methods.

src/whitenoise/event_processor/event_handlers/handle_contact_list.rs
jgmontoya: According to the latest of NIP-02 it's follow list: https://github.com/nostr-protocol/nips/blob/master/02.md

I'd rather we keep it consistent and call it follow instead of contact as that's the term we're using everywhere

src/whitenoise/event_processor/event_handlers/handle_contact_list.rs
```diff
@@ -0,0 +1,104 @@
...
+            .contact_list_guards
+            .entry(account.pubkey)
+            .or_insert_with(|| Arc::new(Semaphore::new(1)))
+            .clone();
```
jgmontoya, line 19: Given that we want a single thread to have access here, I think using a lock/mutex would be a better (and simpler) choice

db_migrations/0011_published_events.sql
```diff
@@ -0,0 +1,31 @@
...
+CREATE INDEX idx_published_events_lookup ON published_events(event_id);
+CREATE INDEX idx_published_events_account_id ON published_events(account_id);
+
+-- Table to track processed events to ensure idempotency
+CREATE TABLE processed_events (
+    id INTEGER PRIMARY KEY AUTOINCREMENT,
+    event_id TEXT NOT NULL
+        CHECK (length(event_id) = 64 AND event_id GLOB '[0-9a-fA-F]*'), -- 64-char hex
+    account_id INTEGER NOT NULL,          -- Which account was it processed for?
+    processed_at DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
```
jgmontoya: Definitely a nit, but I'd use created_at here as well for consistency

src/whitenoise/event_processor/event_handlers/handle_contact_list.rs
erskingardner: Claude and I disagree. 😂 In any case, I think we're splitting hairs. Let's leave as is for now

-----

## Current Implementation Analysis

The current code uses:
- `DashMap<PublicKey, Arc<Semaphore>>` where each semaphore has a capacity of 1
- `Semaphore::acquire_owned()` to get exclusive access per account
- The semaphore permits are dropped automatically when the function completes

## Arc<Mutex<T>> vs Current Semaphore: Trade-offs

### **Current Semaphore Implementation is BETTER** for this use case. Here's why:

**Advantages of the current semaphore approach:**

1. **Cleaner async integration**: 
   - `semaphore.acquire_owned().await` is naturally async and cancellation-safe
   - No risk of holding locks across await points (which can cause deadlocks)
   - Works seamlessly with Tokio's cooperative scheduling

2. **Better error handling**:
   - Semaphore acquisition can fail gracefully if the semaphore is closed
   - The current code properly handles this with `.map_err()`

3. **Automatic cleanup**:
   - `acquire_owned()` returns an owned permit that automatically releases when dropped
   - No need to manually unlock or worry about poisoning

4. **Cancellation safety**:
   - If the task is cancelled while waiting for the permit, cleanup happens automatically
   - No risk of leaving a mutex locked

**Why Arc<Mutex<T>> would be WORSE:**

1. **Async unfriendly**:
   ```rust
   // Bad - holding mutex across await points
   let guard = mutex.lock().await;
   some_async_operation().await; // Could block other tasks
   // guard dropped here
   ```

2. **Potential for deadlocks**:
   - Easy to accidentally hold a mutex guard across async boundaries
   - Can lead to executor starvation

3. **More verbose error handling**:
   - Need to handle poisoning (though less relevant with async mutexes)
   - Less ergonomic than semaphore's simple acquire/release

4. **What would you mutex guard?**:
   - You'd need `Arc<Mutex<()>>` just to get mutual exclusion
   - This is essentially a binary semaphore with worse ergonomics

### **The current implementation is well-designed** because:

- It correctly serializes contact list processing per account [[memory:6095031]]
- Uses async-friendly synchronization primitives
- Has proper error handling and automatic cleanup
- Scales well (separate semaphore per account, not global locking)

**Recommendation**: Keep the current semaphore implementation. It's the right tool for this job and follows Rust async best practices.

jgmontoya: This Claude guy is playing both sides let me tell ya 😂 :

>  Misconceptions in the Response
> 
>   1. "Mutex is async unfriendly" - FALSE
>   - tokio::sync::Mutex is designed specifically for async code
>   - It's perfectly safe to hold across await points
>   - The "bad example" in the response is actually fine with
>   tokio::sync::Mutex
> 
>   2. "Potential for deadlocks" - MISLEADING
>   - Both tokio::sync::Mutex and Semaphore have the same deadlock
>    characteristics
>   - The issue described applies equally to both primitives
> 
>   3. "What would you mutex guard?" - MISSES THE POINT
>   - Arc<Mutex<()>> is the standard idiom for per-key locking in
>   Rust
>   - It's not "worse ergonomics" - it's more explicit about
>   intent
> 
>   Technical Reality Check
> 
>   Both approaches are functionally identical:
> 
>   // Semaphore (current)
>   let _permit = semaphore.acquire_owned().await.map_err(...)?;
>   some_async_operation().await;
>   // permit dropped automatically
> 
>   // Mutex (proposed)  
>   let _guard = mutex.lock().await;
>   some_async_operation().await;
>   // guard dropped automatically
> 
>   Both have:
>   - Automatic cleanup on drop
>   - Async-safe operation
>   - Same cancellation safety
>   - Same deadlock potential
> 
>   The Real Difference
> 
>   The only practical differences are:
> 
>   1. Semantic clarity: Mutex expresses "exclusive access to
>   data", Semaphore expresses "resource counting"
>   2. Error handling: Semaphore acquisition can fail if closed,
>   Mutex cannot
>   3. API complexity: Semaphore requires error mapping, Mutex
>   does not
> 
>   Verdict
> 
>   The pushback is technically incorrect about async Mutex
>   capabilities. However, since:
>   - The current code works fine
>   - The difference is primarily semantic
>   - Both approaches are equally correct functionally
> 
>   I'd recommend keeping the current semaphore implementation -
>   not because it's better, but because it's already working and
>   the benefit of changing is minimal.
> 
>   The response's technical claims about async Mutex are wrong,
>   but the pragmatic conclusion (keep current code) is
>   reasonable.

PR #336, opened 2025-08-29
feat: make user find or create public

josefinalliende:
# Context

The bridge in flutter repo has a `getUser `method that called` find_user_by_pubkey,` but actually needed to create users too if not found.

# Solution
Make public the User `find_or_create_by_pubkey `method, replacing  the use of  `find_or_create_by_pubkey`.


src/whitenoise/users.rs
```diff
@@ -339,6 +339,53 @@ impl Whitenoise {
...
+    /// ```
+    ///
+    /// # Errors
+    ///
+    /// This method will return an error if:
+    /// - There's a database connection or query error
+    /// - The public key format is invalid (though this is typically caught at the type level)
+    pub async fn find_or_create_user_by_pubkey(&self, pubkey: &PublicKey) -> Result<(User, bool)> {
+        User::find_or_create_by_pubkey(pubkey, &self.database).await
+    }
```
jgmontoya: For the created case I think we want to update/fetch metadata and relays

jgmontoya: @josefinalliende take a look at how it's handled in create_group within src/whitenoise/groups.rs

erskingardner: 👍

pepina-dev: Thanks! Working on it!

src/integration_tests/scenarios/user_discovery.rs
```diff
@@ -0,0 +1,46 @@
...
+}
+
+#[async_trait]
+impl Scenario for UserDiscoveryScenario {
+    fn context(&self) -> &ScenarioContext {
+        &self.context
+    }
+
+    async fn run_scenario(&mut self) -> Result<(), WhitenoiseError> {
+        tracing::info!("=== Starting Find Or Create User Scenario ===");
```
jgmontoya: This is not really needed, it's handled automatically for scenarios (check out the execute method on the Scenario trait (`src/integration_tests/core/traits.rs`)

src/integration_tests/test_cases/user_discovery/find_or_create_user.rs
```diff
@@ -0,0 +1,226 @@
...
+        ];
+
+        Self {
+            test_keys: keys,
+            should_have_metadata: false,
+            should_have_relays: true,
+            test_metadata: None,
+            test_relays,
+        }
+    }
```
jgmontoya: Please use the builder pattern, that way we can use both `with_metadata` and `with_relays` together instead of needing to define `with_metadata_and_relays`. See `CreateGroupTestCase` and/or `SendMessageTestCase` for examples of this

src/integration_tests/test_cases/user_discovery/find_or_create_user.rs
```diff
@@ -0,0 +1,226 @@
...
+                        crate::whitenoise::relays::RelayType::Nip65,
+                        context.whitenoise,
+                    )
+                    .await?;
+            }
+        }
+
+        tokio::time::sleep(tokio::time::Duration::from_millis(500)).await;
+        context.whitenoise.logout(&temp_account.pubkey).await?;
+
```
jgmontoya: Using a temp account and then logging out can be messing your test up (because there might be things leaking from that into your test). Use a nostr test client, check out `create_test_client` in `src/integration_tests/core/test_clients.rs`; it's used in `LoginTestCase`.

Also I noticed that I missed using the `create_test_client` method on `PublishMetadataUpdateTestCase` and `PublishRelayListUpdateTestCase`. Could you do it? If not just let me know and  I'll do it later

src/integration_tests/test_cases/user_discovery/find_or_create_user.rs
```diff
@@ -0,0 +1,226 @@
...
+        let user_exists = context
+            .whitenoise
+            .find_user_by_pubkey(&test_pubkey)
+            .await
+            .is_ok();
+        assert!(!user_exists, "User should not exist initially");
+
+        if self.should_have_metadata || self.should_have_relays {
+            self.publish_test_data(context).await?;
+        }
```
jgmontoya: I find that the logic around should_have_metadata and should_have_relays is kinda convoluted: Here you're doing a check, then inside `publish_test_data` you're doing individual checks again. I think you can probably simplify this by having each of them be handled by a different method and then just checking directly before calling each of them. What do you think?

src/whitenoise/users.rs
```diff
@@ -339,6 +339,69 @@ impl Whitenoise {
         User::find_by_pubkey(pubkey, &self.database).await
     }
 
+    /// Finds a user by their public key or creates a new one if not found.
+    ///
+    /// This method looks up a user in the database using their Nostr public key.
+    /// If the user doesn't exist, it creates a new user record with default metadata.
```
jgmontoya: >    /// If the user doesn't exist, it creates a new user record with default metadata.

This seems inaccurate. It creates a new user record ✅  but it tries to fetch its metadata, it doesn't just fill a 'default' one

src/whitenoise/users.rs
```diff
@@ -339,6 +339,69 @@ impl Whitenoise {
...
+    /// It also updates the created user's relay lists and metadata.
+    /// The user may have been discovered through various means such as:
+    /// - Following lists from accounts
+    /// - Message interactions
+    /// - Direct user lookups
+    /// - Metadata events
```
jgmontoya: I don't understand what this is saying really, I think we should delete this part

jgmontoya (commented):
Nice, love to see the integration tests expanding ♥️ ! Just a couple of comments 🙏

src/whitenoise/users.rs
pepina-dev: I copied it from the `find_user_by_pubkey` method. My understanding is that this allows users to be loaded without having directly interacted with them, which is fine ( it doesn’t necessarily mean it’s a user you follow). But yes, it’s probably more close to noise.

src/integration_tests/scenarios/user_discovery.rs
pepina-dev: I'm not sure if I understand correctly, cause other scenarios also have the run_scenario method... 🤔

pepina-dev: ah you meant the log!

src/integration_tests/test_cases/user_discovery/find_or_create_user.rs
pepina-dev: Thanks for the advice!

src/whitenoise/users.rs
pepina-dev: You're right!

src/whitenoise/users.rs
pepina-dev: ✅

src/integration_tests/scenarios/user_discovery.rs
pepina-dev: ✅

src/integration_tests/test_cases/user_discovery/find_or_create_user.rs
pepina-dev: ✅

src/integration_tests/test_cases/user_discovery/find_or_create_user.rs
pepina-dev: ✅

src/whitenoise/users.rs
pepina-dev: ✅

src/whitenoise/users.rs
pepina-dev: ✅

src/integration_tests/test_cases/user_discovery/find_or_create_user.rs
pepina-dev: ✅

jgmontoya (approved):

PR #337, opened 2025-08-29
Manage encrypted group image using Blossom

delcin-raj:
## Summary by CodeRabbit

- New Features
  - Set, update, and retrieve encrypted group images with secure upload and local caching for faster loads.
  - Streamlined profile picture upload flow (no server parameter required).

- Error Handling
  - Clearer, unified media and Blossom-related error reporting for upload/retrieval issues and missing images.

- Database
  - Persist optional group image pointer for MLS groups.

- Chores
  - Updated nostr-related dependencies and initialized dedicated storage for group images.


src/whitenoise/mod.rs
```diff
@@ -338,11 +341,21 @@ impl Whitenoise {
...
         &self.message_aggregator
     }
+
+    fn default_blossom() -> BlossomClient {
+        if cfg!(debug_assertions) {
+            BlossomClient::new(Url::parse("http://localhost:3000").unwrap())
+        } else {
+            BlossomClient::new(Url::parse("https://blossom.primal.net/").unwrap())
+        }
+    }
```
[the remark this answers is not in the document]
delcin-raj: We don't need flexibility now

src/whitenoise/groups.rs
```diff
@@ -486,6 +492,98 @@ impl Whitenoise {
...
+                let decrypted_bytes = decrypt_data(&encrypted_bytes, &image_key, &image_nonce)?;
+
+                // Save the decrypted files locally
+                let image_pointer = self
+                    .config
+                    .data_dir
+                    .join("group_images")
+                    .join(sha256.to_string());
+                fs::write(image_pointer.clone(), &decrypted_bytes).await?;
+
```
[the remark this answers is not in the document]
delcin-raj: This was taken care in, initialization

src/media_old/blossom.rs
erskingardner, line 1: Unless you think these are going to be reused very soon, I think i would just delete all these old files that you don't need. we can always dig them back up via git.

src/whitenoise/groups.rs
```diff
@@ -486,6 +492,102 @@ impl Whitenoise {
         // TODO: Do any local updates to ensure that we're accurately reflecting that the account is trying to leave this group
         Ok(())
     }
+
+    pub async fn update_group_image(
```
erskingardner, line 496: Don't we need to update the group_information pointer here too?

Cargo.toml
```diff
@@ -75,6 +75,7 @@ dotenvy = "0.15"
 [dev-dependencies]
 mockito = "1.2"
 tempfile = "3.19.1"
+image_extern = { package = "image", version = "0.25", features = ["png"] }
```
erskingardner, line 78: We'll need to support jpg, jpeg, png, webp, and gif at a minimum.

src/media_old/blossom.rs
delcin-raj: Yes, that's the best practice

Cargo.toml
delcin-raj: This is used only for testing.
I used it to randomly generate PNG files that are encrypted and uploaded to blossom.

delcin-raj: This is used only for testing.
I used it to randomly generate PNG files that are encrypted and uploaded to blossom.

db_migrations/0012_add_group_image_to_group_information.sql
```diff
@@ -0,0 +1,4 @@
+-- Migration: Store local file pointer for the group image (no inline BLOB)
+
+ALTER TABLE group_information
+  ADD COLUMN image_pointer TEXT;
```
jgmontoya, line 4: I think the name `image_pointer` is a bit confusing, even though conceptually it can be thought as a pointer I would use something like `image_path` just to try to make it clearer that we're talking about a path in the filesystem rather than a memory-address of sorts

db_migrations/0012_add_group_image_to_group_information.sql
jgmontoya, line 1: I'm not completely sure if it's the best idea but maybe we could use the already existing `media_files` table to store the information related to the actual image (which already has the `file_path`) and here just add a foreign key reference to it?

src/whitenoise/groups.rs
```diff
@@ -486,6 +492,102 @@ impl Whitenoise {
...
+            .get_nostr_keys_for_pubkey(&account.pubkey)?;
+        let blob_descriptor = self
+            .blossom
+            .upload_blob(encrypted_bytes, None, None, Some(&account_keys))
+            .await
+            .map_err(BlossomError::from)?;
+
+        tracing::info!(
+            "groups::update_group_image: Uploaded blob to blossom for image in {image_path}"
+        );
```
jgmontoya, line 520: This could probably be extracted to an `upload_image` or `upload_media` method, what do you think?

src/whitenoise/groups.rs
```diff
@@ -486,6 +492,102 @@ impl Whitenoise {
...
+    }
+
+    pub async fn get_group_image(&self, account: &Account, group_id: &GroupId) -> Result<Vec<u8>> {
+        let nostr_mls = Account::create_nostr_mls(account.pubkey, &self.config.data_dir)?;
+        let group_info = GroupInformation::get_by_mls_group_id(group_id, self).await?;
+        match group_info.image_pointer {
+            Some(image_pointer) => {
+                let image_bytes = fs::read(image_pointer).await?;
+                Ok(image_bytes)
+            }
```
jgmontoya, line 538: When we're on the first branch (`Some(image_pointer)`) we don't really need to have instanced the `nostr_mls` object. We should move that to the `None` branch

src/whitenoise/groups.rs
```diff
@@ -486,6 +492,102 @@ impl Whitenoise {
...
+
+                let decrypted_bytes = decrypt_data(&encrypted_bytes, &image_key, &image_nonce)?;
+
+                // Save the decrypted files locally
+                let image_pointer = self
+                    .config
+                    .data_dir
+                    .join("group_images")
+                    .join(sha256.to_string());
+                fs::write(image_pointer.clone(), &decrypted_bytes).await?;
```
jgmontoya, line 576: This could be extracted to a 'download_file' or 'download_image' method or something like that

PR #340, opened 2025-09-02
Feat/global user subs

jgmontoya:
This PR implements a global subscription system to track and update user metadata and relay lists, extending beyond account-specific processing. This enables tracking metadata changes for all users in the network, not just local accounts + follows.

### Stale Users in Subscription Batches

**Issue Description:**
The current implementation has two related issues where users remain in old subscriptions:

**Issue 1: Relay Migration** (Minor Impact)
When a user changes their NIP-65 relay list (e.g., removes relay A and adds relay B), the current implementation:

1. ✅ Correctly adds the user to new relay B subscriptions during refresh
2. ❌ Does NOT remove the user from old relay A subscriptions

**Issue 2: Batch Boundary Changes** (Moderate Impact - More Serious)
When the number of users for a relay crosses batch boundaries (1000, 2000, 3000, etc.), the batch count changes, causing users to be redistributed to different batch IDs via the modulo operation:

1. ✅ New batch subscriptions are created with correct user distribution
2. ❌ Old batch subscriptions retain their previous user lists
3. ❌ **Result: Users appear in multiple batches on the same relay simultaneously**

**Example Scenarios:**

**Scenario 1 - Relay Migration (Minor):**

```
1. User has relay A → added to global_users_relayA_batch0
2. User updates: removes relay A, adds relay B
3. Refresh triggered → user added to global_users_relayB_batch1
4. Result: User remains in BOTH relay A and relay B subscriptions
```

**Scenario 2 - Batch Boundary Changes (More Serious):**

```
1. Relay A has 999 users → 1 batch (batch_count = 1)
   User X: hash(pubkey) % 1 = 0 → global_users_relayA_0

2. Users grow to 1001 → 2 batches (batch_count = 2)
   User X: hash(pubkey) % 2 = might be 1 → global_users_relayA_1
   BUT User X still remains in global_users_relayA_0!

3. Result: User X receives events from BOTH batch 0 and batch 1 subscriptions
   → Duplicate event processing for the same user
```

We can accept this for now (still at startup/setup they would be cleaned, so this would only matter if the app is left open for long and/or across boundaries).

Also, the problem "self-heals" when other users on the previous relays make changes that trigger a subscription refresh.

In any case, there's a simple solution that we could take: when we're right at the boundary for changing the number of batches, update all the batches of that relay. This would mean more processing at some points, but would eliminate the problem of churning a big number of subscriptions.

### Core Infrastructure

- **Database Migration**: Made `account_id` nullable in `processed_events` table to support global event processing alongside account-specific events
- **Dual Event Processing**: Added separate `AccountEventProcessor` and `GlobalEventProcessor` to handle both account and external user events

### Global Subscriptions System

- **Setup Method**: Added `setup_global_users_subscriptions()` to subscribe to all users' metadata and relay list updates across their known relays
- **Refresh Method**: Added `refresh_global_users_subscriptions()` to update subscriptions as users and relays change
- **User Discovery**: Added `User.all()` and `User.all_users_with_relay_urls()` methods for user enumeration

### Enhanced Testing

- **Unified Test Case**: Replaced separate metadata/relay test cases with a single `PublishSubscriptionUpdateTestCase` using builder pattern
- **External User Testing**: Added comprehensive testing for both account-based and external user updates
- **Improved Test Cleanup**: Enhanced scenario cleanup with `wipe_database()` and `reset_nostr_client()`

## Summary by CodeRabbit

* **New Features**
  * Batched per-relay global user subscriptions; automatic refresh on account activation; helpers to enumerate users with relay URLs and refresh per-user global subscriptions.

* **Bug Fixes**
  * Improved duplicate-event prevention with distinct per-account and global tracking; existence checks now support global (account-agnostic) lookups; safer cleanup/error propagation in integration cleanup.

* **Refactor**
  * Split event processing into dedicated account vs global handlers; subscription and relay handling moved to relay-URL based batching and connections.

* **Chores**
  * Database migration to allow account-agnostic processed events and enforce appropriate uniqueness; added new error variant for invalid input.

* **Tests**
  * Consolidated subscription-update tests into a builder-style suite; added user/relay enumeration and updated processed/published event tests.


db_migrations/0012_make_account_id_nullable_on_processed_events.sql
```diff
@@ -0,0 +1,43 @@
...
+
+CREATE TABLE processed_events_new (
+    id INTEGER PRIMARY KEY AUTOINCREMENT,
+    event_id TEXT NOT NULL
+        CHECK (length(event_id) = 64 AND event_id GLOB '[0-9a-fA-F]*'), -- 64-char hex
+    account_id INTEGER,                   -- Now nullable - which account was it processed for (NULL for global)
+    created_at DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
+
+    FOREIGN KEY (account_id) REFERENCES accounts(id) ON DELETE CASCADE,
+    UNIQUE(event_id, account_id)          -- Each account can only process a specific event once (allows multiple NULL account_ids per event_id)
```
delcin-raj, line 17: Just for clarity.
Every event has a Unique event id. What events would be common to multiple accounts?

erskingardner: MLS messages. If two different accounts are in the same group they both need to process the event.

src/nostr_manager/subscriptions.rs
```diff
@@ -20,6 +22,101 @@ impl NostrManager {
         format!("{:x}", hash)[..12].to_string()
     }
 
+    pub(crate) async fn setup_global_users_subscriptions(
+        &self,
+        users_with_relays: Vec<(PublicKey, Vec<RelayUrl>)>,
```
erskingardner, line 31: What's the users here? The entire list of users that have we have stored in our database?

src/nostr_manager/subscriptions.rs
```diff
@@ -20,6 +22,101 @@ impl NostrManager {
...
+            .await
+    }
+
+    fn pubkey_to_global_user_subscription_id(&self, pubkey: PublicKey) -> SubscriptionId {
+        let pubkey_hex = pubkey.to_hex();
+        let pubkey_hex_prefix = &pubkey_hex[..13.min(pubkey_hex.len())];
+        SubscriptionId::new(format!("{}_global_users", pubkey_hex_prefix))
+    }
+
+    async fn subscribe_pubkey_user_with_id(
```
erskingardner: If I'm reading this right, that this is going to create a new sub for every user that have in the database (for all their various events we care about), this is going to create WAY too many open subscriptions. We'll be immediately rate limited by relays.

src/nostr_manager/subscriptions.rs
jgmontoya: Yes

src/nostr_manager/subscriptions.rs
jgmontoya: You're reading it right. I was not aware of the subscription limit by relays. I'm going to change the approach.

src/nostr_manager/subscriptions.rs
```diff
@@ -20,6 +25,192 @@ impl NostrManager {
...
+
+    /// Deterministic batch assignment: hash(pubkey) % batch_count
+    fn user_to_batch_id(&self, pubkey: &PublicKey, batch_count: usize) -> usize {
+        let mut hasher = Sha256::new();
+        hasher.update(pubkey.to_bytes());
+        let hash = hasher.finalize();
+        let hash_int = u32::from_be_bytes([hash[0], hash[1], hash[2], hash[3]]);
+        (hash_int as usize) % batch_count
+    }
+
```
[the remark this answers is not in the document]
jgmontoya: We're aware of this (I added it to the PR description). We'll discuss the approach we want to take, or whether we're gonna roll with this as-is.

src/nostr_manager/subscriptions.rs
```diff
@@ -20,6 +25,192 @@ impl NostrManager {
...
+    fn calculate_batch_count(&self, user_count: usize) -> usize {
+        if user_count == 0 {
+            1
+        } else {
+            user_count.div_ceil(MAX_USERS_PER_GLOBAL_SUBSCRIPTION)
+        }
+    }
+
+    /// Deterministic batch assignment: hash(pubkey) % batch_count
+    fn user_to_batch_id(&self, pubkey: &PublicKey, batch_count: usize) -> usize {
```
delcin-raj, line 82: There isn't a need for hashing in this case.
Given that PublicKey will be unique for every user,
build u32 from the bytes of the Public key itself.

You may ignore this comment.

src/nostr_manager/subscriptions.rs
```diff
@@ -20,6 +25,192 @@ impl NostrManager {
...
+
+    async fn create_deterministic_batches_for_relay(
+        &self,
+        relay_url: RelayUrl,
+        users: Vec<PublicKey>,
+    ) -> Result<()> {
+        let batch_count = self.calculate_batch_count(users.len());
+
+        // Group users into deterministic batches
+        let mut batches: Vec<Vec<PublicKey>> = vec![Vec::new(); batch_count];
```
delcin-raj, line 54: I think we need to have `Vec<[u8; MAX_USERS_PER_GLOBAL_SUBSCRIPTION]>`

erskingardner: why?

src/nostr_manager/subscriptions.rs
erskingardner: agree, just use the pubkey as a u32 number. no need to hash.

src/whitenoise/accounts.rs
delcin-raj, line 1: Looks good

src/whitenoise/event_processor/mod.rs
```diff
@@ -281,59 +114,48 @@ impl Whitenoise {
...
+                if let Err(send_err) = sender.send(retry_event).await {
+                    tracing::error!(
+                        target: "whitenoise::event_processor::schedule_retry",
+                        "Failed to requeue event for retry: {}",
+                        send_err
+                    );
+                }
+            });
+        }
     }
```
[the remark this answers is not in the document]
erskingardner: This looks like a good idea.

src/whitenoise/event_tracker.rs
```diff
@@ -25,19 +25,37 @@ pub trait EventTracker: Send + Sync {
...
+        &self,
+        event_id: &EventId,
+    ) -> Result<bool, Box<dyn std::error::Error + Send + Sync>>;
+
+    /// Track that we processed a specific event for an account
+    async fn track_processed_acount_event(
         &self,
         event_id: &EventId,
         pubkey: &PublicKey,
     ) -> Result<(), Box<dyn std::error::Error + Send + Sync>>;
```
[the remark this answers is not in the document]
erskingardner: Typo to fix.

erskingardner (changes requested):

src/nostr_manager/subscriptions.rs
delcin-raj: my understanding is that every batch should have no more than 1000 users.

src/nostr_manager/subscriptions.rs
jgmontoya: It probably won't make much of a difference really, but the purpose of hashing here is to try to have good 'spread' of public keys in batches. I'd keep it as a sanity measure.

jgmontoya:
> @jgmontoya - just to double check. users are batched by relay? So if user A has relay X & relay Y, user A's pubkey will show up in the respective sub for each of those relays?

Correct

src/whitenoise/event_processor/mod.rs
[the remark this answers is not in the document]
jgmontoya: This was just moved from what our previous retry thing was. Do you want me to implement this improvement on this PR or should we just create an issue and do it in the future? I think it's better to do it on a separate PR to limit the scope

src/nostr_manager/mod.rs
```diff
@@ -286,53 +286,53 @@ impl NostrManager {
...
 
         // Track newly added relays for connection
         let mut newly_added_relays = Vec::new();
 
-        for relay in relays.iter() {
+        for relay_url in relay_urls.iter() {
             // Check if we're already connected to this relay by attempting to get its status
-            match self.client.relay(relay.url.clone()).await {
+            match self.client.relay(relay_url.clone()).await {
                 Ok(_) => {
```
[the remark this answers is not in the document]
jgmontoya: it compiles fine

src/whitenoise/event_processor/account_event_processor.rs
```diff
@@ -0,0 +1,300 @@
...
+        let underscore_pos = subscription_id.find('_');
+        if underscore_pos.is_none() {
+            return Err(WhitenoiseError::InvalidEvent(format!(
+                "Invalid subscription ID: {}",
+                subscription_id
+            )));
+        }
+
+        let hash_str = &subscription_id[..underscore_pos.unwrap()];
+        // Get all accounts and find the one whose hash matches
```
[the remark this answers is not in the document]
jgmontoya: I don't think we really need this

src/nostr_manager/subscriptions.rs
jgmontoya: distribution of users into batches is not perfect, I'd rather keep it flexible to avoid issues

src/whitenoise/event_processor/mod.rs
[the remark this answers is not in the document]
jgmontoya: @coderabbit create an issue for us to handle this on a separate PR

Issue #210, opened 2025-06-17
Update subscriptions on group changes


erskingardner:
When a user's groups changes, we need to update the relevant subscriptions to make sure that we're still watching for the right events at the right time. 

- Create group - make sure to add new nostr_group_id
- Accept invite - make sure to add new nostr_group_id
- Leave group - make sure to remove nostr_group_id
- Group nostr_group_id changes - update the nostr_group_id in the sub
- There might be others...

delcin-raj:
> [@delcin-raj](https://github.com/delcin-raj) can you check these over to make sure we're updating the group_ids values?

https://github.com/parres-hq/whitenoise/pull/249

Let me know whether I'm on the right track. Will be adding tests for other functions as well.

erskingardner:
Going to close for now and we'll handle in the archive and delete work

PR #344, opened 2025-09-07
Feat/account subscription refresh

jgmontoya:
Refreshes subscriptions on relay changes

Addresses #210 

Note:

* Because the 'leave_group' functionality is still in progress, this is still one part where we should update the subscriptions.
* Group creation new subscription was already being handled (by calling `NostrManager#setup_group_messages_subscriptions_with_signer`) the same is the case for `accept_welcome`

Question:

* Should we also update the account subscriptions on mls commit messages due to the possibility of the nostr_group_id changing?

## Summary by CodeRabbit

* **New Features**
  * Subscriptions auto-refresh when your relay list changes.
  * Added a manual refresh action to update an account’s subscriptions.
  * Logout now attempts to clean up account-specific subscriptions.

* **Behavior Changes**
  * Group message subscriptions now consistently use relay URLs across devices.
  * Event syncing can start from a recent timestamp to reduce catch-up noise.

* **Reliability Improvements**
  * Subscription updates use a short buffer and improved logging to avoid missed events during relay changes.

### DIFF
--- a/src/nostr_manager/mod.rs
+++ b/src/nostr_manager/mod.rs
@@ -9,8 +9,7 @@ use tokio::sync::mpsc::Sender;
 use crate::{
     types::ProcessableEvent,
     whitenoise::{
-        accounts::Account, database::DatabaseError, event_tracker::EventTracker, relays::Relay,
-        Whitenoise,
+        accounts::Account, database::DatabaseError, event_tracker::EventTracker, Whitenoise,
     },
 };
 
@@ -192,9 +191,9 @@ impl NostrManager {
     pub(crate) async fn setup_account_subscriptions_with_signer(
         &self,
         pubkey: PublicKey,
-        user_relays: &[Relay],
-        inbox_relays: &[Relay],
-        group_relays: &[Relay],
+        user_relays: &[RelayUrl],
+        inbox_relays: &[RelayUrl],
+        group_relays: &[RelayUrl],
         nostr_group_ids: &[String],
         signer: impl NostrSigner + 'static,
     ) -> Result<()> {
@@ -210,6 +209,7 @@ impl NostrManager {
                 inbox_relays,
                 group_relays,
                 nostr_group_ids,
+                None,
             )
             .await;
         self.client.unset_signer().await;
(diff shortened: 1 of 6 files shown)


src/whitenoise/groups.rs
```diff
@@ -178,15 +178,15 @@ impl Whitenoise {
...
+        for relay_url in &group_relays {
+            let db_relay = self.find_or_create_relay_by_url(relay_url).await?;
             relays.insert(db_relay);
         }
 
         self.nostr
             .setup_group_messages_subscriptions_with_signer(
                 creator_account.pubkey,
-                &relays.into_iter().collect::<Vec<_>>(),
+                &group_relays.into_iter().collect::<Vec<_>>(),
```
erskingardner: Why isn't this `&relays` since you're compacting them into a hashset above?

src/whitenoise/event_processor/event_handlers/handle_relay_list.rs
```diff
@@ -12,8 +12,34 @@ impl Whitenoise {
...
+        let relays_changed = user.sync_relay_urls(self, relay_type, &relay_urls).await?;
+
+        if relays_changed {
+            self.handle_subscriptions_refresh(&user, &event).await;
+        }
 
         Ok(())
     }
+
+    async fn handle_subscriptions_refresh(&self, user: &User, event: &Event) {
```
erskingardner, line 24: Where's the return type on this method?

src/whitenoise/accounts.rs
```diff
@@ -804,64 +815,66 @@ impl Whitenoise {
...
-            .map(|group| hex::encode(group.nostr_group_id))
-            .collect::<Vec<String>>();
+            .map(|r| r.url)
+            .collect();
+
+        let (group_relays_urls, nostr_group_ids) =
+            self.extract_groups_relays_and_ids(account).await?;
+
+        // Ensure group relays are in the database
+        for relay_url in &group_relays_urls {
```
erskingardner, line 864: I don't have a strong opinion on this but this "ensure group relays are in the database" feels like it could belong in the `extract_group_relays_and_ids` method. What were your thoughts on why it should be here?

erskingardner: Ah - looking at the method below, ensuring the relays are in the database really only needs to happen on setup, not refresh... is that why?

src/nostr_manager/mod.rs
```diff
@@ -219,13 +219,51 @@ impl NostrManager {
...
+                pubkey,
+                user_relays,
+                inbox_relays,
+                group_relays,
+                nostr_group_ids,
+                Some(buffer_time),
+            )
             .await;
         self.client.unset_signer().await;
         result
```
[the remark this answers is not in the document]
erskingardner: Feels like a good change to make.

erskingardner (changes requested):

src/nostr_manager/mod.rs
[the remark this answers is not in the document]
jgmontoya: I'm not sure what this change achieves 😬  @coderabbitai please explain further which purpose this suggestion actually achieves. We know for a fact that subtracting 10 seconds from the current timestamp wont underflow.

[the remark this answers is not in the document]
jgmontoya: That's right, I missed the `?` on the unsubscribe when looking at this again 👍

src/whitenoise/event_processor/event_handlers/handle_relay_list.rs
jgmontoya: This does't return anything

image

src/whitenoise/groups.rs
jgmontoya: Because here we need the `RelayUrl`s instead of the `Relay` object. I'm gonna take it from the relays though to make it clearer (and avoid any possible duplication, although I don't think there would be any)

src/whitenoise/accounts.rs
jgmontoya: Yeah, when the refresh method is hit we already know that the relays are there (because any new relays would've been created just before)

src/whitenoise/event_processor/event_handlers/handle_relay_list.rs
erskingardner: TIL. I didn't realize you could just leave methods completely blank like that...

erskingardner (approved):

PR #350, opened 2025-09-09
Feat/concurrent global subscriptions

jgmontoya:


erskingardner (approved):
Looks good, though I don't know it'll make much difference.

PR #354, opened 2025-09-11
Add methods to manage key_packages

erskingardner:
I would like to have a section in the developer settings screen where you can inspect and delete all your key packages that have been published to relays. 

This might end up in a key package settings screen in the future to help users clean up the same.

I believe we might also need to build a little modal in the app to get people to publish a new key package - many early user's will have key packages that are outdated at this point (I ran into this error yesterday testing an account). 

## Summary by CodeRabbit

- New Features
  - New public operations to publish, fetch, delete single, and bulk-delete an account's MLS key packages from configured relays, with optional local key cleanup and per-item/result logging.

- Bug Fixes
  - Clearer error when an account has no key-package relays configured.
  - Minor efficiency improvements reducing unnecessary cloning in relay checks and tests.
  - Key-package publish now only occurs when a prior deletion actually completed, avoiding redundant publishes.

### DIFF
--- a/src/whitenoise/key_packages.rs
+++ b/src/whitenoise/key_packages.rs
@@ -120,7 +120,10 @@ impl Whitenoise {
     /// - Failed to retrieve account's key package relays
     /// - Network error while fetching events from relays
     /// - NostrSDK error during event streaming
-    pub async fn fetch_all_key_packages_for_account(&self, account: &Account) -> Result<Vec<Event>> {
+    pub async fn fetch_all_key_packages_for_account(
+        &self,
+        account: &Account,
+    ) -> Result<Vec<Event>> {
         let key_package_relays = account.key_package_relays(self).await?;
         let relay_urls: Vec<RelayUrl> = key_package_relays.iter().map(|r| r.url.clone()).collect();
 
@@ -234,7 +237,6 @@ impl Whitenoise {
 
         // Delete from relays (always happens regardless of delete_mls_stored_keys)
         for event in &key_package_events {
-
             // Publish deletion event
             match self
                 .nostr


src/whitenoise/event_processor/event_handlers/handle_giftwrap.rs
```diff
@@ -64,16 +64,20 @@ impl Whitenoise {
...
 
-            self.publish_key_package_for_account(account).await?;
-            tracing::debug!(target: "whitenoise::event_processor::process_welcome", "Published new key package");
+            if deleted {
+                tracing::debug!(target: "whitenoise::event_processor::process_welcome", "Deleted used key package from relays");
+                self.publish_key_package_for_account(account).await?;
+                tracing::debug!(target: "whitenoise::event_processor::process_welcome", "Published new key package");
+            } else {
+                tracing::debug!(target: "whitenoise::event_processor::process_welcome", "Key package already deleted, skipping publish");
+            }
```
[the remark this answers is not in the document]
erskingardner: No, this is actually the behavior we want.

src/whitenoise/event_processor/event_handlers/handle_giftwrap.rs
```diff
@@ -64,16 +64,21 @@ impl Whitenoise {
...
-            tracing::debug!(target: "whitenoise::event_processor::process_welcome", "Published new key package");
+            let deleted = self
+                .delete_key_package_for_account(
+                    account,
+                    &key_package_event_id,
+                    false, // For now we don't want to delete the key packages from MLS storage
+                )
+                .await?;
+
+            if deleted {
```
erskingardner, line 75: Previously we were publishing no matter what here which meant that if there were multiple giftwraps that we processed (that referenced the same key package) then we'd end up publishing way more key packages than we needed. Now we only publish a new key package is we actually removed one from relays.

src/whitenoise/key_packages.rs
```diff
@@ -75,18 +85,197 @@ impl Whitenoise {
...
+                return Err(WhitenoiseError::AccountMissingKeyPackageRelays);
+            }
+
+            let result = self
+                .nostr
+                .publish_event_deletion_with_signer(&event.id, &key_package_relays, signer)
                 .await?;
+            if !result.success.is_empty() {
+                return Ok(true);
+            }
```
jgmontoya: How about we replace this with:

```rs
return Ok(!result.success.is_empty());
```

Then because this branch always returns we can remove the `else` and keep just a single `Ok(false)` at the end

src/whitenoise/key_packages.rs
```diff
@@ -75,18 +85,197 @@ impl Whitenoise {
...
+    /// Returns a vector of all key package events found for the account.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if:
+    /// - Account has no key package relays configured
+    /// - Failed to retrieve account's key package relays
+    /// - Network error while fetching events from relays
+    /// - NostrSDK error during event streaming
+    pub async fn find_all_key_packages_for_account(&self, account: &Account) -> Result<Vec<Event>> {
```
jgmontoya: Should we maybe rename this to `fetch_...` to make it obvious that we're going to relays/external network call to get the info?

src/whitenoise/key_packages.rs
```diff
@@ -75,18 +85,197 @@ impl Whitenoise {
...
+
+        let key_package_filter = Filter::new()
+            .kind(Kind::MlsKeyPackage)
+            .author(account.pubkey);
+
+        let mut key_package_stream = self
+            .nostr
+            .client
+            .stream_events_from(relay_urls, key_package_filter, Duration::from_secs(10))
+            .await?;
```
jgmontoya, line 142: This feels a bit out of place. It's the kind of thing I'd expect to have some wrapper function within `nostr_manager`

src/whitenoise/key_packages.rs
```diff
@@ -75,18 +85,197 @@ impl Whitenoise {
...
+
+        if key_package_relays.is_empty() {
+            return Err(WhitenoiseError::AccountMissingKeyPackageRelays);
+        }
+
+        let mut deleted_count = 0;
+
+        for event in &key_package_events {
+            if delete_mls_stored_keys {
+                match Account::create_nostr_mls(account.pubkey, &self.config.data_dir) {
```
jgmontoya: I don't like that this is within the loop. We should instance once outside, not for every event.

src/whitenoise/key_packages.rs
erskingardner: yup. good call.

src/whitenoise/key_packages.rs
erskingardner: going to leave for now.

PR #360, opened 2025-09-12
Normalize relay urls and remove limit(1) from filter

erskingardner:
I keep seeing older relay lists and metadata showing up in the UI. While the nostr-sdk does try and order things by created at, it seems that the `limit(1)` will bail early out of the stream fetching events as soon as it sees a single event. So we don't limit now and do a quick sort to be absolutely sure we are getting the most recent events.

I was also seeing very odd behavior trying to add relays because we weren't normalizing the trailing slash off of the relay urls. 

## Summary by CodeRabbit

- Bug Fixes
  - Uses the latest profile metadata, relay lists, and key packages (selects most recent events and ignores events timestamped far in the future).
  - Eliminates duplicate relays caused by trailing slashes, preventing duplicate entries and lookup issues.

- Chores
  - Adds a migration to normalize relay URLs and consolidate existing duplicates.
  - Standardizes relay URL handling during save/find to ensure consistent storage and lookups.

- Tests
  - Adds tests verifying relay URL normalization, deduplication, and retrieval across URL variants.

### DIFF
--- a/src/whitenoise/database/relays.rs
+++ b/src/whitenoise/database/relays.rs
@@ -32,11 +32,7 @@ fn normalize_relay_url(url: &RelayUrl) -> String {
     let url_str = url.to_string();
 
     // Remove trailing slash if present
-    if url_str.ends_with('/') {
-        url_str.trim_end_matches('/').to_string()
-    } else {
-        url_str
-    }
+    url_str.trim_end_matches('/').to_string()
 }
 
 #[derive(Debug, PartialEq, Eq, Clone, Hash)]
--- a/src/nostr_manager/query.rs
+++ b/src/nostr_manager/query.rs
@@ -1,6 +1,6 @@
 //! This module contains functions for querying Nostr events from relays.
 
-use std::collections::HashSet;
+use std::{collections::HashSet, time::Duration};
 
 use nostr_sdk::prelude::*;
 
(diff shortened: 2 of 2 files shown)


db_migrations/0013_normalize_relay_urls.sql
```diff
@@ -0,0 +1,43 @@
...
+-- Since most WebSocket/relay implementations treat URLs with and without trailing
+-- slashes as equivalent, normalizing to no trailing slash provides the cleanest
+-- and most predictable behavior
+
+-- Remove trailing slashes from all URLs
+UPDATE relays
+SET url = RTRIM(url, '/')
+WHERE url LIKE '%/';
+
+-- Handle any potential duplicates that might have been created
```
jgmontoya: Regarding this and duplicates: My guess is that the only way there could be duplicates is if they're there already in the form of: `relay.com` and `relay.com/` correct? And in that case, won't the unique constraint prevent the update from `relay.com/` to `relay.com`??

src/whitenoise/database/relays.rs
```diff
@@ -4,6 +4,41 @@ use nostr_sdk::RelayUrl;
...
+/// - `"wss://relay.com/path"` stays `"wss://relay.com/path"`
+fn normalize_relay_url(url: &RelayUrl) -> String {
+    let url_str = url.to_string();
+
+    // Remove trailing slash if present
+    if url_str.ends_with('/') {
+        url_str.trim_end_matches('/').to_string()
+    } else {
+        url_str
+    }
```
jgmontoya: I don't think the if/else here is needed:

This should work, right?

```rs
url_str.trim_end_matches('/').to_string()
```

db_migrations/0013_normalize_relay_urls.sql
jgmontoya: coderabbit beat me to it lol

PR #362, opened 2025-09-12
perf(nostr_manager): increase parallelization and reduce scope of signer lock

jgmontoya:
Parallelize ensure_relays_connected, increase parallelization on batch subscription processing and reduce scope of signer lock.

## Summary by CodeRabbit

- New Features
  - View current relay connection status for any relay.

- Improvements
  - Unified signer-aware subscription flows for account and group operations.
  - Centralized, parallel relay connection handling for faster, more reliable connections.
  - Group message subscriptions now use group relays as the source of truth.

- Bug Fixes
  - Fewer subscription failures and “no relay connection” errors via improved batching, aggregation, and error reporting.

- Removed
  - Deprecated signer-wrapped subscription helpers removed.

### DIFF
--- a/src/nostr_manager/mod.rs
+++ b/src/nostr_manager/mod.rs
@@ -389,59 +389,10 @@ impl NostrManager {
             relay_urls.len()
         );
 
-        // Track newly added relays for connection
-        let mut newly_added_relays = Vec::new();
-
-        for relay_url in relay_urls.iter() {
-            // Check if we're already connected to this relay by attempting to get its status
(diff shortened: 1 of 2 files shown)


josefinalliende:
Tested this in ios (emulator an physical) and splashcreen now works really fast for me ⚡ ⚡ 

Account creation in ios feels faster, but not 100% sure cause this issue was flaky. But tried it a bunch of times and takes like 4s, which is not fast but maybe we can be ok for this release?

PR #363, opened 2025-09-13
Media manager to handle media in Chat Messages

delcin-raj:
## Summary by CodeRabbit

- New Features
  - End-to-end encrypted media messaging: send files with captions and decrypt on receive.

- Improvements
  - Per-group on-disk media cache for faster loads and offline tolerance.
  - Integrated Blossom-backed uploads/downloads for media flows.

- Database
  - Migration drops redundant media columns to streamline storage.

- Refactor
  - Media subsystem rebuilt under a new manager with consolidated sanitizer, encryption, cache, and error handling.

- Chores
  - Updated dependency revisions and enabled additional SDK features.


src/whitenoise/messages.rs
```diff
@@ -122,7 +122,7 @@ impl Whitenoise {
     }
 
     /// Creates an unsigned nostr event with the given parameters
-    fn create_unsigned_nostr_event(
+    pub(crate) fn create_unsigned_nostr_event(
```
jgmontoya: Why make this `pub(crate)`??

src/whitenoise/media_manager/cache.rs
```diff
@@ -0,0 +1,159 @@
...
+                .data_dir
+                .to_str()
+                .ok_or(WhitenoiseError::Other(anyhow!(
+                    "Unable to convert PathBuf to string"
+                )))?,
+            MEDIA_CACHE_DIR,
+            hex::encode(group_id.as_slice()),
+            hex::encode(file_hash)
+        ))
+    }
```
jgmontoya: I'm not sure if scoping it to a group / group_id is needed/such a good idea imo. If a user sends/receives the exact same file from multiple groups (say for example, it received it from one group and is sending it to another) then it will need to locally store the file multiple times. This seems unnecessary.

src/whitenoise/media_manager/cache.rs
```diff
@@ -0,0 +1,159 @@
...
+    ///
+    /// # Arguments
+    /// * `mls_group_id` - The MLS group ID
+    /// * `file_hash` - The hash of the file
+    /// * `db` - Database connection
+    ///
+    /// # Returns
+    /// * `Ok(Option<CachedMediaFile>)` - The cached media file if found
+    /// * `Err(MediaError)` - Error if fetch fails or file not found
+    pub(crate) async fn fetch_cached_file(
```
jgmontoya, line 84: Same with this with regards to the group_id, if the user is requesting for a specific file_hash, then I see no reason to scope it to a specific group

db_migrations/0014_drop_redundant_columns_media_files.sql
```diff
@@ -0,0 +1,5 @@
+-- Migration: Drop nostr_key column from media_files
+ALTER TABLE media_files DROP COLUMN nostr_key;
+ALTER TABLE media_files DROP COLUMN blossom_url;
+ALTER TABLE media_files DROP COLUMN file_path;
```
jgmontoya, line 4: If I recall correctly, the reason these columns exist were to help us track things. If we receive some file but still haven't downloaded it we can have a null file_path, and we know to get it from that blossom url.

Similarly, if a file had file_path but didn't have blossom_url we would know that we need to upload it.

How would things like that work on this new way of handling media?

db_migrations/0014_drop_redundant_columns_media_files.sql
```diff
@@ -0,0 +1,5 @@
...
+ALTER TABLE media_files DROP COLUMN file_metadata;
```
jgmontoya, line 5: I think we want to keep this file_metadata. This is useful because we'll end up replacing things like the name of the file with some hash-based name that's internal for whitenoise but likely meaningless for the user. If the user decides at some point to "export the file out of whitenoise" or something like that, we should preserve things like the original file name.

src/whitenoise/media_manager/cache.rs
```diff
@@ -0,0 +1,159 @@
...
+            mls_group_id, account_pubkey,
+            file_hash, created_at
+        ) VALUES (?, ?, ?, ?) RETURNING *",
+        )
+        .bind(group_id.as_slice())
+        .bind(account_pubkey.to_string())
+        .bind(hex::encode(encrypted_file_hash))
+        .bind(created_at as i64)
+        .fetch_one(&self.database.pool)
+        .await?;
```
jgmontoya, line 69: We have to be mindful of how we handle errors to avoid syncing issues. For example here, if something fails we will leave the file orphaned in the cache

src/whitenoise/media_manager/mod.rs
```diff
@@ -0,0 +1,356 @@
...
+        };
+        Self { blossom }
+    }
+    /// Adds a media file, ready to be used in a chat.
+    ///
+    /// This method handles the complete workflow for adding a media file:
+    /// 1. Encrypts the file using ChaCha20-Poly1305
+    /// 2. Uploads the encrypted file to Blossom
+    /// 3. Caches the original file locally
+    /// 4. Generates an IMETA tag with file metadata
```
jgmontoya, line 74: This docstring seems out of date to what this method is actually doing. Let's either remove it or make sure it's accurate

src/whitenoise/media_manager/mod.rs
```diff
@@ -0,0 +1,356 @@
...
+                }
+
+                if letter == SingleLetterTag::lowercase(Alphabet::N) {
+                    nonce = tag
+                        .content()
+                        .map(|s| hex::decode(s).map_err(MediaError::from))
+                        .transpose()?;
+                }
+            }
+        }
```
jgmontoya, line 219: If there are multiple `m`, `x` or `n` tags, we're going to be using the last of each. Is that the correct behavior?

src/whitenoise/media_manager/mod.rs
```diff
@@ -0,0 +1,356 @@
...
+            (Some(_mime), Some(sha256), Some(nonce)) => {
+                let encrypted_file_hash =
+                    hex::decode(&sha256).map_err(|_| BlossomError::InvalidSha256)?;
+                match self
+                    .fetch_cached_file(group_id, &encrypted_file_hash, self)
+                    .await?
+                {
+                    Some(cached_file) => Ok(cached_file),
+                    None => {
+                        // Download media from blossom
```
jgmontoya: We should probably have the download logic be a method that we just call here

src/whitenoise/media_manager/mod.rs
```diff
@@ -0,0 +1,356 @@
...
+        &self,
+        account: &Account,
+        group_id: &GroupId,
+        file: FileDetails,
+        caption: String,
+    ) -> Result<MessageWithTokens> {
+        // Encrypt the file data
+        // Get the raw secret key bytes
+        let nostr_mls = Account::create_nostr_mls(account.pubkey, &self.config.data_dir)?;
+        let exporter_secret = nostr_mls.exporter_secret(group_id)?;
```
jgmontoya, line 270: My understanding is that the exporter secret in MLS should change on every epoch. How are we handling that? Do we need to take any considerations for that to not break the functioning of media or would that already be handled?

src/whitenoise/media_manager/mod.rs
```diff
@@ -0,0 +1,356 @@
...
+
+impl MediaManager {
+    pub fn new() -> Self {
+        let blossom = if cfg!(debug_assertions) {
+            BlossomClient::new(Url::parse("http://localhost:3000").unwrap())
+        } else {
+            BlossomClient::new(Url::parse("https://blossom.primal.net/").unwrap())
+        };
+        Self { blossom }
+    }
```
jgmontoya, line 67: I think we should have the `new` method be able to receive the BlossomClient "from outside" (dependency injection) and have the default call new with the actual default.

This will allow us to more easily test things, and also I think will make it easier to use the MediaManager with multiple/other blossom servers in the future. What do you think?

db_migrations/0014_drop_redundant_columns_media_files.sql
delcin-raj: File path can always be derived from file_hash.
We have decided to use a single blossom server for now.

delcin-raj: We can get nostr_key from secrets_store

db_migrations/0014_drop_redundant_columns_media_files.sql
delcin-raj: This one was done based on the following reasons.

FileMetadata cannot be retreived from Blossom. `get_blob` only returns the blob in bytes format.

FileMetada information will be stored in the nostr event with `Kind::FileMetadata` and that message will be stored in the nostr-mls

src/whitenoise/media_manager/cache.rs
delcin-raj: I agree with that, but I would appreciate @erskingardner's take on this.

src/whitenoise/media_manager/cache.rs
delcin-raj: The file will be stored in a single location only. But as you have noted, it will be created multiple times for different groups. We might not need to scope it to a single group.

src/whitenoise/media_manager/mod.rs
delcin-raj: This event will be created explicitly by whitenoise clients.
Here, the expectation is that there is precisely one of these tags.

src/whitenoise/media_manager/mod.rs
delcin-raj: wip

src/whitenoise/media_manager/cache.rs
delcin-raj: `add_to_cache` is called only after upload/download w.r.t. blossom succeeds.

jgmontoya: Yes, what I'm asking is what happens if the file was already created but the insertion to the table fails for some reason

src/whitenoise/media_manager/mod.rs
jgmontoya: Ok! Whenever we do the specification we should perhaps clarify that multiple `m`, `x` or `n` tags lead to undefined behavior. To me it'd be more natural to default on using the first one rather than the last one but I don't think it really matters

db_migrations/0014_drop_redundant_columns_media_files.sql
jgmontoya: > File path can always be derived from file_hash.

The point is that it serves as a way to monitor syncing

> We have decided to use a single blossom server for now.

Even if we currently use a single default blossom server (which I'm not completely sure why we do btw) we definitely will want to have the ability to use multiple in the future, so I think we should plan a bit ahead and make sure the design won't restrict us too much

> We can get nostr_key from secrets_store

If I recall correctly (and I don't have a very good memory so take this with a grain of salt) the original design used single-use keys for encrypting media. Which I believe is what we stored here. I'm still not so sure if we're going to move towards using the MLS exporter_secret, but even if we do I think we might need to store the decryption key anyway because that might change and I don't think we want to lose the ability to decrypt previous media.

db_migrations/0014_drop_redundant_columns_media_files.sql
jgmontoya: > and that message will be stored in the nostr-mls

I was not aware of this. We definitely need to document and agree better on the design and what goes where

db_migrations/0014_drop_redundant_columns_media_files.sql
delcin-raj: This is a chat media. This should be decryptable by every member of the group.

My assumption is exporter secret is common for every group members.

Even if we are going to use a fresh key to encrypt every image that has to be shared in the nostr-mls event, so that it can be accessed by all the members.

I'm still improving my understanding of the standard.

src/whitenoise/media_manager/cache.rs
delcin-raj: It's highly unlikely that this SQL query will fail at runtime.
Even if that's the case it will trigger `download_from_blossom` function again.

src/whitenoise/media_manager/mod.rs
delcin-raj: I would defer this for the future.

PR #364, opened 2025-09-14
Track event created time on replaceable events

erskingardner:
This PR adds tracking the `created_at` time of the replaceable event into a field on `User`, `UserRelay`, and `UserFollow` rows in the database to be sure that we don't overwrite our local data with stale or outdated data.

### DIFF
--- a/src/nostr_manager/mod.rs
+++ b/src/nostr_manager/mod.rs
@@ -500,18 +500,46 @@ impl NostrManager {
             mut group_messages,
         ) = self
             .with_signer(signer, || async {
-                let mut contacts_and_self =
-                    match self.client.get_contact_list_public_keys(self.timeout).await {
-                        Ok(contacts) => contacts,
-                        Err(e) => {
-                            tracing::error!(
-                                target: "whitenoise::nostr_manager::fetch_all_user_data",
-                                "Failed to get contact list public keys: {}",
-                                e
-                            );
-                            return Err(NostrManagerError::Client(e));
-                        }
-                    };
+                // Get whitenoise instance for database access
+                let whitenoise = Whitenoise::get_instance()
+                    .map_err(|e| NostrManagerError::EventProcessingError(e.to_string()))?;
+
+                // Get user's relay information for contact list fetching
+                let user_relays = match account.nip65_relays(whitenoise).await {
+                    Ok(relays) => relays.into_iter().map(|r| r.url).collect::<Vec<_>>(),
+                    Err(_) => {
+                        // Fallback to default relays if user has no specific relays
+                        vec![
+                            RelayUrl::parse("wss://relay.damus.io").unwrap(),
+                            RelayUrl::parse("wss://nos.lol").unwrap(),
+                            RelayUrl::parse("wss://relay.nostr.band").unwrap(),
+                        ]
+                    }
+                };
+
+                // Fetch the account's contact list using our custom method
+                let mut contacts_and_self = match self.fetch_contact_list_events(account.pubkey, &user_relays).await {
+                    Ok(Some(contact_list_event)) => {
+                        Self::pubkeys_from_event(&contact_list_event)
+                    },
+                    Ok(None) => {
+                        tracing::debug!(
+                            target: "whitenoise::nostr_manager::sync_all_user_data",
+                            "No contact list found for account {}",
+                            account.pubkey.to_hex()
+                        );
+                        Vec::new()
+                    },
+                    Err(e) => {
+                        tracing::error!(
+                            target: "whitenoise::nostr_manager::sync_all_user_data",
+                            "Failed to fetch contact list for account {}: {}",
+                            account.pubkey.to_hex(),
+                            e
+                        );
+                        Vec::new()
+                    }
+                };
                 contacts_and_self.push(account.pubkey);
 
                 let metadata_filter = Filter::new()
--- a/src/nostr_manager/query.rs
+++ b/src/nostr_manager/query.rs
@@ -91,6 +91,33 @@ impl NostrManager {
             .max_by_key(|e| (e.created_at, e.id));
         Ok(latest)
     }
+
+    pub(crate) async fn fetch_contact_list_events(
+        &self,
+        pubkey: PublicKey,
+        relays: &[RelayUrl],
+    ) -> Result<Option<Event>> {
+        let filter = Filter::new().author(pubkey).kind(Kind::ContactList);
+        let events = self
+            .client
+            .fetch_events_from(relays, filter, self.timeout)
+            .await?;
+
+        // Filter out events with timestamps too far in the future
+        let cutoff = Timestamp::now() + MAX_FUTURE_SKEW;
+        let latest = events
+            .into_iter()
+            .filter(|e| e.created_at <= cutoff)
+            .max_by_key(|e| (e.created_at, e.id));
+
+        tracing::debug!(
+            "Fetched contact list events for {}, using latest: {:?}",
+            pubkey.to_hex(),
+            latest.as_ref().map(|e| e.created_at)
+        );
+
+        Ok(latest)
+    }
 }
 
 #[cfg(test)]
--- a/src/nostr_manager/query.rs
+++ b/src/nostr_manager/query.rs
@@ -18,7 +18,7 @@ impl NostrManager {
         &self,
         nip65_relays: &[Relay], // TODO: Replace with &[RelayUrl]
         pubkey: PublicKey,
-    ) -> Result<Option<Metadata>> {
+    ) -> Result<Option<(Metadata, Timestamp)>> {
         let filter: Filter = Filter::new().author(pubkey).kind(Kind::Metadata);
         let urls: Vec<RelayUrl> = nip65_relays.iter().map(|r| r.url.clone()).collect();
         let events: Events = self
@@ -33,7 +33,7 @@ impl NostrManager {
             .filter(|e| e.created_at <= cutoff)
             .max_by_key(|e| (e.created_at, e.id));
         match latest {
-            Some(event) => Ok(Some(Metadata::try_from(&event)?)),
+            Some(event) => Ok(Some((Metadata::try_from(&event)?, event.created_at))),
             None => Ok(None),
         }
     }
@@ -43,7 +43,7 @@ impl NostrManager {
         pubkey: PublicKey,
         relay_type: RelayType,
         nip65_relays: &[Relay], // TODO: Replace with &[RelayUrl]
-    ) -> Result<HashSet<RelayUrl>> {
+    ) -> Result<Option<(HashSet<RelayUrl>, Timestamp)>> {
         let filter = Filter::new().author(pubkey).kind(relay_type.into());
         let urls: Vec<RelayUrl> = nip65_relays.iter().map(|r| r.url.clone()).collect();
         let relay_events = self
@@ -64,8 +64,11 @@ impl NostrManager {
         );
 
         match latest {
-            None => Ok(HashSet::new()),
-            Some(event) => Ok(Self::relay_urls_from_event(event)),
+            None => Ok(None),
+            Some(event) => {
+                let relay_urls = Self::relay_urls_from_event(event.clone());
+                Ok(Some((relay_urls, event.created_at)))
+            }
         }
     }
 
--- a/src/whitenoise/accounts.rs
+++ b/src/whitenoise/accounts.rs
@@ -137,7 +137,7 @@ impl Account {
         whitenoise: &Whitenoise,
     ) -> Result<()> {
         let user = self.user(&whitenoise.database).await?;
-        user.add_relay(relay, relay_type, &whitenoise.database)
+        user.add_relay(relay, relay_type, Some(Utc::now()), &whitenoise.database)
             .await?;
         whitenoise
             .background_publish_account_relay_list(self, relay_type)
@@ -641,15 +641,17 @@ impl Whitenoise {
         relay_type: RelayType,
         source_relays: &[Relay],
     ) -> Result<Vec<Relay>> {
-        let relay_urls = self
+        let relay_result = self
             .nostr
             .fetch_user_relays(pubkey, relay_type, source_relays)
             .await?;
 
         let mut relays = Vec::new();
-        for url in relay_urls {
-            let relay = self.find_or_create_relay_by_url(&url).await?;
-            relays.push(relay);
+        if let Some((relay_urls, _timestamp)) = relay_result {
+            for url in relay_urls {
+                let relay = self.find_or_create_relay_by_url(&url).await?;
+                relays.push(relay);
+            }
         }
 
         Ok(relays)
@@ -1080,11 +1082,11 @@ mod tests {
 
         // Verify that the relay list events were published
         assert!(
-            !inbox_events.is_empty(),
+            inbox_events.is_some() && !inbox_events.unwrap().0.is_empty(),
             "Inbox relays list (kind 10050) should be published for new accounts"
         );
         assert!(
-            !key_package_relays_events.is_empty(),
+            key_package_relays_events.is_some() && !key_package_relays_events.unwrap().0.is_empty(),
             "Key package relays list (kind 10051) should be published for new accounts"
         );
         assert!(
@@ -1361,7 +1363,7 @@ mod tests {
             .await
             .expect("Failed to fetch metadata from relays");
 
-        if let Some(published_metadata) = fetched_metadata {
+        if let Some((published_metadata, _timestamp)) = fetched_metadata {
             assert_eq!(published_metadata.name, new_metadata.name);
             assert_eq!(published_metadata.display_name, new_metadata.display_name);
             assert_eq!(published_metadata.about, new_metadata.about);
--- a/src/whitenoise/database/accounts.rs
+++ b/src/whitenoise/database/accounts.rs
@@ -311,6 +311,117 @@ impl Account {
         Ok(())
     }
 
+    /// Updates this account's follow list from a ContactList event.
+    ///
+    /// This method atomically replaces all current follows with the follows from the
+    /// given ContactList event, tracking the event timestamp to prevent stale overwrites.
+    /// All operations are performed within a transaction to ensure data consistency.
+    ///
+    /// # Arguments
+    ///
+    /// * `contacts` - Vector of public keys from the ContactList event
+    /// * `event_timestamp` - Timestamp of the ContactList event (as u64)
+    /// * `database` - A reference to the `Database` instance for database operations
+    ///
+    /// # Returns
+    ///
+    /// Returns `Ok(Vec<PublicKey>)` containing the public keys of newly created users.
+    ///
+    /// # Errors
+    ///
+    /// Returns a [`WhitenoiseError`] if the database operations fail.
+    pub(crate) async fn update_follows_from_event(
+        &self,
+        contacts: Vec<nostr_sdk::PublicKey>,
+        event_timestamp: u64,
+        database: &Database,
+    ) -> Result<Vec<nostr_sdk::PublicKey>, WhitenoiseError> {
+        use crate::whitenoise::users::User;
+        use chrono::Utc;
+
+        // Start transaction for atomic operations
+        let mut tx = database.pool.begin().await.map_err(DatabaseError::Sqlx)?;
+
+        // Remove all current follows for this account
+        sqlx::query("DELETE FROM account_follows WHERE account_id = ?")
+            .bind(self.id)
+            .execute(&mut *tx)
+            .await
+            .map_err(DatabaseError::Sqlx)?;
+
+        // Store the length before consuming the vector
+        let contacts_len = contacts.len();
+        let mut newly_created_users = Vec::new();
+
+        // Insert new follows with event timestamp
+        for pubkey in contacts {
+            let (user, newly_created) = User::find_or_create_by_pubkey(&pubkey, database).await?;
+
+            if newly_created {
+                // Track newly created users so the caller can fetch their metadata
+                newly_created_users.push(pubkey);
+                tracing::debug!("Created new user for follow: {}", pubkey.to_hex());
+            }
+
+            sqlx::query("INSERT INTO account_follows (account_id, user_id, event_created_at, created_at, updated_at) VALUES (?, ?, ?, ?, ?)")
+                .bind(self.id)
+                .bind(user.id)
+                .bind(event_timestamp as i64)
+                .bind(Utc::now().timestamp_millis())
+                .bind(Utc::now().timestamp_millis())
+                .execute(&mut *tx)
+                .await
+                .map_err(DatabaseError::Sqlx)?;
+        }
+
+        // Commit transaction
+        tx.commit().await.map_err(DatabaseError::Sqlx)?;
+
+        tracing::debug!(
+            target: "whitenoise::database::accounts::update_follows_from_event",
+            "Updated follows for account {} with {} contacts from event timestamp {}, {} newly created users",
+            self.pubkey.to_hex(),
+            contacts_len,
+            event_timestamp,
+            newly_created_users.len()
+        );
+
+        Ok(newly_created_users)
+    }
+
+    /// Gets the newest event timestamp among existing follows for this account.
+    ///
+    /// This method queries the account_follows table to find the latest event_created_at
+    /// timestamp among all follows for this account. This is used to determine if a
+    /// ContactList event is newer than what we already have stored.
+    ///
+    /// # Arguments
+    ///
+    /// * `database` - A reference to the `Database` instance for database operations
+    ///
+    /// # Returns
+    ///
+    /// Returns `Ok(Some(timestamp))` if follows exist and have event timestamps,
+    /// `Ok(None)` if no follows exist or all follows have NULL event_created_at.
+    ///
+    /// # Errors
+    ///
+    /// Returns a [`WhitenoiseError`] if the database query fails.
+    pub(crate) async fn get_latest_follow_list_timestamp(
+        &self,
+        database: &Database,
+    ) -> Result<Option<u64>, WhitenoiseError> {
+        let result = sqlx::query_scalar::<_, Option<i64>>(
+            "SELECT MAX(event_created_at) FROM account_follows WHERE account_id = ?",
+        )
+        .bind(self.id)
+        .fetch_one(&database.pool)
+        .await
+        .map_err(DatabaseError::Sqlx)?;
+
+        Ok(result.map(|ts| ts as u64))
+    }
+
     /// Saves this account to the database.
     ///
     /// # Arguments
--- a/src/whitenoise/database/accounts.rs
+++ b/src/whitenoise/database/accounts.rs
@@ -197,7 +197,7 @@ impl Account {
     /// Returns a [`WhitenoiseError::AccountNotFound`] if the database query fails.
     pub(crate) async fn follows(&self, database: &Database) -> Result<Vec<User>, WhitenoiseError> {
         let user_rows = sqlx::query_as::<_, UserRow>(
-            "SELECT u.id, u.pubkey, u.metadata, u.created_at, u.updated_at
+            "SELECT u.id, u.pubkey, u.metadata, u.created_at, u.updated_at, u.event_created_at
              FROM account_follows af
              JOIN users u ON af.user_id = u.id
              WHERE af.account_id = ?",
@@ -215,6 +215,7 @@ impl Account {
                 metadata: row.metadata,
                 created_at: row.created_at,
                 updated_at: row.updated_at,
+                event_created_at: row.event_created_at,
             })
             .collect();
 
@@ -856,6 +857,7 @@ mod tests {
                 metadata: metadata.clone(),
                 created_at: test_timestamp,
                 updated_at: test_timestamp,
+                event_created_at: None,
             };
 
             // Save user first
@@ -980,6 +982,7 @@ mod tests {
             metadata: user_metadata.clone(),
             created_at: test_timestamp,
             updated_at: test_timestamp,
+            event_created_at: None,
         };
 
         // Save user
@@ -1057,6 +1060,7 @@ mod tests {
             metadata: user_metadata.clone(),
             created_at: test_timestamp,
             updated_at: test_timestamp,
+            event_created_at: None,
         };
 
         user.save(&whitenoise.database).await.unwrap();
@@ -1154,6 +1158,7 @@ mod tests {
                 metadata: user_metadata,
                 created_at: test_timestamp,
                 updated_at: test_timestamp,
+                event_created_at: None,
             };
 
             user.save(&whitenoise.database).await.unwrap();
--- a/src/whitenoise/database/users.rs
+++ b/src/whitenoise/database/users.rs
@@ -1,7 +1,7 @@
 use chrono::{DateTime, Utc};
 use nostr_sdk::{Metadata, PublicKey};
 
-use super::{relays::RelayRow, utils::parse_timestamp, Database, DatabaseError};
+use super::{relays::RelayRow, utils::{parse_timestamp, parse_optional_timestamp}, Database, DatabaseError};
 use crate::{
     whitenoise::{
         relays::{Relay, RelayType},
@@ -22,6 +22,8 @@ pub(crate) struct UserRow {
     pub created_at: DateTime<Utc>,
     // updated_at is the timestamp of the last update
     pub updated_at: DateTime<Utc>,
+    // event_created_at is the timestamp of the original metadata event (None for legacy data)
+    pub event_created_at: Option<DateTime<Utc>>,
 }
 
 impl<'r, R> sqlx::FromRow<'r, R> for UserRow
@@ -51,6 +53,7 @@ where
 
         let created_at = parse_timestamp(row, "created_at")?;
         let updated_at = parse_timestamp(row, "updated_at")?;
+        let event_created_at = parse_optional_timestamp(row, "event_created_at")?;
 
         Ok(UserRow {
             id,
@@ -58,6 +61,7 @@ where
             metadata,
             created_at,
             updated_at,
+            event_created_at,
         })
     }
 }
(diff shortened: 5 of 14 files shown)


src/whitenoise/database/accounts.rs
```diff
@@ -310,6 +311,117 @@ impl Account {
...
+
+        // Start transaction for atomic operations
+        let mut tx = database.pool.begin().await.map_err(DatabaseError::Sqlx)?;
+
+        // Remove all current follows for this account
+        sqlx::query("DELETE FROM account_follows WHERE account_id = ?")
+            .bind(self.id)
+            .execute(&mut *tx)
+            .await
+            .map_err(DatabaseError::Sqlx)?;
```
jgmontoya, line 353: I find it a bit strange that we remove all of them and then create them all instead of just handling the difference. To me the "natural" way would be to insert the ones that we're missing, remove the extra ones and make sure that the ones remaining have the proper `event_created_at`.

Now, because of that last step (updating the `event_created_at` of all) it might be worth it to keep it as-is with just deleting all and inserting all. Not sure.

src/whitenoise/database/accounts.rs
```diff
@@ -310,6 +311,117 @@ impl Account {
...
+    /// * `database` - A reference to the `Database` instance for database operations
+    ///
+    /// # Returns
+    ///
+    /// Returns `Ok(Vec<PublicKey>)` containing the public keys of newly created users.
+    ///
+    /// # Errors
+    ///
+    /// Returns a [`WhitenoiseError`] if the database operations fail.
+    pub(crate) async fn update_follows_from_event(
```
jgmontoya, line 332: We should have some test for this method

src/whitenoise/database/accounts.rs
```diff
@@ -310,6 +311,117 @@ impl Account {
...
+    ///
+    /// # Returns
+    ///
+    /// Returns `Ok(Some(timestamp))` if follows exist and have event timestamps,
+    /// `Ok(None)` if no follows exist or all follows have NULL event_created_at.
+    ///
+    /// # Errors
+    ///
+    /// Returns a [`WhitenoiseError`] if the database query fails.
+    pub(crate) async fn get_latest_follow_list_timestamp(
```
jgmontoya: we should have a test for this method

src/whitenoise/database/utils.rs
```diff
@@ -51,6 +51,41 @@ where
...
+            None => Ok(None),
+        };
+    }
+
+    Err(create_column_decode_error(
+        column_name,
+        "Could not parse as INTEGER or DATETIME",
+    ))
+}
+
```
jgmontoya: I think we can reuse our existent function to reduce the duplicated logic here. What do you think about something like this:

```rs
pub(crate) fn parse_optional_timestamp<'r, R>(
    row: &'r R,
    column_name: &'r str,
) -> Result<Option<DateTime<Utc>>, sqlx::Error>
where
    R: Row,
    &'r str: sqlx::ColumnIndex<R>,
    i64: sqlx::Decode<'r, R::Database> + sqlx::Type<R::Database>,
    String: sqlx::Decode<'r, R::Database> + sqlx::Type<R::Database>,
    Option<i64>: sqlx::Decode<'r, R::Database> + sqlx::Type<R::Database>,
    Option<String>: sqlx::Decode<'r, R::Database> + sqlx::Type<R::Database>,
{
    // Check if the column is NULL for either type
    if let Ok(None) = row.try_get::<Option<i64>, _>(column_name) {
        return Ok(None);
    }
    if let Ok(None) = row.try_get::<Option<String>, _>(column_name) {
        return Ok(None);
    }
    
    // If not NULL, delegate to the existing function
    parse_timestamp(row, column_name).map(Some)
}
```

src/whitenoise/database/utils.rs
```diff
@@ -51,6 +51,41 @@ where
     ))
 }
 
+/// Parses an optional timestamp column, returning None for NULL values
+pub(crate) fn parse_optional_timestamp<'r, R>(
```
jgmontoya: we should add tests for this method

src/whitenoise/event_processor/event_handlers/handle_contact_list.rs
```diff
@@ -30,72 +30,57 @@ impl Whitenoise {
...
-            .difference(&contacts_set)
-            .cloned()
-            .collect();
+        // Check if this event is newer than what we already have
+        let current_event_time = account
+            .get_latest_follow_list_timestamp(&self.database)
+            .await?;
+        let event_timestamp = event.created_at.as_u64() * 1000; // Convert seconds to milliseconds
+
+        if let Some(current_time) = current_event_time {
```
jgmontoya, line 60: if the event doesn't have a created_at then we're moving forward and updating as well, is that correct? is that even possible?

src/whitenoise/event_processor/event_handlers/handle_contact_list.rs
```diff
@@ -30,72 +30,57 @@ impl Whitenoise {
...
 
-        for pubkey in &users_to_unfollow {
-            let (user, created) =
-                crate::whitenoise::users::User::find_or_create_by_pubkey(pubkey, &self.database)
-                    .await?;
-            if created {
+        // Background fetch user data for newly created users
+        for pubkey in newly_created_pubkeys {
+            if let Ok((user, _)) =
+                crate::whitenoise::users::User::find_or_create_by_pubkey(&pubkey, &self.database)
```
jgmontoya: can we move the import to the top of the file?

jgmontoya: It's annoying having to do find_or_create_by_pubkey again, even though we basically just did it within `update_follows_from_event` but couldn't find a clean way to make it better so I guess it's fine at least for now

src/whitenoise/users.rs
```diff
@@ -219,13 +234,81 @@ impl User {
...
+        &self,
+        relay_type: RelayType,
+        database: &crate::whitenoise::database::Database,
+    ) -> Result<Option<DateTime<Utc>>> {
+        let user_id = self.id.ok_or(WhitenoiseError::UserNotPersisted)?;
+        let relay_type_str = String::from(relay_type);
+
+        let result: Option<i64> = sqlx::query_scalar(
+            "SELECT MAX(event_created_at) FROM user_relays WHERE user_id = ? AND relay_type = ?",
+        )
```
jgmontoya: 👮 sql should be within database/ ‼️

src/whitenoise/event_processor/event_handlers/handle_contact_list.rs
erskingardner: it's not possible - all events will have a `created_at` field

src/whitenoise/event_processor/event_handlers/handle_contact_list.rs
erskingardner: I agree. I looked at this for a bit as well. Very unsatisfying but I'm going to leave it for now.

src/whitenoise/users.rs
erskingardner: Thank you. Will fix

src/whitenoise/database/accounts.rs
erskingardner: yeah - I'm going to leave this as is for now. the batch delete and create means that we're atomic about making sure the new follows are accurate.

jgmontoya:
> > Also, there's an edgecase for the relays:
> > If an event with created_at = 10 deletes relay X from a user, and then we get an earlier event with created_at = 5 which adds the relay X. We won't know that we already processed the first one (with higher timestamp) because we're looking at the event_created_at of remaining relays.
> 
> I'm not sure I'm following this.. if the later event is the one that removed the relay (and we removed it when we processed that one) we wouldn't want to process the earlier event, right?

We wouldn't want to process the earlier event. But we would process it if we were to receive it, because we won't have an updated `event_created_at` timestamp to compare it to (as the only change was that it removed some relay, so the row is gone)

src/nostr_manager/utils.rs
```diff
@@ -4,19 +4,17 @@ use nostr_sdk::prelude::*;
...
-            .tags
-            .iter()
-            .filter(|tag| {
-                tag.kind() == TagKind::SingleLetter(SingleLetterTag::lowercase(Alphabet::P))
-            })
-            .filter_map(|tag| tag.content().and_then(|c| c.parse::<PublicKey>().ok()))
-            .collect()
-    }
+/// Extracts public keys from an event's tags.
+pub fn pubkeys_from_event(event: &Event) -> Vec<PublicKey> {
```
jgmontoya: Should this be `pub(crate)` rather than `pub`?

src/whitenoise/database/users.rs
```diff
@@ -323,6 +474,29 @@ impl User {
...
+        .bind(user_id)
+        .bind(relay_type_str)
+        .fetch_optional(&database.pool)
+        .await
+        .map_err(|e| {
+            WhitenoiseError::Database(crate::whitenoise::database::DatabaseError::Sqlx(e))
+        })?;
+
+        Ok(crate::whitenoise::database::utils::parse_optional_scalar_timestamp(result))
+    }
```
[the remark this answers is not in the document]
erskingardner: @jgmontoya this is the thing you're referring to isn't it?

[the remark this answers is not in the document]
jgmontoya: @erskingardner yes! the first one! Using the processed_event table is a good idea actually

src/integration_tests/test_cases/metadata_management/update_metadata.rs
```diff
@@ -48,6 +48,11 @@ impl TestCase for UpdateMetadataTestCase {
     async fn run(&self, context: &mut ScenarioContext) -> Result<(), WhitenoiseError> {
         tracing::info!("Updating metadata for account: {}", self.account_name);
 
+        // Wait a moment to ensure any initial metadata events are fully processed
+        // This prevents race conditions where the initial petname metadata and
+        // test metadata events have the same timestamp
+        tokio::time::sleep(tokio::time::Duration::from_secs(1)).await;
+
```
jgmontoya: I find it odd needing this here, are you sure we need this?

db_migrations/0014_enhance_processed_events.sql
```diff
@@ -0,0 +1,23 @@
...
+-- NULL for legacy data where we don't have the author information
+ALTER TABLE processed_events ADD COLUMN author TEXT DEFAULT NULL;
+
+-- Create indexes for efficient querying
+CREATE INDEX idx_processed_events_event_kind ON processed_events(event_kind);
+CREATE INDEX idx_processed_events_event_created_at ON processed_events(event_created_at);
+CREATE INDEX idx_processed_events_kind_timestamp ON processed_events(event_kind, event_created_at);
+CREATE INDEX idx_processed_events_author ON processed_events(author);
+CREATE INDEX idx_processed_events_author_kind ON processed_events(author, event_kind);
+CREATE INDEX idx_processed_events_author_kind_timestamp ON processed_events(author, event_kind, event_created_at);
```
jgmontoya: Just to confirm: all of these indexes are due to specific queries we either have implemented or have planned to implement?

src/nostr_manager/query.rs
```diff
@@ -43,7 +47,7 @@ impl NostrManager {
         pubkey: PublicKey,
         relay_type: RelayType,
         nip65_relays: &[Relay], // TODO: Replace with &[RelayUrl]
-    ) -> Result<HashSet<RelayUrl>> {
+    ) -> Result<Option<(HashSet<RelayUrl>, Timestamp, EventId)>> {
```
jgmontoya: Is it preferable to have this possibly return None rather than just an empty set?

src/nostr_manager/query.rs
```diff
@@ -88,11 +95,39 @@ impl NostrManager {
             .max_by_key(|e| (e.created_at, e.id));
         Ok(latest)
     }
+
+    pub(crate) async fn fetch_contact_list_events(
+        &self,
+        pubkey: PublicKey,
+        relays: &[RelayUrl],
+    ) -> Result<Option<Event>> {
+        let filter = Filter::new().author(pubkey).kind(Kind::ContactList);
```
jgmontoya, line 61: I think we could use `.until` on the filter as well maybe?

src/whitenoise/database/accounts.rs
```diff
@@ -310,6 +311,83 @@ impl Account {
...
+                // Track newly created users so the caller can fetch their metadata
+                newly_created_users.push(pubkey);
+                tracing::debug!("Created new user for follow: {}", pubkey.to_hex());
+            }
+
+            sqlx::query("INSERT INTO account_follows (account_id, user_id, created_at, updated_at) VALUES (?, ?, ?, ?)")
+                .bind(self.id)
+                .bind(user.id)
+                .bind(Utc::now().timestamp_millis())
+                .bind(Utc::now().timestamp_millis())
```
jgmontoya: we should do the now() before and here just assign so that we get the exact same value on both created_at and updated_at

src/whitenoise/database/accounts.rs
```diff
@@ -1236,4 +1323,403 @@ mod tests {
...
+        .unwrap();
+
+        let account = Account {
+            id: Some(1),
+            pubkey: test_pubkey,
+            user_id: test_user_id,
+            last_synced_at: None,
+            created_at: chrono::DateTime::from_timestamp_millis(test_timestamp).unwrap(),
+            updated_at: chrono::DateTime::from_timestamp_millis(test_timestamp).unwrap(),
+        };
```
jgmontoya: Isn't it simpler to call account.save rather than have this raw insert?

src/whitenoise/database/accounts.rs
```diff
@@ -1236,4 +1323,403 @@ mod tests {
...
+        let newly_created = result.unwrap();
+        assert!(newly_created.is_empty());
+
+        // Verify no follows exist
+        let count: i64 =
+            sqlx::query_scalar("SELECT COUNT(*) FROM account_follows WHERE account_id = ?")
+                .bind(account.id.unwrap())
+                .fetch_one(&pool)
+                .await
+                .unwrap();
```
jgmontoya: account.follows, no?

src/whitenoise/database/processed_events.rs
```diff
@@ -29,11 +33,37 @@ where
 
         let created_at = parse_timestamp(row, "created_at")?;
 
+        // Handle nullable event_created_at
+        let event_created_at = match row.try_get::<Option<i64>, &str>("event_created_at")? {
+            Some(timestamp) => {
+                Some(DateTime::from_timestamp_millis(timestamp).ok_or_else(|| {
+                    sqlx::Error::Decode("Invalid event_created_at timestamp".into())
+                })?)
```
jgmontoya: I think we should use our parse_timestamp util

src/whitenoise/database/processed_events.rs
```diff
@@ -57,6 +98,7 @@ impl ProcessedEvent {
 
     /// Checks if we already processed a specific event
     /// - account_id: Some(id) for account-specific processing, None for global processing
+    /// - author: Optional event author for enhanced duplicate detection
```
jgmontoya: This comment line is wrong

src/whitenoise/database/users.rs
```diff
@@ -239,6 +242,122 @@ impl User {
...
+                    created_at: Utc::now(),
+                    updated_at: Utc::now(),
+                };
+                user = user.save_tx(tx).await?;
+                Ok((user, true))
+            }
+            Err(e) => Err(e),
+        }
+    }
+
```
jgmontoya: I don't like these repeated logic. I think we can just use the normal methods if we replace the type (currently `&Database`) with the proper thing (I think `Executor` from sqlx, not 100% sure maybe we need to specify some traits)

src/whitenoise/event_processor/event_handlers/handle_metadata.rs
```diff
@@ -14,23 +16,51 @@ impl Whitenoise {
...
+                    DateTime::from_timestamp_millis((event.created_at.as_u64() * 1000) as i64)
+                        .ok_or_else(|| {
+                            WhitenoiseError::EventProcessor(format!(
+                                "Invalid timestamp in metadata event: {}",
+                                event.created_at.as_u64()
+                            ))
+                        })?;
+
+                let should_update = newly_created
+                    || user.metadata == Metadata::default()
```
jgmontoya: Why do we need this case?

src/whitenoise/database/processed_events.rs
```diff
@@ -29,11 +33,37 @@ where
...
+                    sqlx::Error::Decode("Invalid event_created_at timestamp".into())
+                })?)
+            }
+            None => None,
+        };
+
+        // Handle nullable event_kind
+        let event_kind = row
+            .try_get::<Option<i64>, &str>("event_kind")?
+            .map(|kind| kind as u16);
```
jgmontoya: It'd be cool if we could deal with rust-nostr `Kind` enum objects outside of this. would that be too much of a pain in the ass to handle?

src/whitenoise/event_processor/account_event_processor.rs
```diff
@@ -164,22 +174,39 @@ impl Whitenoise {
         event: &Event,
         account: &Account,
     ) -> Result<Option<&'static str>> {
-        // Check if we already processed this event
+        // Check if we already processed this event (with author consideration)
         let already_processed = match self
             .nostr
             .event_tracker
-            .already_processed_account_event(&event.id, &account.pubkey)
+            .already_processed_account_event_with_author(&event.id, &account.pubkey, &event.pubkey)
```
jgmontoya: I don't understand the need of this "author consideration". Why is it needed?

src/whitenoise/event_processor/global_event_processor.rs
```diff
@@ -96,18 +101,35 @@ impl Whitenoise {
...
-            .already_processed_global_event(&event.id)
+            .already_processed_global_event_with_author(&event.id, &event.pubkey)
             .await
         {
             Ok(v) => v,
             Err(e) => {
                 tracing::error!(
                     target: "whitenoise::event_processor::should_skip_global_event_processing",
-                    "Already processed check failed for {}: {}",
+                    "Already processed check with author failed for {}: {}",
```
jgmontoya: Same here, why do we need the `with_author` thing?

src/whitenoise/event_processor/global_event_processor.rs
```diff
@@ -119,22 +141,26 @@ impl Whitenoise {
...
-                    target: "whitenoise::event_processor::should_skip_global_event_processing",
-                    "Global published check failed for {}: {}",
-                    event.id.to_hex(),
-                    e
-                );
-                false
-            }
+        // We don't skip metadata events because we need them to update local database
+        let should_skip = match event.kind {
+            Kind::Metadata => false, // Always process metadata events to update local database
```
jgmontoya: This feels wrong, or the start of something that could get way too convoluted 🤔

src/whitenoise/accounts.rs
```diff
@@ -857,10 +849,77 @@ impl Whitenoise {
...
+                    account_clone.pubkey.to_hex(),
+                    e
+                );
+            }
+
+            // Fetch the account's contact list
+            let fetched_contact_list = whitenoise
+                .nostr
+                .fetch_contact_list_events(account_clone.pubkey, &user_relays)
+                .await;
```
jgmontoya, line 896: I think this logic should be within the `fetch_user_event_streams` method (or within a similar, dedicated one) in NostrManager. It should receive the vector of relay urls.

This would:
a) Make sure we have a signer set
b) Run in parallel with other relay event fetching

src/nostr_manager/mod.rs
```diff
@@ -473,23 +470,15 @@ impl NostrManager {
         &self,
         signer: impl NostrSigner + 'static,
         pubkey: PublicKey,
+        contact_list_pubkeys: Vec<PublicKey>,
```
jgmontoya, line 475: Currently I know we're doing this for "contacts and self" however I think that's a mistake and we have to do it for all users

src/whitenoise/accounts.rs
```diff
@@ -857,10 +849,77 @@ impl Whitenoise {
...
+                    account_clone.pubkey.to_hex()
+                );
+                if let Err(e) = whitenoise
+                    .handle_contact_list(&account_clone, contact_list_event)
+                    .await
+                {
+                    tracing::error!(target: "whitenoise::background_fetch_account_data", "Failed to process contact list for account {}: {}", account_clone.pubkey.to_hex(), e);
+                }
+            }
+
```
jgmontoya, line 924: I think we can move this to the `process_user_event_streams`, I'm not sure why it needs to be processed before fetching the other events

src/whitenoise/users.rs
```diff
@@ -35,14 +36,81 @@ impl User {
...
+                                target: "whitenoise::users::sync_metadata",
+                                "Ignoring stale metadata event for user {} (event: {}, stored: {})",
+                                self.pubkey,
+                                event_datetime.timestamp_millis(),
+                                stored_timestamp.timestamp_millis()
+                            );
+                        }
+                        is_newer_or_equal
+                    }
+                };
```
jgmontoya: Can we maybe move all of this to a private `should_update_metadata` or similar private method?

src/whitenoise/users.rs
```diff
@@ -35,14 +36,81 @@ impl User {
...
+                            target: "whitenoise::users::sync_metadata",
+                            "No processed metadata events for user {}, accepting new event",
+                            self.pubkey
+                        );
+                        true
+                    }
+                    Some(stored_timestamp) => {
+                        let is_newer_or_equal = event_datetime.timestamp_millis()
+                            >= stored_timestamp.timestamp_millis();
+                        if !is_newer_or_equal {
```
jgmontoya: when timestamps are equal we should disambiguate with the event id, no?

src/whitenoise/users.rs
```diff
@@ -316,8 +427,59 @@ impl User {
...
+                let event_created_at = Some(
+                    DateTime::from_timestamp_millis((event_timestamp.as_u64() * 1000) as i64)
+                        .ok_or_else(|| {
+                            WhitenoiseError::Database(
+                                crate::whitenoise::database::DatabaseError::InvalidTimestamp {
+                                    timestamp: (event_timestamp.as_u64() * 1000) as i64,
+                                },
+                            )
+                        })?,
+                );
```
jgmontoya: Maybe we should make it so the methods in nostr_manager already return the timestamp in milliseconds?

src/whitenoise/users.rs
erskingardner: I don't think that we need to. I've moved the conditional logic about whether we should attempt an update into a helper and now, if we have two metadata events with the same timestamp but we don't have a processed_event record for that specific event, we should definitely process it.

src/nostr_manager/mod.rs
erskingardner: Hmm, right now in the `setup_accounts_sync_and_subscriptions` method, we're looping through the accounts and triggering the background refresh for both/all. Sounds like we might need to break out the background sync stuff that is common to all accounts (like syncing user data) and stuff that is specific to each account individually. What do you think? 

Then on startup we have: 
1. Setup global subs
1. Background sync global stuff
2. Loop through each account and...
    1. Background sync account specific stuff
    2. Set up subscriptions for account specific stuff.

src/whitenoise/database/users.rs
erskingardner: I tried this out but the problem is the `sqlx::Executor` trait is consuming - when you pass an executor to a query, it takes ownership of it. This means you can only use it once per method. So we could have a mixed bag of methods where we only have query using the generic or we can have more methods but WAY more clarity as to what you pass and why. I like the clarity.

src/whitenoise/event_processor/event_handlers/handle_metadata.rs
erskingardner: Because users without metadata come back with Metadata::default() which is just a Metadata with all fields blank. This is sort of a null check to quickly jump ahead to process it if the user doesn't have anything.

src/integration_tests/test_cases/metadata_management/update_metadata.rs
erskingardner: It's now passing without 🤷‍♂️ I'll remove.

src/nostr_manager/query.rs
erskingardner: it does return none. If `latest` comes back as None, we bail and return None for the whole function

src/nostr_manager/query.rs
erskingardner: for what reason?

src/whitenoise/database/processed_events.rs
erskingardner: Great catch. fixed.

src/nostr_manager/query.rs
jgmontoya: I know, but it used to return an empty set, that's why I'm asking if it's something we want in general

src/nostr_manager/query.rs
jgmontoya: Just so we might get fewer results to filter

src/whitenoise/event_processor/account_event_processor.rs
erskingardner: Removed all these methods

src/whitenoise/database/processed_events.rs
```diff
@@ -29,11 +33,37 @@ where
 
         let created_at = parse_timestamp(row, "created_at")?;
 
+        // Handle nullable event_created_at
+        let event_created_at = match row.try_get::<Option<i64>, &str>("event_created_at")? {
+            Some(timestamp) => {
+                Some(DateTime::from_timestamp_millis(timestamp).ok_or_else(|| {
+                    sqlx::Error::Decode("Invalid event_created_at timestamp".into())
+                })?)
+            }
```
jgmontoya: I think this should use our `parse_timestamp` util

src/whitenoise/database/processed_events.rs
```diff
@@ -83,6 +124,79 @@ impl ProcessedEvent {
...
+        // Build query based on parameters
+        let query = match (account_id.is_some(), author_pubkey.is_some()) {
+            (true, _) => {
+                // Account-specific events (author filter not applicable)
+                "SELECT MAX(event_created_at) FROM processed_events WHERE account_id = ? AND event_kind = ?"
+            }
+            (false, true) => {
+                // Global events with author filter
+                "SELECT MAX(event_created_at) FROM processed_events WHERE account_id IS NULL AND event_kind = ? AND author = ?"
+            }
```
jgmontoya, line 148: There's gonna be some weirdness around author = NULL entries in our db, which will be the old entries that are already present for which we don't have the author.

Maybe we should just wipe the processed_events data entirely and make the field non-optional?

src/whitenoise/database/processed_events.rs
```diff
@@ -43,12 +73,23 @@ impl ProcessedEvent {
     pub(crate) async fn create(
         event_id: &EventId,
         account_id: Option<i64>,
+        event_created_at: Option<DateTime<Utc>>,
+        event_kind: Option<u16>,
```
jgmontoya: Should we use Kind instead of raw u16?

src/whitenoise/accounts.rs
```diff
@@ -857,10 +849,77 @@ impl Whitenoise {
...
+                .await;
+
+            let contact_list_pubkeys = match &fetched_contact_list {
+                Ok(Some(contact_list_event)) => pubkeys_from_event(contact_list_event),
+                Ok(None) => Vec::new(),
+                Err(e) => {
+                    tracing::error!(target: "whitenoise::background_fetch_account_data", "Failed to fetch contact list for account {}: {}", account_clone.pubkey.to_hex(), e);
+                    Vec::new()
+                }
+            };
```
jgmontoya: I think there are some issues:

1. This will probably need to use the nostr signer lock, so it should probably be in a method within nostr_manager and using the `with_signer` helper

2. We're now processing the relays for the account before calling `fetch_user_event_streams` which can I can foresee being a bit problematic. For example:
  a. User receives some invite (giftwrap) on some relay X
  b. User then changes to relay Y
  c. We first switch to relay Y, then we miss the invite because we're not looking for it on relay X

The easiest fix I can think of right now would be to:
1. Fetch all the events / streams with current relays
2. Update the relays / process the events
3. Loop if there's any change on the relays (`GOTO 1`)

src/whitenoise/event_tracker.rs
```diff
@@ -174,15 +183,24 @@ impl EventTracker for WhitenoiseEventTracker {
         &self,
         event_id: &EventId,
         pubkey: &PublicKey,
+        event_created_at: Option<DateTime<Utc>>,
+        event_kind: Option<u16>,
```
jgmontoya: Maybe it'd be more natural to just receive the event in this method and internally extract all we need?

src/whitenoise/event_tracker.rs
```diff
@@ -203,11 +221,20 @@ impl EventTracker for WhitenoiseEventTracker {
     async fn track_processed_global_event(
         &self,
         event_id: &EventId,
+        event_created_at: Option<DateTime<Utc>>,
+        event_kind: Option<u16>,
```
jgmontoya: Same here, I think we could just pass the entire event

src/whitenoise/event_tracker.rs
```diff
@@ -276,14 +303,23 @@ impl EventTracker for TestEventTracker {
         &self,
         event_id: &EventId,
         pubkey: &PublicKey,
+        event_created_at: Option<DateTime<Utc>>,
+        event_kind: Option<u16>,
```
jgmontoya: and here

src/whitenoise/event_tracker.rs
```diff
@@ -303,10 +339,19 @@ impl EventTracker for TestEventTracker {
     async fn track_processed_global_event(
         &self,
         event_id: &EventId,
+        event_created_at: Option<DateTime<Utc>>,
+        event_kind: Option<u16>,
```
jgmontoya: and here and in all of these methods, no?

src/whitenoise/users.rs
```diff
@@ -334,6 +446,106 @@ impl User {
...
+
+        // 3. If user has default metadata, always accept the event
+        if self.metadata == Metadata::default() {
+            tracing::debug!(
+                target: "whitenoise::users::should_update_metadata",
+                "Accepting metadata event for user {} with default metadata",
+                self.pubkey
+            );
+            return Ok(true);
+        }
```
jgmontoya: I still don't understand why this needs to be a special case and why we wouldn't handle it with the rest of the logic

src/whitenoise/users.rs
```diff
@@ -334,6 +446,106 @@ impl User {
...
+    /// # Arguments
+    /// * `event_id` - The ID of the metadata event being considered
+    /// * `event_datetime` - The datetime of the metadata event being considered
+    /// * `newly_created` - Whether this user was just created
+    /// * `database` - Database connection for checking processed events
+    ///
+    /// # Returns
+    /// * `Ok(true)` if metadata should be updated
+    /// * `Ok(false)` if the event should be ignored
+    pub(crate) async fn should_update_metadata(
```
jgmontoya, line 464: Before, we were also checking whether the metadata actually changed, now we don't. I don't think it makes much of a difference other than it can potentially save us a couple of queries

src/whitenoise/database/users.rs
```diff
@@ -213,15 +213,18 @@ impl User {
...
 
-        // Use INSERT ON CONFLICT to handle both insert and update cases without deleting/replacing rows
+        // Use INSERT ON CONFLICT to handle both insert and update cases
         sqlx::query(
-            "INSERT INTO users (pubkey, metadata, created_at, updated_at) VALUES (?, ?, ?, ?) ON CONFLICT(pubkey) DO UPDATE SET metadata = excluded.metadata, updated_at = ?",
+            "INSERT INTO users (pubkey, metadata, created_at, updated_at)
+             VALUES (?, ?, ?, ?)
+             ON CONFLICT(pubkey) DO UPDATE SET
+               metadata = excluded.metadata,
+               updated_at = excluded.updated_at",
```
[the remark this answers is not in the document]
jgmontoya: @erskingardner what do you think of this one?

src/whitenoise/users.rs
erskingardner: Removed

src/whitenoise/database/processed_events.rs
erskingardner: It ends up being quite confusing when you write it with the util helper (not the database helper) because of the error conversions and different type params. I'm going to leave this.

src/whitenoise/database/users.rs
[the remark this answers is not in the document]
erskingardner: Yeah - we should be using the actual time for the updated_at fields.

src/whitenoise/database/processed_events.rs
erskingardner: Yeah, that's right. I can't really figure out a way to make this query handle that either... 

What about dropping only all the rows that don't have an account_id value? That would mean that we'd keep the account-specific stuff (where we don't need to filter by author)?

src/whitenoise/accounts.rs
erskingardner: I think we should handle this in a later removal of the background_sync.

src/whitenoise/database/processed_events.rs
erskingardner: Oh wait, actually, this query is the same as deleting all the ProcessedEvents that don't have an author field. 

`"SELECT MAX(event_created_at) FROM processed_events WHERE account_id IS NULL AND event_kind = ? AND author = ?"`

The author filter there will exclude events with null authors. So there isn't any benefit from deleting things.

src/whitenoise/users.rs
```diff
@@ -334,6 +443,99 @@ impl User {
...
+                    "No processed metadata events for user {}, accepting new event",
+                    self.pubkey
+                );
+                true
+            }
+            Some(stored_timestamp) => {
+                let event_datetime = timestamp_to_datetime(event.created_at)?;
+                let is_newer_or_equal =
+                    event_datetime.timestamp_millis() >= stored_timestamp.timestamp_millis();
+                if !is_newer_or_equal {
```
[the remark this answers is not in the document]
erskingardner: This is fine. Stop suggesting it

PR #366, opened 2025-09-15
Edge case fix for old data migration

erskingardner:
This changes an older migration so it's only possible to merge this before our next release. 

In the case that an Account didn't have a corresponding contact record in the system before the migration, that account would not get created in the new table because of a pesky `JOIN` clause in the data migration. This creates a user record before creating the new account to fix this.

Potentially part of the solution to: https://github.com/parres-hq/whitenoise_flutter/issues/557

## Summary by CodeRabbit

* **Bug Fixes**
  * Backfills missing user records for accounts and prevents duplicate relay and user-relay entries to improve data consistency.

* **Chores**
  * Makes migration idempotent across relay and relationship inserts and validates derived user metadata to avoid invalid values.

* **Documentation**
  * Adds a note clarifying why a later join is safe due to the earlier backfill and the added safety guard.

### DIFF
--- a/db_migrations/0006_data_migration.sql
+++ b/db_migrations/0006_data_migration.sql
@@ -15,6 +15,15 @@ SELECT
 FROM contacts
 WHERE NOT EXISTS (SELECT 1 FROM users WHERE users.pubkey = contacts.pubkey);
 
+-- Step 1.5: FIXES EDGE CASE - Create users for accounts that don't have corresponding contacts
+-- This ensures every account will have a user record before Step 6
+INSERT INTO users (pubkey, metadata)
+SELECT
+    pubkey,
+    NULL as metadata  -- No metadata available for accounts without contacts
+FROM accounts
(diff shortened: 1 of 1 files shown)


db_migrations/0006_data_migration.sql
```diff
@@ -13,7 +13,27 @@ SELECT
...
+  AND TRIM(pubkey) != '';
+
+-- Step 1.5: FIXES EDGE CASE - Create users for accounts that don't have corresponding contacts
+-- This ensures every account will have a user record before Step 6
+INSERT INTO users (pubkey, metadata)
+SELECT
+    pubkey,
+    NULL as metadata  -- No metadata available for accounts without contacts
+FROM accounts
+WHERE NOT EXISTS (SELECT 1 FROM users WHERE users.pubkey = accounts.pubkey);
```
jgmontoya: Why do we need this one? Looks redundant

erskingardner: It is. accidental.

PR #368, opened 2025-09-18
feat(justfile): add int-test-flamegraph to generate a flamegraph profile using the integration tests on release mode

jgmontoya:
Also add production relays to integration tests for release mode profiling

flamegraph


src/integration_tests/core/context.rs
```diff
@@ -15,9 +15,18 @@ pub struct ScenarioContext {
 
 impl ScenarioContext {
     pub fn new(whitenoise: &'static Whitenoise) -> Self {
+        let relays = if cfg!(debug_assertions) {
+            vec!["ws://localhost:8080", "ws://localhost:7777"]
+        } else {
+            vec![
```
erskingardner, line 21: I have a feeling we're going to get seriously rate limited on these relays with the tests. 

I wonder if it makes more sense for us to set up a few testing relays in a few different regions on AWS or something so we can test it?

src/integration_tests/test_cases/user_discovery/find_or_create_user.rs
```diff
@@ -36,10 +36,18 @@ impl FindOrCreateUserTestCase {
     }
 
     pub fn with_relays(mut self) -> Self {
-        let test_relays = vec![
-            RelayUrl::parse("ws://localhost:8080").unwrap(),
-            RelayUrl::parse("ws://localhost:7777").unwrap(),
-        ];
+        let test_relays = if cfg!(debug_assertions) {
+            vec![
```
erskingardner, line 40: can we centralize the relays definition somewhere?

src/integration_tests/core/context.rs
jgmontoya: > I have a feeling we're going to get seriously rate limited on these relays with the tests.

Yes we do get rate limited when running this tests, but we don't really need to do it all that often so while we could set up some relays to test I don't think we really need to, at least for now.

src/integration_tests/test_cases/user_discovery/find_or_create_user.rs
jgmontoya: I'd leave them for now, these are specific to the test (they just happen to be the same as we're using elsewhere, but we could as well be using others)

PR #373, opened 2025-09-25
Refactor/remove background sync

jgmontoya:
## Summary by CodeRabbit

- New Features
  - Account-scoped follow-list publishing and verification; per-account and global optional "since" timestamps for subscription resumption.
  - Utilities to validate/cap event timestamps and adjust lookback for giftwrap events.
  - New test case to verify last-synced timestamp behavior for follow vs global metadata events.

- Bug Fixes
  - Prevents last-synced timestamp regression; skips events with invalid future timestamps.

- Refactor
  - Subscription setup reorganized to compute global since and run per-account subscription setup.

- Tests
  - Expanded integration and unit tests for follow-list flows, timestamp handling, and giftwrap lookback.

### DIFF
--- a/src/whitenoise/event_processor/account_event_processor.rs
+++ b/src/whitenoise/event_processor/account_event_processor.rs
@@ -2,6 +2,7 @@ use nostr_sdk::prelude::*;
 use sha2::{Digest, Sha256};
 
 use crate::{
+    nostr_manager::utils::cap_timestamp_to_now,
     types::RetryInfo,
     whitenoise::{
         accounts::Account,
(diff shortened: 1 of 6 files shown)


src/whitenoise/event_processor/account_event_processor.rs
```diff
@@ -85,6 +85,36 @@ impl Whitenoise {
                         e
                     );
                 }
+
+                // Advance account sync timestamp only for account-scoped events
+                match event.kind {
+                    Kind::ContactList | Kind::GiftWrap | Kind::MlsGroupMessage => {
+                        let created_ms = (event.created_at.as_u64() as i64) * 1000;
```
erskingardner: what happens if we receive an event that has a far future timestamp. I don't know if we automatically reject those already but I can see bad events causing problems here.

PR #383, opened 2025-10-07
Media files support

erskingardner:
## Summary by CodeRabbit

- **New Features**
  - Local media cache with on‑demand and background syncing for group images; public media orchestrator and storage APIs for content‑addressed storage, deduplication, and retrieval.

- **Improvements**
  - Automatic image type detection/validation and renaming to JPEG variant; richer conversions and extensions.
  - Immediate caching after uploads, background sync triggers on relevant events, and prefix-based cached-file lookup.
  - More descriptive error variants for media, downloads, decryption, and hash mismatches.

- **Database**
  - Migration recreates normalized media_files table with indexes and unique constraint.

- **Documentation**
  - Media storage architecture guide added.

- **Tests**
  - Expanded tests for storage, caching, detection, and DB interactions.

- **Chores**
  - Dependency manifest formatting and declaration updates; dev dependency relocated.

### DIFF
--- a/src/whitenoise/groups.rs
+++ b/src/whitenoise/groups.rs
@@ -721,6 +721,12 @@ impl Whitenoise {
     }
 
     /// Links an already cached image to a group in the database
+    ///
+    /// When an image file is already cached on disk (from another group using the same image),
+    /// this method creates a database record linking this group/account to the existing file.
+    ///
+    /// If an existing database record exists for this file hash, it copies metadata from it.
+    /// If no database record exists, it detects the MIME type from the cached file itself.
     async fn link_cached_image_to_group(
         &self,
         account_pubkey: &PublicKey,
@@ -728,21 +734,133 @@ impl Whitenoise {
         cached_path: &Path,
         image_hash: &[u8; 32],
     ) {
+        // Try to get metadata from any existing record with this hash
+        let existing_record_opt =
+            match crate::whitenoise::database::media_files::MediaFile::find_by_hash(
+                &self.database,
+                image_hash,
+            )
+            .await
+            {
+                Ok(record_opt) => record_opt,
+                Err(e) => {
+                    tracing::warn!(
+                        target: "whitenoise::groups::link_cached_image_to_group",
+                        "Failed to query database for existing record: {}",
+                        e
+                    );
+                    return;
+                }
+            };
+
+        if let Some(existing_record) = existing_record_opt {
+            self.link_cached_image_from_existing_record(
+                account_pubkey,
+                group_id,
+                cached_path,
+                image_hash,
+                existing_record,
+            )
+            .await;
+        } else {
+            self.link_cached_image_with_detection(
+                account_pubkey,
+                group_id,
+                cached_path,
+                image_hash,
+            )
+            .await;
+        }
+    }
+
+    /// Links a cached image using metadata from an existing database record
+    async fn link_cached_image_from_existing_record(
+        &self,
+        account_pubkey: &PublicKey,
+        group_id: &GroupId,
+        cached_path: &Path,
+        image_hash: &[u8; 32],
+        existing_record: crate::whitenoise::database::media_files::MediaFile,
+    ) {
+        let metadata_ref = existing_record.file_metadata.as_ref();
         let upload = MediaFileUpload {
-            data: &[],
+            data: &[], // Empty is OK - file already exists on disk
             file_hash: *image_hash,
-            mime_type: "", // MIME type not needed for update
-            media_type: "group_image",
-            blossom_url: None,
-            file_metadata: None,
+            mime_type: &existing_record.mime_type,
+            media_type: &existing_record.media_type,
+            blossom_url: existing_record.blossom_url.as_deref(),
+            file_metadata: metadata_ref,
         };
+
         if let Err(e) = self
             .media_files()
             .record_in_database(account_pubkey, group_id, cached_path, upload)
             .await
         {
             tracing::warn!(
-                target: "whitenoise::groups::link_cached_image_to_group",
+                target: "whitenoise::groups::link_cached_image_from_existing_record",
+                "Failed to record file in database: {}",
+                e
+            );
+        }
+    }
+
+    /// Links a cached image by detecting its MIME type from the file
+    async fn link_cached_image_with_detection(
+        &self,
+        account_pubkey: &PublicKey,
+        group_id: &GroupId,
+        cached_path: &Path,
+        image_hash: &[u8; 32],
+    ) {
+        tracing::debug!(
+            target: "whitenoise::groups::link_cached_image_with_detection",
+            "No existing database record for hash {}, detecting MIME type from cached file",
+            hex::encode(image_hash)
+        );
+
+        let file_data = match tokio::fs::read(cached_path).await {
+            Ok(data) => data,
+            Err(e) => {
+                tracing::warn!(
+                    target: "whitenoise::groups::link_cached_image_with_detection",
+                    "Failed to read cached file {}: {}",
+                    cached_path.display(),
+                    e
+                );
+                return;
+            }
+        };
+
+        let image_type = match ImageType::detect(&file_data) {
+            Ok(img_type) => img_type,
+            Err(e) => {
+                tracing::warn!(
+                    target: "whitenoise::groups::link_cached_image_with_detection",
+                    "Failed to detect image type for cached file {}: {}",
+                    cached_path.display(),
+                    e
+                );
+                return;
+            }
+        };
+
+        let upload = MediaFileUpload {
+            data: &[], // Empty is OK - file already exists on disk
+            file_hash: *image_hash,
+            mime_type: image_type.mime_type(),
+            media_type: "group_image",
+            blossom_url: None,
+            file_metadata: None,
+        };
+
+        if let Err(e) = self
+            .media_files()
+            .record_in_database(account_pubkey, group_id, cached_path, upload)
+            .await
+        {
+            tracing::warn!(
+                target: "whitenoise::groups::link_cached_image_with_detection",
                 "Failed to record file in database: {}",
                 e
             );


db_migrations/0015_recreate_media_files.sql
```diff
@@ -0,0 +1,44 @@
...
+    mime_type TEXT NOT NULL,          -- Canonical MIME type (e.g., "image/jpeg")
+    media_type TEXT NOT NULL,         -- "group_image" or "chat_media"
+    blossom_url TEXT,                 -- Optional: NULL for default server, set for custom
+    nostr_key TEXT,                   -- Optional: For future MIP-04 cleanup
+    dimensions TEXT,                  -- Optional: Image/video dimensions as "widthxheight" (e.g., "1920x1080")
+    blurhash TEXT,                    -- Optional: BlurHash string for progressive loading
+    created_at INTEGER NOT NULL,
+    accessed_at INTEGER NOT NULL,     -- For LRU eviction
+    FOREIGN KEY (account_pubkey) REFERENCES accounts(pubkey) ON DELETE CASCADE
+);
```
jgmontoya, line 30: Previously we had a `file_metadata JSONB` field. One of it purposes (if I recall correctly) was to store the original filename. I think we should still have that.

Things like dimensions and blurhash I believe should be in there as well. The reason being that those are metadata 'attributes' that only make sense on certain types of files, namely images, so I think it kinda maps better to an unstructured json field.

db_migrations/0015_recreate_media_files.sql
```diff
@@ -0,0 +1,44 @@
...
+    file_path TEXT NOT NULL,
+    file_hash TEXT NOT NULL,          -- Hex-encoded hash of ENCRYPTED data (image_hash for group images, per MIP-01)
+    mime_type TEXT NOT NULL,          -- Canonical MIME type (e.g., "image/jpeg")
+    media_type TEXT NOT NULL,         -- "group_image" or "chat_media"
+    blossom_url TEXT,                 -- Optional: NULL for default server, set for custom
+    nostr_key TEXT,                   -- Optional: For future MIP-04 cleanup
+    dimensions TEXT,                  -- Optional: Image/video dimensions as "widthxheight" (e.g., "1920x1080")
+    blurhash TEXT,                    -- Optional: BlurHash string for progressive loading
+    created_at INTEGER NOT NULL,
+    accessed_at INTEGER NOT NULL,     -- For LRU eviction
```
jgmontoya: This seems a bit premature but ok, it doesn't hurt too much

docs/storage-architecture.md
```diff
@@ -0,0 +1,189 @@
...
+**Key Types:**
+- `MediaFile` - Database record representation
+- `MediaFileParams` - Parameters for saving records
+- `MediaFileRow` - Internal database row type
+
+### 3. `storage/media_files.rs` (Middle Branch)
+**Purpose:** Filesystem operations for media files
+
+**Responsibilities:**
+- Create directory structures (`<cache_dir>/<group_id_hex>/<subdirectory>/`)
```
jgmontoya: I'm not sure I would scope it inside a group (group_id). If a user receives an image/media from group X, then forwards it to groups Y and Z there's no need to copy/store/cache the actual image multiple times. We'll just have the distinct database records pointing to the same local file, I think that's fine.

docs/storage-architecture.md
```diff
@@ -0,0 +1,189 @@
...
+- Coordinate between storage layer and database layer
+- Provide convenience methods for common operations
+- Combine filesystem and database operations atomically
+- Business logic and validation
+
+**Does NOT:**
+- Touch the filesystem directly (delegates to storage layer)
+- Execute SQL directly (delegates to database layer)
+- Handle network operations (use BlossomClient)
+- Handle encryption/decryption (MDK's responsibility)
```
jgmontoya: Maybe clarify "Handle encryption/decryption **directly**", as the above diagram says:

```
┌─────────────────────────────────────────────────────────────┐
│                    whitenoise/media_files                   │
│                   (High-level orchestration)                │
│  - Coordinates between storage, database, and Blossom       │
│  - Handles encryption/decryption                            │
│  - Manages business logic                                   │
└────────────┬──────────────┬──────────────┬──────────────────┘
```

so it seems a bit contradictory or at least it can be confusing

docs/storage-architecture.md
```diff
@@ -0,0 +1,189 @@
...
+- `Storage` - Wrapper struct for all storage subsystems
+
+**Directory Structure:**
+```
+<data_dir>/media_cache/
+  <mls_group_id_hex>/
+    group_images/          # Group profile images (marmot extension)
+      <hash>.<ext>
+    media/                 # Chat media files
+      <hash>.<ext>
```
jgmontoya: same comment here regarding scoping the files on the fs. I think we can rely on the hash to deduplicate without needing to rely on directory structure

docs/storage-architecture.md
```diff
@@ -0,0 +1,189 @@
...
+        file_path: &path,
+        file_hash: &hash,
+        mime_type: "image/jpeg",
+        media_type: "group_image",
+        blossom_url: Some("https://..."),
+        dimensions: None,
+        blurhash: None,
+    },
+).await?;
+```
```
jgmontoya, line 196: I'd either remove this or explicitly say this is intended to be used only on `whitenoise/media_files.rs`. I can foresee there being issues with db/storage syncing if we're not careful about how we're handling things

docs/storage-architecture.md
```diff
@@ -0,0 +1,189 @@
...
+
+### Chat Media Files
+
+Regular chat media (images, videos, etc.) would be stored in the `media` subdirectory.
```
jgmontoya: I'm not sure I see a reason to have a separate (sub)directory

db_migrations/0015_recreate_media_files.sql
jgmontoya: Actually thinking about it a bit more, I think we should remove this. We might want to implement this in a different way with a different table maybe in the future. I just don't think we need to really think about this too much and it's gonna complicate things to keep it 'in sync' (update in the db when we grab the file)

src/whitenoise/event_processor/event_handlers/handle_giftwrap.rs
```diff
@@ -46,12 +46,27 @@ impl Whitenoise {
...
             tracing::debug!(target: "whitenoise::event_processor::process_welcome", "Processed welcome event");
-        } // mdk lock released here
+            welcome.mls_group_id
+        }; // mdk lock released here
+
+        // After processing welcome, proactively cache the group image if it has one
+        // This ensures the image is ready when the UI displays the group
+        if let Err(e) = self
+            .sync_group_image_cache_if_needed(account, &group_id)
+            .await
```
jgmontoya: I think we should do this in a background task to not block the event processor

src/whitenoise/event_processor/event_handlers/handle_mls_message.rs
```diff
@@ -22,6 +23,17 @@ impl Whitenoise {
...
                   result
                 );
+
+                if let MessageProcessingResult::Commit = result {
+                    let groups_result = self.groups(account, true).await;
+                    if let Ok(groups) = groups_result {
+                        // TODO: This is a blunt instrument, in future we can improve this by looking up the specific group the Commit was against.
+                        for group in groups {
+                            self.sync_group_image_cache_if_needed(account, &group.mls_group_id)
+                                .await?;
```
jgmontoya: Here as well I'd do it as a background task to not block the event processor

src/whitenoise/storage/media_files.rs
```diff
@@ -0,0 +1,314 @@
...
+use tokio::fs;
+
+/// Filesystem storage for media files organized by MLS group
+///
+/// Directory structure:
+/// ```
+/// <cache_dir>/
+///   <mls_group_id_hex>/
+///     <subdirectory>/
+///       <filename>
```
jgmontoya: Just a reminder to update here as well if you take my comment regarding the scoping of files by group

src/whitenoise/groups.rs
```diff
@@ -502,72 +498,247 @@ impl Whitenoise {
...
+    /// # Arguments
+    /// * `account_pubkey` - The account accessing the image
+    /// * `group_id` - The MLS group ID
+    /// * `image_hash` - SHA-256 hash of the encrypted image
+    /// * `image_key` - Encryption key
+    /// * `image_nonce` - Encryption nonce
+    ///
+    /// # Returns
+    /// Path to the cached image file
+    async fn download_and_cache_group_image(
```
jgmontoya, line 642: This method is way too large, it also seems out of place to be here in groups. It's doing a bunch of things I'd expect to be in `media_files` instead of `groups`

src/whitenoise/groups.rs
```diff
@@ -502,72 +498,247 @@ impl Whitenoise {
...
+        if !response.status().is_success() {
+            return Err(WhitenoiseError::BlossomDownload(format!(
+                "Download failed with status: {}",
+                response.status()
+            )));
+        }
+
+        let encrypted_data = response.bytes().await.map_err(|e| {
+            WhitenoiseError::BlossomDownload(format!("Failed to read response: {}", e))
+        })?;
```
jgmontoya: Maybe we do need a custom whitenoise BlossomClient "layer" to do this, it feels very out of place here.

src/whitenoise/media_files.rs
```diff
@@ -0,0 +1,335 @@
...
+                file_path: &file_path,
+                file_hash: &upload.file_hash,
+                mime_type: upload.mime_type,
+                media_type: upload.media_type,
+                blossom_url: upload.blossom_url,
+                dimensions: upload.dimensions,
+                blurhash: upload.blurhash,
+            },
+        )
+        .await?;
```
jgmontoya: If this fails we should clean up the storage (if needed)

src/whitenoise/storage/media_files.rs
```diff
@@ -0,0 +1,314 @@
...
+    ///
+    /// # Errors
+    /// Returns error if filesystem operations fail
+    pub(crate) async fn store_file(
+        &self,
+        group_id: &GroupId,
+        subdirectory: &str,
+        filename: &str,
+        data: &[u8],
+    ) -> Result<PathBuf> {
```
jgmontoya: It could be useful for this method that it returns whether the file was already created or if it created it

src/whitenoise/media_files.rs
jgmontoya: Shouldn't this call the `record_in_database` method?

src/whitenoise/media_files.rs
```diff
@@ -0,0 +1,335 @@
...
+    ///
+    /// # Arguments
+    /// * `group_id` - The MLS group ID
+    /// * `subdirectory` - Subdirectory within the group
+    /// * `filename` - The filename to look for
+    ///
+    /// # Returns
+    /// The path if the file exists, None otherwise
+    #[allow(dead_code)]
+    pub(crate) fn get_file_path(
```
jgmontoya: If this is dead code why do we want it around?

src/whitenoise/storage/media_files.rs
```diff
@@ -0,0 +1,314 @@
...
+        // Write atomically: write to temp file, then rename
+        let temp_path = file_path.with_extension("tmp");
+
+        fs::write(&temp_path, data).await.map_err(|e| {
+            WhitenoiseError::MediaCache(format!("Failed to write file data: {}", e))
+        })?;
+
+        fs::rename(&temp_path, &file_path).await.map_err(|e| {
+            WhitenoiseError::MediaCache(format!("Failed to rename temp file: {}", e))
+        })?;
```
jgmontoya: I think there could be race conditions lurking here 👀

src/whitenoise/storage/media_files.rs
```diff
@@ -0,0 +1,314 @@
...
+    /// * `data_dir` - The application data directory
+    ///
+    /// # Returns
+    /// A new MediaFileStorage instance with cache directory at `<data_dir>/media_cache/`
+    pub(crate) fn new(data_dir: &Path) -> Result<Self> {
+        let cache_dir = data_dir.join("media_cache");
+
+        // Create cache directory if it doesn't exist
+        if !cache_dir.exists() {
+            std::fs::create_dir_all(&cache_dir).map_err(|e| {
```
jgmontoya: this should use the tokio function instead of the std one

src/whitenoise/storage/media_files.rs
```diff
@@ -0,0 +1,314 @@
...
+        subdirectory: &str,
+        prefix: &str,
+    ) -> Option<PathBuf> {
+        let group_dir = self.get_group_dir(group_id, subdirectory);
+
+        if !group_dir.exists() {
+            return None;
+        }
+
+        if let Ok(entries) = std::fs::read_dir(&group_dir) {
```
jgmontoya: this should use the tokio function instead of the std one

docs/storage-architecture.md
erskingardner: I like organization. I can imagine that we might want to selectively delete media based on what type it is. That said, I guess since we're storing the type of media in the media_files table, we can easily find the right files to delete.

src/whitenoise/media_files.rs
jgmontoya: Related to this, after chatting with claude about it, I think we can just leave it for now. We can implement some cleanup/garbage collection later in a way that is not race-condition prone.

src/whitenoise/storage/media_files.rs
jgmontoya: Given [this](https://github.com/parres-hq/whitenoise/pull/383#discussion_r2427713726):

> Related to this, after chatting with claude about it, I think we can just leave it for now. We can implement some cleanup/garbage collection later in a way that is not race-condition prone.

I think this is no longer needed

src/whitenoise/storage/media_files.rs
```diff
@@ -0,0 +1,328 @@
...
+            WhitenoiseError::MediaCache(format!("Failed to write file data: {}", e))
+        })?;
+
+        // Persist atomically renames the temp file to the final path
+        temp_file.persist(&file_path).map_err(|e| {
+            WhitenoiseError::MediaCache(format!("Failed to persist temp file: {}", e))
+        })?;
+
+        Ok(file_path)
+    }
```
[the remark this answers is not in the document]
jgmontoya: I don't think there's an issue here

src/whitenoise/database/media_files.rs
```diff
@@ -0,0 +1,444 @@
...
+{
+    fn from_row(row: &'r R) -> std::result::Result<Self, sqlx::Error> {
+        let id: i64 = row.try_get("id")?;
+        let mls_group_id_bytes: Vec<u8> = row.try_get("mls_group_id")?;
+        let account_pubkey_str: String = row.try_get("account_pubkey")?;
+        let file_path_str: String = row.try_get("file_path")?;
+        let file_hash_hex: String = row.try_get("file_hash")?;
+
+        // Parse MLS group ID
+        let mls_group_id = GroupId::from_slice(&mls_group_id_bytes);
```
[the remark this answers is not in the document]
jgmontoya: I think GroupId::from_slice does not panic on variable-length input, it can handle any length

src/whitenoise/groups.rs
```diff
@@ -502,82 +522,516 @@ impl Whitenoise {
...
+            .record_in_database(account_pubkey, group_id, cached_path, upload)
+            .await
+        {
+            tracing::warn!(
+                target: "whitenoise::groups::link_cached_image_with_detection",
+                "Failed to record file in database: {}",
+                e
+            );
+        }
+    }
```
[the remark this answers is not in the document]
jgmontoya: The (empty or not) data is not stored in the database, in this case the data is files which are stored using the filesystem

src/whitenoise/event_processor/event_handlers/handle_mls_message.rs
```diff
@@ -22,6 +23,20 @@ impl Whitenoise {
                   "Handled MLS message - Result: {:?}",
                   result
                 );
+
+                if let MessageProcessingResult::Commit = result {
+                    let groups_result = self.groups(account, true).await;
+                    if let Ok(groups) = groups_result {
+                        // Spawn background tasks for image caching to avoid blocking event processing
+                        // TODO: This is a blunt instrument, in future we can improve this by looking up the specific group the Commit was against.
```
erskingardner, line 31: I really hate this part still. I don't know how we can get the group_id without re-architecting MDK though. Would love to hear if you have any ideas

src/whitenoise/groups.rs
```diff
@@ -1196,5 +1790,139 @@ mod tests {
...
+            .unwrap();
+
+        assert!(cached_again.is_some());
+        assert_eq!(
+            cached_again.unwrap(),
+            cached_path,
+            "Second retrieval should return same cached path"
+        );
     }
 }
```
erskingardner, line 1928: This file is getting very big. We don't need to do it now, but we should think about splitting it up soon.

jgmontoya: agreed

PR #385, opened 2025-10-10
feat(nostr_manager): add background_publish_mls_message_to so we can send messages without blocking

jgmontoya:
also refactors away some usages of Relay from the nostr_manager

## Summary by CodeRabbit

- New Features
  - Publishing of MLS/events now runs as a background, fire-and-forget task to keep the app responsive.

- Refactor
  - Relay handling unified to use direct relay URLs across publishing flows.
  - Removed blocking MLS-specific publish paths in favor of background publishing for group and message flows.

- Performance
  - Reduced blocking during publish operations for a snappier experience.

- Tests
  - Added coverage for publish and background-publish success/error scenarios.

- Style
  - Minor import cleanup.

### DIFF
--- a/src/nostr_manager/publisher.rs
+++ b/src/nostr_manager/publisher.rs
@@ -11,50 +11,58 @@ use crate::{
 };
 
 impl NostrManager {
-    /// Publishes an MLS commit event to the specified relays.
+    /// Publishes an event to the specified relays in a background task.
     ///
-    /// This method allows publishing an MLS commit event (which is already signed) to a list of relay URLs. It ensures that the client
-    /// is connected to all specified relays before attempting to publish the event and then tracks that we published the event to the database.
-    /// This is the public crate interface to ensure you're using the correct publishing method.
-    pub(crate) async fn publish_mls_commit_to(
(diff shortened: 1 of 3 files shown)


src/nostr_manager/publisher.rs
```diff
@@ -34,11 +34,47 @@ impl NostrManager {
...
         self.publish_event_to(event, account, relays).await
     }
 
+    /// Publishes an MLS message event to the specified relays in a background task.
+    ///
+    /// This is a fire-and-forget operation that spawns a background task to publish the message
+    /// without blocking the caller. Errors are logged but not returned. This is useful for
+    /// scenarios where you want to queue a publish operation but don't need to wait for completion.
+    ///
+    /// The method clones the necessary data to ensure `'static` lifetime for the spawned task.
```
erskingardner, line 20: I think the method name should be more generic, there's nothing specific about this being only for mls messages right? 

what about just `background_publish_event_to`?

jgmontoya: nice catch

PR #390, opened 2025-10-20
Feat/upload chat media

jgmontoya:
## Summary by CodeRabbit

* **New Features**
  * Added chat media upload functionality with optional Blossom server support

* **Improvements**
  * Group image uploads accept an optional Blossom URL with fallback
  * Media uploads now return richer media records including file metadata and linked cached files
  * Public API now exposes MediaFile and FileMetadata types

* **Tests**
  * New integration scenario and test cases validating end-to-end chat media upload and metadata recording


src/whitenoise/media_files.rs
```diff
@@ -29,50 +27,14 @@ pub(crate) struct MediaFileUpload<'a> {
...
     pub blossom_url: Option<&'a str>,
+    /// Optional Nostr key (hex-encoded secret key) used for upload authentication/cleanup
+    /// For group images: deterministically derived from image_key (stored for convenience)
+    /// For chat images: randomly generated per upload (must be stored)
+    pub nostr_key: Option<String>,
     /// Optional file metadata (original filename, dimensions, blurhash, duration, etc.)
     pub file_metadata: Option<&'a FileMetadata>,
 }
 
-impl<'a> From<(&'a GroupImageUpload, &'a [u8], &'a str, &'a str)> for MediaFileUpload<'a> {
```
erskingardner, line 36: Were these not used anywhere?

src/whitenoise/groups.rs
```diff
@@ -1054,6 +1052,136 @@ impl Whitenoise {
...
+    /// * `account` - The account uploading the image
+    /// * `group_id` - The ID of the group where the image will be used
+    /// * `file_path` - Path to the image file to upload
+    /// * `blossom_server_url` - Optional Blossom server URL (uses default if None)
+    /// * `options` - Optional media processing options (defaults to standard options if None)
+    ///
+    /// # Returns
+    /// * `Ok(MediaFile)` - MediaFile record containing file hash and metadata
+    /// * `Err(WhitenoiseError)` - If image validation, upload, or caching fails
+    pub async fn upload_chat_media(
```
erskingardner, line 1074: This all looks good. I guess we're assuming that the UI will upload a single image a time or that, if a user can select multiple images, that this method will be called in series for each image? Is that right?

erskingardner (commented):
Looks good just a small question

src/whitenoise/media_files.rs
jgmontoya: nope

src/whitenoise/groups.rs
jgmontoya: yeah I assumed that

jgmontoya: I guess whenever we need to handle multiple on a single call we can just change this to receive a vec of paths, wouldn't worry too much before we actually need it because with batches it gets more cumbersome to handle failures (what happens if some of the selected images uploaded fine and others failed?)

PR #397, opened 2025-11-01
Feat/download chat media

jgmontoya:
## Summary by CodeRabbit

* **New Features**
  * Idempotent chat media download (decrypt, cache, integrity checks); media listing for groups; blossom-URL hash extraction; explicit tracking of encrypted and original file hashes; API to fetch chat media.

* **Bug Fixes**
  * Stronger validations and clearer errors for missing/invalid metadata, URL/hash mismatches; faster persistence of parsed media references from incoming messages; attachment resolution now uses original hashes.

* **Tests**
  * Expanded integration tests for upload, receive, download, metadata, hashing, idempotency, and edge cases; new download/receive scenarios.

* **Chores**
  * Database migration to add explicit encrypted_file_hash and original_file_hash columns.

### DIFF
--- a/src/whitenoise/database/media_files.rs
+++ b/src/whitenoise/database/media_files.rs
@@ -418,7 +418,6 @@ impl MediaFile {
     /// let cached_path = PathBuf::from("/cache/media/abc123.jpg");
     /// let updated = MediaFile::update_file_path(&db, media_file.id.unwrap(), &cached_path).await?;
     /// ```
-    #[allow(dead_code)] // Will be used in download_chat_media implementation
     pub(crate) async fn update_file_path(
         database: &Database,
         id: i64,
--- a/src/whitenoise/media_files.rs
+++ b/src/whitenoise/media_files.rs
@@ -1,7 +1,11 @@
 use std::path::{Path, PathBuf};
 
-use mdk_core::GroupId;
-use nostr_sdk::{PublicKey, prelude::*};
+use mdk_core::{
+    GroupId,
+    encrypted_media::{manager::EncryptedMediaManager, types::MediaReference},
+    prelude::MdkStorageProvider,
+};
+use nostr_sdk::prelude::*;
 
 pub use crate::whitenoise::database::media_files::MediaFile;
 use crate::whitenoise::{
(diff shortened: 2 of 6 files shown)


src/whitenoise/groups.rs
```diff
@@ -862,6 +874,84 @@ impl Whitenoise {
...
+    ) -> Result<Vec<u8>> {
+        // Extract filename required for MDK AAD (cryptographically bound)
+        let filename = media_file
+            .file_metadata
+            .as_ref()
+            .and_then(|m| m.original_filename.as_deref())
+            .ok_or_else(|| {
+                WhitenoiseError::MediaCache("Missing required filename metadata".to_string())
+            })?;
+        // Parse encrypted_file_hash for download verification
```
pepina-dev, line 907: I don't fully understand for what the encrypted_file_hash is necessary... couldn't this be checked after decryption? comparing with the original file hash? And that way we can save only one hash?

pepina-dev: Aah wait we talked about it... was it because the group image needs the encrypted one in the event? So we save both always now?

erskingardner: yup

src/whitenoise/media_files.rs
```diff
@@ -1,27 +1,165 @@
 use std::path::{Path, PathBuf};
 
 use mdk_core::GroupId;
-use nostr_sdk::PublicKey;
+use nostr_sdk::{PublicKey, prelude::*};
```
erskingardner: nit: we can remove `PublicKey` here, it's included in the prelude I think.

src/whitenoise/media_files.rs
```diff
@@ -1,27 +1,165 @@
...
+    blurhash: Option<String>,
+}
+
+/// Extracts imeta tags from Nostr event tags (MIP-04 format)
+///
+/// Parses tags with the format:
+/// ["imeta", "url <blossom_url>", "x <original_hash>", "m <mime>", "name <filename>", ...]
+///
+/// Returns an empty vector if no valid imeta tags are found.
+fn extract_imeta_tags(tags: &Tags) -> Vec<ImetaTag> {
```
erskingardner: We have a method in the media_manager in MDK that does this same thing... do we need this method here or should we use the other method? 

https://github.com/parres-hq/mdk/blob/df92478b29d54bb166e99e835afe7d03eaab6e43/crates/mdk-core/src/encrypted_media/manager.rs#L211

src/whitenoise/database/media_files.rs
```diff
@@ -320,6 +338,119 @@ impl MediaFile {
...
+    /// * `Ok(MediaFile)` - The updated media file record
+    /// * `Err(WhitenoiseError)` - If the record doesn't exist or database error
+    ///
+    /// # Example
+    /// ```ignore
+    /// // After downloading and caching a file
+    /// let cached_path = PathBuf::from("/cache/media/abc123.jpg");
+    /// let updated = MediaFile::update_file_path(&db, media_file.id.unwrap(), &cached_path).await?;
+    /// ```
+    #[allow(dead_code)] // Will be used in download_chat_media implementation
```
erskingardner: i think we have a download implementation now?

jgmontoya: haha yeah, good catch!

src/whitenoise/media_files.rs
jgmontoya: done

erskingardner (approved):

PR #401, opened 2025-11-07
Fetch user metadata in background with TTL

erskingardner:
> [!NOTE]
> Fetch user metadata and relays in the background, refreshing only when metadata is stale per a 24h TTL, and update tests to await background completion.
> 
> - **Users/Backend**:
>   - Add `METADATA_TTL_HOURS` (24h) and `User::needs_metadata_refresh()` to determine staleness of `metadata`.
>   - Change `Whitenoise::find_or_create_user_by_pubkey()` to start `background_fetch_user_data()` for new users and for existing users with stale metadata; skip sync when fresh.
>   - `background_fetch_user_data()` spawns a task to update relay lists first, then sync metadata, and refresh global subscription.
> - **Tests**:
>   - Integration test `user_discovery/find_or_create_user.rs`: wait/retry for background metadata and NIP-65 relay fetch completion before assertions.
>   - Add unit tests covering `needs_metadata_refresh()` (default, fresh, stale, boundary).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 789625974afcea971c64c6da1009877902908a6c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Added background synchronization for user metadata with TTL-based freshness validation.
  * Introduced retry mechanism for fetching and polling user relay information.

* **Improvements**
  * Enhanced reliability of user data retrieval through asynchronous background processing instead of immediate synchronous operations.

### DIFF
--- a/src/integration_tests/test_cases/user_discovery/find_or_create_user.rs
+++ b/src/integration_tests/test_cases/user_discovery/find_or_create_user.rs
@@ -139,7 +139,10 @@ impl TestCase for FindOrCreateUserTestCase {
                         )))
                     }
                 },
-                &format!("wait for background metadata fetch for user {}", &test_pubkey.to_hex()[..8]),
+                &format!(
+                    "wait for background metadata fetch for user {}",
+                    &test_pubkey.to_hex()[..8]
+                ),
             )
             .await?;
         }
@@ -195,7 +198,10 @@ impl TestCase for FindOrCreateUserTestCase {
                         Ok(relays)
                     }
                 },
-                &format!("wait for background relay fetch for user {}", &test_pubkey.to_hex()[..8]),
+                &format!(
+                    "wait for background relay fetch for user {}",
+                    &test_pubkey.to_hex()[..8]
+                ),
             )
             .await?;
 


src/whitenoise/users.rs
```diff
@@ -604,33 +631,43 @@ impl Whitenoise {
...
                 tracing::warn!(
                     target: "whitenoise::users::find_or_create_user_by_pubkey",
-                    "Failed to update relay lists for new user {}: {}",
+                    "Failed to start background fetch for new user {}: {}",
                     user.pubkey,
                     e
                 );
             }
-            if let Err(e) = self.refresh_global_subscription_for_user(&user).await {
+        } else if user.needs_metadata_refresh() {
```
dannym-arx, line 646: rather than have a cache TTL for deciding when to fetch, can't we always try to fetch it in the background?

erskingardner: I thought about that first but given the frequency of these find_or_create calls, I think that we'd end up triggering background refreshes way too frequently. 

We should definitely keep the synchronous sync_metadata calls for locations where we want to force a fresh fetch (like when creating groups).

dannym-arx: that makes sense!

dannym-arx (approved):
looks good to me

PR #403, opened 2025-11-07
Add param to force sync_metadata on find_or_create user

erskingardner:
> [!NOTE]
> Adds `UserSyncMode` to `find_or_create_user_by_pubkey` enabling blocking or background sync with TTL-based refresh, plus test-only timestamp setter and comprehensive integration/unit tests.
> 
> - **Users API**:
>   - Introduce `UserSyncMode` (`Blocking` | `Background`).
>   - Change `Whitenoise::find_or_create_user_by_pubkey` to accept `sync_mode` and delegate to `sync_user_blocking` or `sync_user_background`.
>   - Background path: immediate return; starts `background_fetch_user_data`; refresh only if `needs_metadata_refresh()` based on 24h TTL.
>   - Blocking path: synchronously update relay lists (for new users), refresh global subscription, and `sync_metadata` before returning.
>   - Add test-only `set_user_updated_at_for_testing` (behind `integration-tests`) to manipulate `updated_at`.
>   - Re-export `UserSyncMode` in `lib.rs`.
> - **Integration Tests**:
>   - Add scenarios: `find_or_create_user_background`, `find_or_create_user_blocking`, `find_or_create_user_force_sync_existing`, `find_or_create_user_stale_metadata`; wire into `UserDiscoveryScenario`.
>   - Update subscription processing test to call `find_or_create_user_by_pubkey(..., UserSyncMode::Background)`.
> - **Unit Tests**:
>   - Expand tests for TTL refresh logic, relay sync behaviors, and new `find_or_create` modes; remove outdated tests relying on old signature.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2ae556593831f0da2bfc367ac1ab882919892c49. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Added synchronization mode option for user discovery and creation.
  * Blocking mode ensures metadata is available immediately upon user creation.
  * Background mode enables non-blocking operations with asynchronous metadata fetching.
  * Automatic background refresh for outdated user metadata.

* **Tests**
  * Expanded integration tests covering blocking, background, force-sync, and stale-metadata scenarios.

### DIFF
--- a/src/whitenoise/users.rs
+++ b/src/whitenoise/users.rs
@@ -676,15 +676,15 @@ impl Whitenoise {
         let (user, created) = User::find_or_create_by_pubkey(pubkey, &self.database).await?;
 
         if sync_mode == UserSyncMode::Blocking {
-            self.sync_user_blocking(&user, created).await?;
+            let updated_user = self.sync_user_blocking(&user, created).await?;
+            Ok(updated_user)
         } else {
             self.sync_user_background(&user, created).await?;
+            Ok(user)
         }
-
-        Ok(user)
     }
 
-    async fn sync_user_blocking(&self, user: &User, is_new: bool) -> Result<()> {
+    async fn sync_user_blocking(&self, user: &User, is_new: bool) -> Result<User> {
         tracing::debug!(
             target: "whitenoise::users::sync_user_blocking",
             "Force sync requested for user {}, performing blocking metadata and relay sync",
@@ -724,7 +724,7 @@ impl Whitenoise {
             );
         }
 
-        Ok(())
+        Ok(user_clone)
     }
 
     async fn sync_user_background(&self, user: &User, is_new: bool) -> Result<()> {
--- a/src/integration_tests/test_cases/user_discovery/find_or_create_user_background.rs
+++ b/src/integration_tests/test_cases/user_discovery/find_or_create_user_background.rs
@@ -63,7 +63,7 @@ impl TestCase for FindOrCreateUserBackgroundModeTestCase {
     async fn run(&self, context: &mut ScenarioContext) -> Result<(), WhitenoiseError> {
         let test_pubkey = self.test_keys.public_key();
         tracing::info!(
-            "Testing find_or_create_user with force_sync=false for pubkey: {}",
+            "Testing find_or_create_user with UserSyncMode::Background for pubkey: {}",
             test_pubkey
         );
 


src/whitenoise/users.rs
```diff
@@ -630,42 +640,86 @@ impl Whitenoise {
...
+            if created {
+                // For new users, sync relay lists first, then metadata
+                if let Err(e) = user.update_relay_lists(self).await {
+                    tracing::warn!(
+                        target: "whitenoise::users::find_or_create_user_by_pubkey",
+                        "Failed to sync relay lists for new user {}: {}",
+                        user.pubkey,
+                        e
+                    );
+                }
```
jgmontoya: Question: Why do we only update the relay lists if the user is new?

src/whitenoise/users.rs
```diff
@@ -630,42 +640,86 @@ impl Whitenoise {
...
+                if let Err(e) = self.refresh_global_subscription_for_user(&user).await {
+                    tracing::warn!(
+                        target: "whitenoise::users::find_or_create_user_by_pubkey",
+                        "Failed to refresh global subscription for new user {}: {}",
+                        user.pubkey,
+                        e
+                    );
+                }
```
jgmontoya: We don't do subscriptions for all of the users in our db (just follows + groups peers), so this is a bit odd.

We could try to handle all users through subscriptions, but I'm not sure if that's what we want (there could be a bunch of users created in the db that we're not really interacting with).

src/whitenoise/users.rs
```diff
@@ -630,42 +640,86 @@ impl Whitenoise {
     /// This method will return an error if:
     /// - There's a database connection or query error
     /// - The public key format is invalid (though this is typically caught at the type level)
-    pub async fn find_or_create_user_by_pubkey(&self, pubkey: &PublicKey) -> Result<User> {
-        let (user, created) = User::find_or_create_by_pubkey(pubkey, &self.database).await?;
+    /// - Network errors occur during forced synchronous syncing
+    pub async fn find_or_create_user_by_pubkey(
```
jgmontoya, line 671: This method is getting too big/convoluted, can we refactor it by extracting some private methods?

Something like this perhaps:

```rs
pub async fn find_or_create_user_by_pubkey(
    &self,
    pubkey: &PublicKey,
    force_sync: bool,
) -> Result<User> {
    let (user, created) = User::find_or_create_by_pubkey(pubkey, &self.database).await?;
    
    if force_sync || created || user.needs_metadata_refresh() {
        self.sync_user_data(&user, force_sync, created).await?;
    }
    
    Ok(user)
}

async fn sync_user_data(&self, user: &User, blocking: bool, is_new: bool) -> Result<()> {
    // Sync logic
}
```

jgmontoya (commented):
Left a couple of comments/questions

src/whitenoise/users.rs
erskingardner:
```rust
pub async fn find_or_create_user_by_pubkey(&self, pubkey: &PublicKey) -> Result<User> {
        let (mut user, created) = User::find_or_create_by_pubkey(pubkey, &self.database).await?;
        if created {
            if let Err(e) = user.update_relay_lists(self).await {
                tracing::warn!(
                    target: "whitenoise::users::find_or_create_user_by_pubkey",
                    "Failed to update relay lists for new user {}: {}",
                    user.pubkey,
                    e
                );
            }
            if let Err(e) = self.refresh_global_subscription_for_user(&user).await {
                tracing::warn!(
                    target: "whitenoise::users::find_or_create_user_by_pubkey",
                    "Failed to refresh global subscription for new user {}: {}",
                    user.pubkey,
                    e
                );
            }
        }
        if let Err(e) = user.sync_metadata(self).await {
            tracing::warn!(
                target: "whitenoise::users::find_or_create_user_by_pubkey",
                "Failed to sync metadata for new user {}: {}",
                user.pubkey,
                e
            );
        }
        Ok(user)
    }
```

This was the method before I changed anything. That call for refreshing subs was yours I think...

src/whitenoise/users.rs
erskingardner: I think that it's ok - it's not that big and splitting it up would just make you jump around to understand the conditional properly. If it gets any bigger in the future we can break it up.

src/whitenoise/users.rs
erskingardner: Because if we have a set of the user's relays (which we do after they're first created and we fetch them) then we're very likely to find the user's events. We keep this up to date (for the users we most care about) from the global subs.

src/whitenoise/users.rs
erskingardner: In any case, I looked through and it's because we need to update the global batches of users that are in the subs. If we create a new user, and it's one that we care about - we should add it to the global subscriptions batch it belongs in, which is what this method does.

jgmontoya: I was remembering this incorrectly. It turns out that we are using subscriptions for all users, so this makes sense.

src/whitenoise/users.rs
jgmontoya: Yeah I was remembering the global subs part incorrectly, this makes sense 👍

src/whitenoise/users.rs
```diff
@@ -710,6 +792,53 @@ impl Whitenoise {
...
+
+        tracing::debug!(
+            target: "whitenoise::users::set_user_updated_at_for_testing",
+            "Set updated_at for user {} to {} (TEST ONLY)",
+            pubkey,
+            updated_at
+        );
+
+        Ok(())
+    }
```
jgmontoya, line 853: I don't like this, can we set the integration tests in a way where this is not needed?

src/integration_tests/test_cases/user_discovery/find_or_create_user.rs
jgmontoya, line 1: Each individual TestCase should be on its own file. In this case I think we could probably use a single TestCase and have its behavior change a bit using the builder pattern as we do for other tests

src/whitenoise/users.rs
dannym-arx: I feel like that's a clean way to do it personally. The alternatives are way messier

1) it's a function that's only callable if the flag integration-tests is on. 
2) the alternatives would be to mock time everywhere in the codebase (which maybe is something we should do eventually, but is outside of scope for this pr imo), or let tests override TTL windows, which is basically the same thing but with extra steps and worse API.

This approach is explicit, feature-gated so it literally can't exist in production builds, and isolated to exactly where we need it. The method name screams "this is for testing" and the feature gate enforces it at compile time.

src/whitenoise/users.rs
dannym-arx: honestly I kinda agree with jgmontoya, his approach would be much cleaner, and would probably result in fewer bugs in the future

src/whitenoise/users.rs
erskingardner: I tried to do a few things but this was by far the easiest and most straightforward. We need some way of forcing the passage of time if we want to test the TTL. 🤷‍♂️

jgmontoya: ok fair

PR #405, opened 2025-11-14
test(benchmarks/scenarios/message_aggregation): add benchmark test for fetch_aggregated_messages_for_group

jgmontoya:
image

## Summary by CodeRabbit

* **Tests**
  * Added a Message Aggregation Performance benchmark covering regular and media messages, reactions, deletions, and replies.
  * Added a new test case to measure fetching aggregated messages.
  * Registered the new scenario to run alongside existing benchmarks.
* **Chores**
  * Reduced the relay open-files limit in dev configuration.

### DIFF
--- a/dev/strfry.conf
+++ b/dev/strfry.conf
@@ -47,7 +47,7 @@ relay {
     port = 7777
 
     # Set OS-limit on maximum number of open files/sockets (if 0, don't attempt to set) (restart required)
-    nofiles = 1000000
+    nofiles = 65536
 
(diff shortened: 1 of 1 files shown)


dev/strfry.conf
```diff
@@ -47,7 +47,7 @@ relay {
     port = 7777
 
     # Set OS-limit on maximum number of open files/sockets (if 0, don't attempt to set) (restart required)
-    nofiles = 1000000
+    nofiles = 65536
```
erskingardner, line 50: Curious. I remember this causing issues with tests when we had it set to something lower but let's see if things work ok like this

PR #412, opened 2025-12-07
Fix/aggregated messages reaction deletion

jgmontoya:
## Summary by CodeRabbit

* **New Features**
  * Reaction deletion flow: users can remove reactions and see them disappear from parent messages.

* **Tests**
  * New integration tests verify reaction deletion and confirm removed reactions no longer appear.

* **Refactor**
  * Message/reaction handling and aggregated-message handling updated so reaction state and cached views stay in sync after deletions.

### DIFF
--- a/src/nostr_manager/parser.rs
+++ b/src/nostr_manager/parser.rs
@@ -110,12 +110,13 @@ mod tests {
     use std::{sync::Arc, time::Duration};
 
     use super::*;
-    use crate::whitenoise::event_tracker::WhitenoiseEventTracker;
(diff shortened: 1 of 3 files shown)


src/whitenoise/message_aggregator/reaction_handler.rs
dannym-arx, line 1: why are we doing one reaction per user? I think that allowing multiple different types of reactions makes sense (at least, I find myself doing that often in apps that support it)

not sure if this needs to be addressed, per se, but personally I like having that feature

PR #414, opened 2025-12-09
Message streams

jgmontoya:
## Summary by CodeRabbit

* **New Features**
  * Real-time group message streaming: initial snapshot plus live updates for new messages, reactions, and deletions; subscribe-to-group-messages API and public update types.

* **Improvements**
  * Event-driven caching and broadcast pipeline for more granular, timely message/reaction/deletion handling and improved orphan resolution.
  * Added a manager for per-group broadcast streams with lazy creation and cleanup.

* **Tests**
  * New end-to-end streaming scenario and test cases validating snapshots, stream updates, and timing.

* **Chores**
  * Reorganized test modules and moved shared delete-message test case; registered the new streaming scenario.

### DIFF
--- a/src/whitenoise/message_streaming/manager.rs
+++ b/src/whitenoise/message_streaming/manager.rs
@@ -30,19 +30,21 @@ impl MessageStreamManager {
     }
 
     pub fn emit(&self, group_id: &GroupId, update: MessageUpdate) {
-        if let Some(sender) = self.streams.get(group_id) {
-            // Attempt to send; if all receivers dropped, clean up
-            if let Err(error) = sender.send(update)
-                && sender.receiver_count() == 0
+        if let Some(sender) = self.streams.get(group_id)
+            && sender.send(update).is_err()
+        {
+            drop(sender);
+            // Atomically check and remove to avoid race with concurrent subscribe()
+            if self
+                .streams
+                .remove_if(group_id, |_, s| s.receiver_count() == 0)
+                .is_some()
             {
                 tracing::debug!(
                     target: "whitenoise::message_streaming",
-                    "Cleaning up stream for group {} (no active receivers, error: {})",
+                    "Cleaned up stream for group {} (no active receivers)",
                     hex::encode(group_id.as_slice()),
-                    error
                 );
-                drop(sender);
-                self.streams.remove(group_id);
             }
         }
     }


src/whitenoise/mod.rs
```diff
@@ -538,6 +543,63 @@ impl Whitenoise {
...
+                Err(broadcast::error::TryRecvError::Empty) => break,
+                Err(broadcast::error::TryRecvError::Lagged(_)) => continue,
+                Err(broadcast::error::TryRecvError::Closed) => {
+                    // Channel closed unexpectedly - should be unreachable since we hold a receiver
+                    return Err(WhitenoiseError::Other(anyhow::anyhow!(
+                        "Message stream closed unexpectedly during subscription"
+                    )));
+                }
+            }
+        }
```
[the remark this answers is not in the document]
jgmontoya: I don't think we need to worry about this. The time window is very tiny between fetching the aggregated messages and opening the subscription

src/whitenoise/message_streaming/manager.rs
```diff
@@ -0,0 +1,180 @@
...
+            .or_insert_with(|| broadcast::channel(BUFFER_SIZE).0)
+            .subscribe()
+    }
+
+    pub fn emit(&self, group_id: &GroupId, update: MessageUpdate) {
+        if let Some(sender) = self.streams.get(group_id) {
+            // Attempt to send; if all receivers dropped, clean up
+            if sender.send(update).is_err() && sender.receiver_count() == 0 {
+                drop(sender);
+                self.streams.remove(group_id);
```
dannym-arx: should this have a debug log statement?

jgmontoya: done!

src/whitenoise/event_processor/event_handlers/handle_mls_message.rs
```diff
@@ -106,174 +128,255 @@ impl Whitenoise {
...
+        AggregatedMessage::update_reactions(
+            &target.id,
+            group_id,
+            &target.reactions,
+            &self.database,
+        )
+        .await?;
 
-        Ok(())
+        Ok(Some(target))
```
mubarakcoded, line 245: Here, two reactions arriving simultaneously for the same message will both read the current state, each add their reaction locally, then write. The second write overwrites the first. Emitted updates will show incomplete reaction state. 
So users may see stale reaction counts in real-time updates.

jgmontoya: The event processor goes through each event serially, so I don't think we have that problem

src/whitenoise/event_processor/event_handlers/handle_mls_message.rs
```diff
@@ -106,174 +128,255 @@ impl Whitenoise {
...
-                    .await?;
+    /// Cache a new chat message and return it for emission.
+    ///
+    /// Processes the message through the aggregator, inserts into database,
+    /// and applies any orphaned reactions/deletions that arrived before this message.
+    async fn cache_chat_message(
+        &self,
+        group_id: &GroupId,
+        message: &Message,
+    ) -> Result<Option<ChatMessage>> {
```
mubarakcoded: Here the return type is a bit confusing, there's no code path returning Ok(None), the function always returns Some. Should be Result<ChatMessage> or we document when None could occur.

src/whitenoise/mod.rs
```diff
@@ -538,6 +543,64 @@ impl Whitenoise {
...
+            .collect();
+
+        loop {
+            match updates.try_recv() {
+                Ok(update) => {
+                    // Apply update: insert or replace by message ID
+                    messages_map.insert(update.message.id.clone(), update.message);
+                }
+                Err(broadcast::error::TryRecvError::Empty) => break,
+                Err(broadcast::error::TryRecvError::Lagged(_)) => continue,
```
mubarakcoded, line 584: When buffer overflows, messages are lost with no indication. We can a debug log here for observability.

src/whitenoise/message_streaming/manager.rs
```diff
@@ -0,0 +1,188 @@
...
+        let _rx2 = manager.subscribe(&group_id);
+
+        // Should still only have one entry
+        assert_eq!(manager.streams.len(), 1);
+
+        // Both should receive from same sender (receiver_count should be 2)
+        let sender = manager.streams.get(&group_id).unwrap();
+        assert_eq!(sender.receiver_count(), 2);
+    }
+
```
mubarakcoded, line 123: can we add a test that test that verifies both subscribers actually receive the emitted update, not just that they share a sender.

src/whitenoise/event_processor/event_handlers/handle_mls_message.rs
mubarakcoded: You're right, I missed that.

src/whitenoise/mod.rs
jgmontoya: I agree, however it's very unlikely that the buffer would overflow at this point. This is just to catch messages between the stream is open and the aggregated messages are fetched, so the time window is tiny

src/whitenoise/event_processor/event_handlers/handle_mls_message.rs
jgmontoya: I did it on purpose just to have it be "symmetrical" with `cache_reaction` but now that I'm looking at it I don't think it was a good idea, I will simplify the return type

src/whitenoise/message_streaming/manager.rs
jgmontoya: Good idea, I will add a test

src/whitenoise/mod.rs
mubarakcoded: Yeah. Thats fair point and I agree with you

mubarakcoded (approved):

PR #416, opened 2025-12-12
Feat/chat list method

jgmontoya:
## Summary by CodeRabbit

* **New Features**
  * Chat list view: shows groups and DMs sorted by recent activity with names, last-message previews, creation dates, and participant images
  * Direct messages resolve and display the other participant’s name from user metadata

* **API/Public Types**
  * Added message-summary and chat-list item types and an API to fetch a user’s chat list and last-message summaries

* **Tests**
  * Expanded unit and integration tests, plus a new chat-list integration scenario covering creation, DM/group flows, ordering, and image resolution

### DIFF
--- a/src/whitenoise/secrets_store.rs
+++ b/src/whitenoise/secrets_store.rs
@@ -131,7 +131,7 @@ impl SecretsStore {
     /// * Setting the password in the keyring fails
     /// * The secret key cannot be retrieved from the keypair
     pub fn store_private_key(&self, keys: &Keys) -> Result<(), SecretsStoreError> {
-        if cfg!(target_os = "android") {
+        if cfg!(target_os = "android") || cfg!(test) {
             let mut secrets = self.read_secrets_file().unwrap_or(json!({}));
             let obfuscated_key = self.obfuscate(keys.secret_key().to_secret_hex().as_str());
             secrets[keys.public_key().to_hex()] = json!(obfuscated_key);
@@ -168,7 +168,7 @@ impl SecretsStore {
     /// * Parsing the private key into a `Keys` object fails
     pub fn get_nostr_keys_for_pubkey(&self, pubkey: &PublicKey) -> Result<Keys, SecretsStoreError> {
         let hex_pubkey = pubkey.to_hex();
-        if cfg!(target_os = "android") {
+        if cfg!(target_os = "android") || cfg!(test) {
             let secrets = self.read_secrets_file()?;
             let obfuscated_key = secrets[&hex_pubkey.as_str()]
                 .as_str()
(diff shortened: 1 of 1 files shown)


src/whitenoise/chat_list.rs
```diff
@@ -0,0 +1,856 @@
...
+
+    /// When this group was created in our database (`DateTime<Utc>` for sorting consistency)
+    pub created_at: DateTime<Utc>,
+
+    /// Path to cached decrypted group image (Groups only, None for DMs)
+    pub group_image_path: Option<PathBuf>,
+
+    /// Profile picture URL of the other user (DMs only, None for Groups)
+    /// From the other participant's metadata.picture
+    pub group_image_url: Option<String>,
```
erskingardner, line 43: @jgmontoya there was some talk about fetching user metadata images and caching them locally. did we push that off for the future? I don't think we should hold up this PR but just wanted to check up on that.

src/whitenoise/database/aggregated_messages.rs
```diff
@@ -560,6 +560,89 @@ impl AggregatedMessage {
...
+                   SELECT id
+                   FROM aggregated_messages am2
+                   WHERE am2.mls_group_id = am1.mls_group_id
+                     AND am2.kind = 9
+                     AND am2.deletion_event_id IS NULL
+                   ORDER BY created_at DESC, id DESC
+                   LIMIT 1
+               )",
+            placeholders
+        );
```
dannym-arx, line 605: The correlated subquery executes once per row in am1, causing causing $O(n^2)$ time complexity, this is (mostly) fine because it doesn't get called that often, but if the user is part of many chats or they message a lot, it may be slower than it needs to be.
Why not use a CTE instead?

dannym-arx: something like this (please double check the query, did not test):

```sql
WITH ranked_messages AS (
    SELECT 
        mls_group_id,
        author,
        content,
        created_at,
        json_array_length(media_attachments) as media_count,
        ROW_NUMBER() OVER (
            PARTITION BY mls_group_id 
            ORDER BY created_at DESC, id DESC
        ) as rn
    FROM aggregated_messages
    WHERE kind = 9
      AND mls_group_id IN ({})
      AND deletion_event_id IS NULL
)
SELECT mls_group_id, author, content, created_at, media_count
FROM ranked_messages
WHERE rn = 1;
```

src/whitenoise/chat_list.rs
jgmontoya: yes, the flutter side already handles caching of images it gets by url on its own so there's really little benefit on changing that for the time being

src/whitenoise/database/aggregated_messages.rs
jgmontoya: With the existing indexes in place, the correlated query does an index seek with limit 1 per candidate row, which makes it O(log n) rather than O(n).

I tested both approaches with 50K messages across 50 groups, on my machine:
- Correlated subquery: ~15ms
- Window function CTE: ~24ms

I think we can revisit it in the future if it becomes a problem but for now I'd leave it as-is.

dannym-arx: ok then, sounds good! thanks for looking into it

PR #417, opened 2025-12-12
feat: add accounts_groups table for per-account group relationship tracking

mubarakcoded:
Add AccountGroup struct and database layer to track account-group relationships with user confirmation state (pending/accepted/declined). This enables auto-joining MLS groups on welcome messages while preserving UI control for accept/decline.

- Add migration 0018_accounts_groups.sql with indexes
- Add AccountGroup struct with CRUD operations and visibility helpers
- Integrate with process_welcome (creates pending record)
- Integrate with accept_welcome (sets user_confirmation = true)
- Integrate with decline_welcome (sets user_confirmation = false)4

## Summary by CodeRabbit

* **New Features**
  * Persistent account↔group membership with tri-state confirmation (pending/accepted/declined). Welcomes auto-finalize membership so recipients can participate immediately; non-critical provisioning (subscriptions, keys, images) runs in background.

* **Chores**
  * DB migration and indexes to store account↔group mappings; lazy migration creates missing mappings when listing/fetching groups. Account→group type is publicly exported.

* **Tests**
  * Added wait-for-welcome and account-group lifecycle/visibility tests; removed manual accept-invite test case.

### DIFF
--- a/src/whitenoise/database/accounts_groups.rs
+++ b/src/whitenoise/database/accounts_groups.rs
@@ -176,9 +176,7 @@ impl AccountGroup {
         user_confirmation: bool,
         database: &Database,
     ) -> Result<Self, sqlx::Error> {
-        let id = self.id.ok_or_else(|| {
-            sqlx::Error::Protocol("Cannot update unsaved AccountGroup".to_string())
-        })?;
+        let id = self.id.expect("AccountGroup must be persisted");
 
         let now_ms = Utc::now().timestamp_millis();
         let confirmation_int: i64 = if user_confirmation { 1 } else { 0 };


dannym-arx (changes requested):
- the two comments that coderabbit left are actually very important to fix
- the migration should mark all existing groups as accepted
- we already have group_information, why not ALTER TABLE that instead of creating a new accounts_groups?

src/whitenoise/accounts_groups.rs
```diff
@@ -0,0 +1,400 @@
...
+            .await?;
+        Ok(updated)
+    }
+
+    /// Removes this AccountGroup from the database.
+    pub async fn remove(&self, whitenoise: &Whitenoise) -> Result<(), WhitenoiseError> {
+        self.delete(&whitenoise.database).await?;
+        Ok(())
+    }
+}
```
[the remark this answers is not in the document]
mubarakcoded: These methods are defined in src/whitenoise/database/accounts_groups.rs (lines 218 and 246). Rust allows multiple impl blocks for the same struct across different files - the public API is in accounts_groups.rs while the database operations are in database/accounts_groups.rs. The code compiles and passes clippy.

db_migrations/0018_accounts_groups.sql
```diff
@@ -0,0 +1,22 @@
+-- Create accounts_groups table to track account-group relationships
```
dannym-arx, line 1: the migration should mark all existing groups as accepted

db_migrations/0018_accounts_groups.sql
```diff
@@ -0,0 +1,22 @@
...
+-- This is a join table between accounts and groups (via mls_group_id)
+-- user_confirmation tracks whether the user has accepted or declined a group invite:
+-- NULL = pending (auto-joined but awaiting user decision)
+-- 1 (true) = accepted
+-- 0 (false) = declined (hidden from UI)
+CREATE TABLE accounts_groups (
+    id INTEGER PRIMARY KEY AUTOINCREMENT,
```
dannym-arx, line 8: we already have group_information, why not `ALTER TABLE` that instead of creating a new accounts_groups?

mubarakcoded: We discussed this with @erskingardner and @jgmontoya, the decision was to keep accounts_groups separate from group_information because of scoping and separation of concern. 

accounts_groups specifically tracks the user confirmation state (pending/accepted/declined) for the auto-accept invite flow, which is a distinct concept from group metadata

db_migrations/0018_accounts_groups.sql
mubarakcoded: Good point. I will add a migration step to mark existing groups as accepted.

db_migrations/0018_accounts_groups.sql
dannym-arx: alright, sounds good :+1:

db_migrations/0018_accounts_groups.sql
```diff
@@ -0,0 +1,29 @@
...
+    account_pubkey TEXT NOT NULL,
+    mls_group_id BLOB NOT NULL,
+    user_confirmation INTEGER DEFAULT NULL,
+    created_at INTEGER NOT NULL,
+    updated_at INTEGER NOT NULL,
```
jgmontoya: I think on other tables we're using `DATETIME` instead of `INTEGER`, let's do the same to keep things consistent:


```
    created_at DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
    updated_at DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP
```

db_migrations/0018_accounts_groups.sql
```diff
@@ -0,0 +1,29 @@
...
+-- This is a join table between accounts and groups (via mls_group_id)
+-- user_confirmation tracks whether the user has accepted or declined a group invite:
+-- NULL = pending (auto-joined but awaiting user decision)
+-- 1 (true) = accepted
+-- 0 (false) = declined (hidden from UI)
+CREATE TABLE accounts_groups (
+    id INTEGER PRIMARY KEY AUTOINCREMENT,
+    account_pubkey TEXT NOT NULL,
+    mls_group_id BLOB NOT NULL,
+    user_confirmation INTEGER DEFAULT NULL,
```
jgmontoya: SQLite doesn't have explicit booleans but we can improve this by adding a check:

`user_confirmation INTEGER DEFAULT NULL CHECK (user_confirmation IN (0, 1))`

Also maybe we want to change the column name to make it more evident that its a boolean? Maybe something like has_user_confirmed, or is_user_confirmed or something like that? I'm not sure though

db_migrations/0018_accounts_groups.sql
```diff
@@ -0,0 +1,29 @@
...
+CREATE INDEX idx_accounts_groups_account ON accounts_groups(account_pubkey);
+
+-- Index for efficient lookups by group
+CREATE INDEX idx_accounts_groups_group ON accounts_groups(mls_group_id);
+
+-- Migrate existing groups from group_information as accepted (user_confirmation = 1)
+-- These were created before auto-accept flow, so users explicitly accepted them
+-- Using INSERT OR IGNORE to safely handle any potential duplicates
+INSERT OR IGNORE INTO accounts_groups (account_pubkey, mls_group_id, user_confirmation, created_at, updated_at)
+SELECT account_pubkey, mls_group_id, 1, created_at, updated_at
```
jgmontoya: I think we can use `TRUE` instead of `1` here, not 100% sure though, would need to check. If that works, it'd be better imo to make it more explicit

src/whitenoise/database/accounts_groups.rs
```diff
@@ -0,0 +1,510 @@
...
+use nostr_sdk::PublicKey;
+
+use super::{Database, utils::parse_timestamp};
+use crate::whitenoise::accounts_groups::AccountGroup;
+
+/// Internal database row representation for accounts_groups table
+#[derive(Debug, PartialEq, Eq, Clone, Hash)]
+struct AccountGroupRow {
+    id: i64,
+    account_pubkey: String,
```
jgmontoya: This should be `PublicKey`, not `String` (see for example `AccountRow` in `src/whitenoise/database/accounts.rs`). Parse it on the `from_row` instead of the `into_account_group`, then  replace `into_account_group` with `From`:

```rs
impl From<AccountGroupRow> for AccountGroup {
    fn from(row: AccountGroupRow) -> Self {
```

src/whitenoise/database/accounts_groups.rs
```diff
@@ -0,0 +1,510 @@
...
+    /// Finds all AccountGroups for a given account.
+    pub(crate) async fn find_all_for_account(
+        account_pubkey: &PublicKey,
+        database: &Database,
+    ) -> Result<Vec<Self>, sqlx::Error> {
+        let rows = sqlx::query_as::<_, AccountGroupRow>(
+            "SELECT id, account_pubkey, mls_group_id, user_confirmation, created_at, updated_at
+             FROM accounts_groups
+             WHERE account_pubkey = ?
+             ORDER BY created_at DESC",
```
jgmontoya: I don't think we should order by default. There's no guarantee that the caller needs the results ordered and ensuring the order has a performance cost

src/whitenoise/database/accounts_groups.rs
```diff
@@ -0,0 +1,510 @@
...
+    /// Visible means: user_confirmation is NULL (pending) or true (accepted).
+    /// Declined groups (user_confirmation = false) are hidden.
+    pub(crate) async fn find_visible_for_account(
+        account_pubkey: &PublicKey,
+        database: &Database,
+    ) -> Result<Vec<Self>, sqlx::Error> {
+        let rows = sqlx::query_as::<_, AccountGroupRow>(
+            "SELECT id, account_pubkey, mls_group_id, user_confirmation, created_at, updated_at
+             FROM accounts_groups
+             WHERE account_pubkey = ? AND (user_confirmation IS NULL OR user_confirmation = 1)
```
jgmontoya: if possible I'd use `TRUE` instead of `1` (please verify that it works though 🙏)

src/whitenoise/database/accounts_groups.rs
```diff
@@ -0,0 +1,510 @@
...
+             ORDER BY created_at DESC",
```
jgmontoya: I'd remove the `ORDER BY` here as well

src/whitenoise/database/accounts_groups.rs
```diff
@@ -0,0 +1,510 @@
...
+    /// Pending means: user_confirmation is NULL.
+    pub(crate) async fn find_pending_for_account(
+        account_pubkey: &PublicKey,
+        database: &Database,
+    ) -> Result<Vec<Self>, sqlx::Error> {
+        let rows = sqlx::query_as::<_, AccountGroupRow>(
+            "SELECT id, account_pubkey, mls_group_id, user_confirmation, created_at, updated_at
+             FROM accounts_groups
+             WHERE account_pubkey = ? AND user_confirmation IS NULL
+             ORDER BY created_at DESC",
```
jgmontoya: and here

src/whitenoise/database/accounts_groups.rs
```diff
@@ -0,0 +1,510 @@
...
+
+        rows.into_iter()
+            .map(|r| r.into_account_group())
+            .collect::<Result<Vec<_>, _>>()
+    }
+
+    /// Updates the user_confirmation status for this AccountGroup.
+    pub(crate) async fn update_user_confirmation(
+        &self,
+        user_confirmation: Option<bool>,
```
jgmontoya: Why does this need to be an `Option<...>` ? I think it should be just `bool`, no?

src/whitenoise/database/accounts_groups.rs
```diff
@@ -0,0 +1,510 @@
...
+        sqlx::query("DELETE FROM accounts_groups WHERE id = ?")
+            .bind(id)
+            .execute(&database.pool)
+            .await?;
+
+        Ok(())
+    }
+
+    /// Inserts a new AccountGroup with user_confirmation = NULL (pending).
+    async fn insert_new(
```
jgmontoya: I think we've named this `create` on other entities. It'd be best to keep the same kind of naming

src/whitenoise/accounts_groups.rs
```diff
@@ -0,0 +1,400 @@
...
+        AccountGroup::get(self, &account.pubkey, mls_group_id).await
+    }
+
+    /// Gets all AccountGroups for the given account.
+    pub async fn get_all_account_groups(
+        &self,
+        account: &Account,
+    ) -> Result<Vec<AccountGroup>, WhitenoiseError> {
+        AccountGroup::all_for_account(self, &account.pubkey).await
+    }
```
jgmontoya: What's this method needed for?

src/whitenoise/accounts_groups.rs
```diff
@@ -0,0 +1,400 @@
...
+
+    /// Gets an AccountGroup for the given account and MLS group, if it exists.
+    pub async fn get_account_group(
+        &self,
+        account: &Account,
+        mls_group_id: &GroupId,
+    ) -> Result<Option<AccountGroup>, WhitenoiseError> {
+        AccountGroup::get(self, &account.pubkey, mls_group_id).await
+    }
```
jgmontoya: I think only one of these is really needed

src/whitenoise/accounts_groups.rs
```diff
@@ -0,0 +1,400 @@
...
+        whitenoise: &Whitenoise,
+    ) -> Result<Self, WhitenoiseError> {
+        let updated = self
+            .update_user_confirmation(None, &whitenoise.database)
+            .await?;
+        Ok(updated)
+    }
+
+    /// Removes this AccountGroup from the database.
+    pub async fn remove(&self, whitenoise: &Whitenoise) -> Result<(), WhitenoiseError> {
```
jgmontoya: I think I'd name this `delete` as well

src/whitenoise/accounts_groups.rs
```diff
@@ -0,0 +1,400 @@
...
+    /// Resets confirmation to pending state (None).
+    pub async fn reset_confirmation(
+        &self,
+        whitenoise: &Whitenoise,
+    ) -> Result<Self, WhitenoiseError> {
+        let updated = self
+            .update_user_confirmation(None, &whitenoise.database)
+            .await?;
+        Ok(updated)
+    }
```
jgmontoya: Do we actually want to support this? I'm not sure I understand how this would work tbh

src/whitenoise/accounts_groups.rs
```diff
@@ -0,0 +1,400 @@
...
+
+    /// Gets all AccountGroups for the given account.
+    pub async fn all_for_account(
+        whitenoise: &Whitenoise,
+        account_pubkey: &PublicKey,
+    ) -> Result<Vec<Self>, WhitenoiseError> {
+        let groups = Self::find_all_for_account(account_pubkey, &whitenoise.database).await?;
+        Ok(groups)
+    }
```
jgmontoya: what do we need this one for? I mean it seems like a reasonable method to have but now I can't think of a case where we also need the declined ones

src/whitenoise/accounts_groups.rs
```diff
@@ -0,0 +1,400 @@
...
+    pub async fn get(
+        whitenoise: &Whitenoise,
+        account_pubkey: &PublicKey,
+        mls_group_id: &GroupId,
+    ) -> Result<Option<Self>, WhitenoiseError> {
+        let account_group =
+            Self::find_by_account_and_group(account_pubkey, mls_group_id, &whitenoise.database)
+                .await?;
+        Ok(account_group)
+    }
```
jgmontoya, line 73: Do we need both of these as public methods?

src/whitenoise/event_processor/event_handlers/handle_giftwrap.rs
jgmontoya, line 1: Let's be mindful of what we put on the event handlers, as the event processor runs serially that means we have to be careful and make sure handlers always execute fast. For that reason, I think the changes here could be done in a background task. What do you think?

jgmontoya (changes requested):
I might be missing something but it seems to me like this is not really doing the thing, shouldn't this be calling the accept_welcome method from the event handler?

Also, can you please double check that the new public methods are indeed needed? I think in general is good to try to keep the public API from growing too much as that can become a maintenance burden.

db_migrations/0018_accounts_groups.sql
jgmontoya: > We discussed this with @erskingardner and @jgmontoya, the decision was to keep accounts_groups separate from group_information because of scoping and separation of concern.
> 
> accounts_groups specifically tracks the user confirmation state (pending/accepted/declined) for the auto-accept invite flow, which is a distinct concept from group metadata

Just a clarification:

`group_information` is meant to be for things that are intrinsic to the group (regardless of user/account/etc..)

`accounts_groups` is a new table for things that specifically relate an account to a group. As of this PR it only has one "thing": `user_confirmation` which tracks the user confirmation state, but this table would also be suitable for other stuff as long as it's (account, group)-scoped. For example let's say in the future we implement an "archived" feature that allows us to hide some groups and display them in a different part. This table would be the natural place to put such information.

db_migrations/0018_accounts_groups.sql
```diff
@@ -0,0 +1,27 @@
...
+-- NULL = pending (auto-joined but awaiting user decision)
+-- 1 (true) = accepted
+-- 0 (false) = declined (hidden from UI)
+CREATE TABLE accounts_groups (
+    id INTEGER PRIMARY KEY AUTOINCREMENT,
+    account_pubkey TEXT NOT NULL,
+    mls_group_id BLOB NOT NULL,
+    user_confirmation INTEGER DEFAULT NULL CHECK (user_confirmation IS NULL OR user_confirmation IN (0, 1)),
+    created_at DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    updated_at DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
```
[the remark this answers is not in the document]
mubarakcoded: The trigger is not needed and would cause recursion issues. The DEFAULT CURRENT_TIMESTAMP handles the initial insert, and the application code handles updates.

src/whitenoise/accounts_groups.rs
mubarakcoded: Yes, both are needed. get_or_create is used externally in 3 places for auto-join flows. get is only used internally now, but keeping it public provides API symmetry and allows future callers to check existence without creating. This matches similar pattern in the codebase. I have seen it in GroupInformation, with different naming though.

db_migrations/0018_accounts_groups.sql
mubarakcoded: I added the CHECK constraint. Regarding renaming to has_user_confirmed, the current name user_confirmation is a tri-state (NULL/0/1), not a simple boolean. has_user_confirmed implies boolean. Should we keep user_confirmation or rename?

db_migrations/0018_accounts_groups.sql
mubarakcoded: I have removed the INSERT statement because group_information doesn't have account_pubkey, and SQLite does support TRUE/FALSE literals mapped to 1/0

src/whitenoise/accounts_groups.rs
mubarakcoded: Yes. It has been removed.

db_migrations/0018_accounts_groups.sql
jgmontoya: I don't feel strongly about renaming, let's keep whichever name you feel is more readable 👍

db_migrations/0018_accounts_groups.sql
jgmontoya: > I have removed the INSERT statement because group_information doesn't have account_pubkey

Oh that's right, I missed that on review 🤦

It's gonna be a bit challenging to do the initial data migration to populate the accounts_groups

src/whitenoise/event_processor/event_handlers/handle_giftwrap.rs
erskingardner: Looks like you've addressed this one @mubarakcoded - the method here is very long but it's very easy to follow so I actually wouldn't suggest breaking it out into separate smaller methods.

jgmontoya: > so I actually wouldn't suggest breaking it out into separate smaller methods

Image

You know I would

db_migrations/0018_accounts_groups.sql
```diff
@@ -0,0 +1,27 @@
...
+-- Index for efficient lookups by account
+CREATE INDEX idx_accounts_groups_account ON accounts_groups(account_pubkey);
+
+-- Index for efficient lookups by group
+CREATE INDEX idx_accounts_groups_group ON accounts_groups(mls_group_id);
+
+-- Note: We cannot migrate existing groups from group_information because that table
+-- doesn't track account_pubkey (it's global, not per-account). Existing groups will
+-- get accounts_groups records created when users interact with them through the
+-- normal flow (process_welcome, accept_welcome, etc.).
```
jgmontoya: This seems problematic. It would make it so that users will have to manually re-accept every group they're in. Maybe we should add some little infrastructure to do "code-based migrations" at startup.

src/integration_tests/benchmarks/scenarios/message_aggregation.rs
```diff
@@ -120,18 +119,61 @@ impl BenchmarkScenario for MessageAggregationBenchmark {
...
+                    } else {
+                        Err(WhitenoiseError::Other(anyhow::anyhow!(
+                            "charlie has not yet received the group"
+                        )))
+                    }
+                }
+            },
+            "charlie welcome auto-finalization",
+        )
+        .await?;
```
jgmontoya: This kind of logic should be encapsulated into a TestCase rather than be directly on the scenario

src/whitenoise/database/accounts_groups.rs
```diff
@@ -0,0 +1,462 @@
...
+        .bind(now_ms)
+        .bind(id)
+        .fetch_one(&database.pool)
+        .await?;
+
+        Ok(row.into())
+    }
+
+    /// Deletes this AccountGroup from the database.
+    #[allow(dead_code)]
```
jgmontoya: If this is dead code perhaps we should remove the method for now until we actually use it somewhere?

src/whitenoise/event_processor/event_handlers/handle_giftwrap.rs
```diff
@@ -69,28 +78,202 @@ impl Whitenoise {
...
-                    account,
-                    &key_package_event_id,
-                    false, // For now we don't want to delete the key packages from MLS storage
+        // Clone data needed for background task
+        let account_clone = account.clone();
+        let data_dir = self.config.data_dir.clone();
+
+        // Spawn background task for remaining operations (DB writes, network calls)
+        // All operations are idempotent and failures are logged but don't stop other operations
+        tokio::spawn(async move {
```
jgmontoya: Yeah I would definitely extract to a private method here, something like `background_finalize_welcome` or something

src/whitenoise/event_processor/event_handlers/handle_giftwrap.rs
```diff
@@ -69,28 +78,202 @@ impl Whitenoise {
...
-                .await?;
+                .await
+            {
+                tracing::error!(
+                    target: "whitenoise::event_processor::process_welcome::background",
+                    "Failed to setup group subscriptions for account {}: {}",
+                    account_clone.pubkey.to_hex(),
+                    e
+                );
+            }
```
jgmontoya: Then probably each of these (or at least 3 and 4 which are the most complex ones) I would extract to their own private method as well

src/whitenoise/event_processor/event_handlers/handle_giftwrap.rs
jgmontoya: It seems like the different things being done here are also quite independent of each other, so maybe it makes sense to allow them to execute concurrently so everything happens faster

src/whitenoise/welcomes.rs
jgmontoya, line 1: What happens when you try to call mdk.accept_welcome on a welcome that was already accepted?

Because the event processor will call mdk.accept_welcome, and then the accept_welcome method here is calling it again.

It seems to me like we don't really need the accept_welcome and decline_welcome public methods anymore as that would be automatically handled by the event processor, no? The flutter app can call accept_account_group directly

src/whitenoise/event_processor/event_handlers/handle_giftwrap.rs
mubarakcoded: > Yeah I would definitely extract to a private method here, something like `background_finalize_welcome` or something

Okay. On this which approach should I take? @erskingardner want us to keep the long function as-is, and @jgmontoya suggest breaking it into separate smaller methods.

db_migrations/0018_accounts_groups.sql
erskingardner: yeah, I missed this. I agree. at a push, i think it's probably fine to just auto-accept any pending welcomes that users might have on first startup and also create the necessary new records for all current groups.

src/whitenoise/welcomes.rs
mubarakcoded: You're right, I will simplify it and remove the redundancy.

db_migrations/0018_accounts_groups.sql
mubarakcoded: @erskingardner @jgmontoya Instead of a startup migration, I suggest lazy migration in groups(), if a group exists in MLS but has no AccountGroup record, we auto-create it as accepted. This is simpler (no version tracking), self-healing, and handles edge cases like accounts created mid-migration. Does this resonate with you?

src/whitenoise/database/accounts_groups.rs
mubarakcoded: Yeah. Will be removed.

src/integration_tests/benchmarks/scenarios/message_aggregation.rs
mubarakcoded: You're right! I will extract the retry logic into a reusable TestCase

db_migrations/0018_accounts_groups.sql
jgmontoya: Ok I'm liking this idea, but instead of just checking that the group exists in MLS I think you should check that it's also active (group.state = group_types::GroupState::Active;)

src/whitenoise/event_processor/event_handlers/handle_giftwrap.rs
jgmontoya: image

erskingardner: as you wish :)

db_migrations/0018_accounts_groups.sql
erskingardner: yeah, that sounds good.

src/whitenoise/welcomes.rs
mubarakcoded: Removed redundant mdk.accept_welcome call, calling twice would fail. We still need accept_welcome/decline_welcome methods for UI setup (GroupInformation, subscriptions, AccountGroup state).
MLS acceptance is already done by event processor.

src/integration_tests/test_cases/shared/wait_for_welcome.rs
```diff
@@ -0,0 +1,83 @@
...
+            tracing::info!(
+                "Waiting for {} to receive welcome for group '{}'...",
+                account_name,
+                self.group_name
+            );
+
+            let wn = context.whitenoise;
+            let acc = account.clone();
+            let gid = group_id.clone();
+            let name = account_name.clone();
```
jgmontoya: These are not needed here

src/integration_tests/test_cases/shared/wait_for_welcome.rs
```diff
@@ -0,0 +1,83 @@
...
+                        let groups = wn.groups(&acc, true).await?;
+                        if groups.iter().any(|g| g.mls_group_id == gid) {
+                            Ok(())
+                        } else {
+                            Err(WhitenoiseError::Other(anyhow::anyhow!(
+                                "{} has not yet received the group",
+                                name
+                            )))
+                        }
+                    }
```
jgmontoya, line 61: Suggestion:

```rs
async fn run(&self, context: &mut ScenarioContext) -> Result<(), WhitenoiseError> {
    let group = context.get_group(&self.group_name)?;
    let group_id = group.mls_group_id.clone();
    let wn = context.whitenoise;

    for account_name in &self.account_names {
        let account = context.get_account(account_name)?;
        tracing::info!(
            "Waiting for {} to receive welcome for group '{}'...",
            account_name,
            self.group_name
        );

        retry_default(
            || {
                let account = account.clone();
                let group_id = group_id.clone();
                let account_name = account_name.clone();
                async move {
                    let groups = wn.groups(&account, true).await?;
                    if groups.iter().any(|g| g.mls_group_id == group_id) {
                        Ok(())
                    } else {
                        Err(WhitenoiseError::Other(anyhow::anyhow!(
                            "{} has not yet processed the welcome for the group",
                            account_name
                        )))
                    }
                }
```

src/integration_tests/test_cases/shared/wait_for_welcome.rs
```diff
@@ -0,0 +1,83 @@
...
+
+        for account_name in &self.account_names {
+            let account = context.get_account(account_name)?.clone();
+            tracing::info!(
+                "Waiting for {} to receive welcome for group '{}'...",
+                account_name,
+                self.group_name
+            );
+
+            let wn = context.whitenoise;
```
jgmontoya: this should be outside of the loop

src/integration_tests/test_cases/shared/wait_for_welcome.rs
```diff
@@ -0,0 +1,83 @@
...
+}
+
+#[async_trait]
+impl TestCase for WaitForWelcomeTestCase {
+    async fn run(&self, context: &mut ScenarioContext) -> Result<(), WhitenoiseError> {
+        let group = context.get_group(&self.group_name)?;
+        let group_id = group.mls_group_id.clone();
+
+        for account_name in &self.account_names {
+            let account = context.get_account(account_name)?.clone();
```
jgmontoya: this clone ends up being not needed

src/whitenoise/event_processor/event_handlers/handle_giftwrap.rs
```diff
@@ -69,28 +80,286 @@ impl Whitenoise {
...
+                tracing::debug!(
+                    target: "whitenoise::event_processor::process_welcome::background",
+                    "Published new key package"
+                );
+            }
+            false => {
+                tracing::debug!(
+                    target: "whitenoise::event_processor::process_welcome::background",
+                    "Key package already deleted, skipping publish"
+                );
```
jgmontoya: Using `match` with boolean values is unusual and harder to read, please change this to `if`/`else`

src/whitenoise/event_processor/event_handlers/handle_giftwrap.rs
```diff
@@ -69,28 +80,286 @@ impl Whitenoise {
...
+        tokio::spawn(async move {
+            Self::background_finalize_welcome(
+                account_clone,
+                group_id,
+                group_name,
+                key_package_event_id,
+                data_dir,
+            )
+            .await;
+        });
```
jgmontoya: This async move { ... .await } is not needed. It's cleaner to just pass the Self::background_finalize_welcome( (without the await):

```rs
tokio::spawn(Self::background_finalize_welcome(
    account.clone(),  // account_clone is a weird name, it's better to just call clone here
    group_id,
    group_name,
    key_package_event_id,
    data_dir,
));
```

src/whitenoise/groups.rs
```diff
@@ -284,12 +285,35 @@ impl Whitenoise {
...
+                    tracing::debug!(
+                        target: "whitenoise::groups",
+                        "Lazy migration: created and accepted AccountGroup for existing group {}",
+                        hex::encode(group.mls_group_id.as_slice())
+                    );
+                }
+            }
+        }
+
+        Ok(groups)
```
jgmontoya, line 303: This could be extracted to a private method

src/whitenoise/groups.rs
jgmontoya, line 1: I think the singular `group` method should also perform/check the lazy migration. What do you think?

src/whitenoise/welcomes.rs
jgmontoya, line 1: I'm not sure we need the `accept_welcome` and `decline_welcome` methods at all. Why not have the flutter app directly call accept on account group?

src/whitenoise/database/accounts_groups.rs
```diff
@@ -0,0 +1,423 @@
...
+        )
+        .bind(account_pubkey.to_hex())
+        .bind(mls_group_id.as_slice())
+        .fetch_optional(&database.pool)
+        .await?;
+
+        match row {
+            Some(r) => Ok(Some(r.into())),
+            None => Ok(None),
+        }
```
jgmontoya: This can be simplified to: `Ok(row.map(Into::into))`

src/whitenoise/database/accounts_groups.rs
```diff
@@ -0,0 +1,423 @@
...
+    }
+
+    /// Updates the user_confirmation status for this AccountGroup.
+    pub(crate) async fn update_user_confirmation(
+        &self,
+        user_confirmation: bool,
+        database: &Database,
+    ) -> Result<Self, sqlx::Error> {
+        let id = self.id.ok_or_else(|| {
+            sqlx::Error::Protocol("Cannot update unsaved AccountGroup".to_string())
```
jgmontoya: `sqlx::Error::Protocol` is kind of a weird error but I'm not sure which one would be better here, what do you think?

src/whitenoise/database/accounts_groups.rs
```diff
@@ -0,0 +1,423 @@
...
+        let rows = sqlx::query_as::<_, AccountGroupRow>(
+            "SELECT id, account_pubkey, mls_group_id, user_confirmation, created_at, updated_at
+             FROM accounts_groups
+             WHERE account_pubkey = ? AND (user_confirmation IS NULL OR user_confirmation = 1)",
+        )
+        .bind(account_pubkey.to_hex())
+        .fetch_all(&database.pool)
+        .await?;
+
+        Ok(rows.into_iter().map(|r| r.into()).collect())
```
jgmontoya: Not that it matters, but this could also be `rows.into_iter().map(Into::into).collect()`

src/whitenoise/database/accounts_groups.rs
```diff
@@ -0,0 +1,423 @@
...
+        let rows = sqlx::query_as::<_, AccountGroupRow>(
+            "SELECT id, account_pubkey, mls_group_id, user_confirmation, created_at, updated_at
+             FROM accounts_groups
+             WHERE account_pubkey = ? AND user_confirmation IS NULL",
+        )
+        .bind(account_pubkey.to_hex())
+        .fetch_all(&database.pool)
+        .await?;
+
+        Ok(rows.into_iter().map(|r| r.into()).collect())
```
jgmontoya: same here with `rows.into_iter().map(Into::into).collect()`

jgmontoya (changes requested):
Some comments!

src/whitenoise/database/accounts_groups.rs
mubarakcoded: I think the cleanest approach is to return a WhitenoiseError::InvalidState("Cannot update unsaved AccountGroup") instead of forcing an sqlx error. This requires changing the return type but is more semantically correct. What do you think?

src/whitenoise/database/accounts_groups.rs
```diff
@@ -0,0 +1,420 @@
...
+        let user_confirmation_int: Option<i64> = row.try_get("user_confirmation")?;
+
+        // Parse pubkey from hex string
+        let account_pubkey =
+            PublicKey::parse(&account_pubkey_str).map_err(|e| sqlx::Error::ColumnDecode {
+                index: "account_pubkey".to_string(),
+                source: Box::new(e),
+            })?;
+
+        let mls_group_id = GroupId::from_slice(&mls_group_id_bytes);
```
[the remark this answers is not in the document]
mubarakcoded: GroupId::from_slice is used consistently across all database row parsers without validation in the codebase and data comes from our own database schema where we control the BLOB format

src/whitenoise/database/accounts_groups.rs
jgmontoya: Hmm given that this seems rather unreachable (I don't think we ever expect the crate user to manually create an unpersisted account_group, maybe we can just use .expect (as we do for example in `src/whitenoise/database/users.rs`, lines 387 and 441)

Something like this: 

```rs
let id = self.id.expect("AccountGroup must be persisted")
```

PR #418, opened 2025-12-19
Amber integration

dannym-arx:
## Summary by CodeRabbit

* **New Features**
  * External-signer support: register/get/remove signers, login with external signer, and signer-aware publish/delete and subscription flows (including no-signer batched subscription paths).

* **Bug Fixes / Behavior**
  * Login/logout now migrate Local↔External correctly, honor account type for relay/subscription setup, decrypt with appropriate signer, and handle key cleanup with nuanced error handling.

* **Chores**
  * DB migration adds account_type column with allowed values.

* **Tests**
  * Expanded tests for external-signer flows, migrations, relays, publishing/deletion, and logout.

### DIFF
--- a/src/whitenoise/accounts.rs
+++ b/src/whitenoise/accounts.rs
@@ -3053,6 +3053,13 @@ mod tests {
     async fn test_publish_relay_lists_with_signer_publishes_when_flags_true() {
         let (whitenoise, _data_temp, _logs_temp) = create_mock_whitenoise().await;
         let keys = Keys::generate();
+        let pubkey = keys.public_key();
+
+        // Create an external account first - required for event tracking
+        let _account = whitenoise
+            .login_with_external_signer_for_test(pubkey)
+            .await
+            .unwrap();
 
         // Load default relays to have valid relay URLs
         let default_relays = whitenoise.load_default_relays().await.unwrap();


src/nostr_manager/subscriptions.rs
```diff
@@ -44,15 +44,49 @@ impl NostrManager {
 
         // 2. Set signer once and process all relays in parallel
         self.with_signer(signer, || async {
-            let futures = relay_user_map.into_iter().map(|(relay_url, users)| async move {
+            self.setup_batched_relay_subscriptions_inner(relay_user_map, since)
+                .await
```
jgmontoya, line 48: Should we propagate here if there's an error?

src/whitenoise/accounts.rs
```diff
@@ -631,6 +875,77 @@ impl Whitenoise {
...
+                .await?;
+            Ok(default_relays.to_vec())
+        } else {
+            // Found existing relay lists - use them
+            user.add_relays(&fetched_relays, relay_type, &self.database)
+                .await?;
+            Ok(fetched_relays)
+        }
+    }
+
```
jgmontoya, line 1035: How do we get the user to publish the relay lists if needed? Does that need to be controlled from the flutter side?

src/whitenoise/accounts.rs
```diff
@@ -371,7 +565,17 @@ impl Whitenoise {
...
+        if account.has_local_key() {
+            self.secrets_store.remove_private_key_for_pubkey(pubkey)?;
+        } else if let Err(e) = self.secrets_store.remove_private_key_for_pubkey(pubkey) {
+            tracing::debug!(
+                target: "whitenoise::logout",
+                "No local key to remove for external account {}: {}",
+                pubkey,
+                e
+            );
+        }
```
jgmontoya: Other option:

```rs
let result = self.secrets_store.remove_private_key_for_pubkey(pubkey);
match (account.has_local_key(), result) {
    (true, Err(e)) => return Err(e.into()),  // Local account MUST have key
    (false, Err(e)) => tracing::debug!("Expected - no key for external account: {}", e),
    _ => {}
}
```

src/nostr_manager/mod.rs
```diff
@@ -288,6 +288,35 @@ impl NostrManager {
         .await
     }
 
+    /// Updates account subscriptions without a signer (for external signer accounts).
+    ///
+    /// This version is used for accounts that use external signers (like Amber) where
+    /// the private key is not available locally.
+    pub(crate) async fn update_account_subscriptions(
```
jgmontoya, line 295: This got me thinking about whether we actually really need to do the whole client.set_signer thing for account subscriptions. Maybe there's a chance to simplify things such that we could reuse the subscriptions setup for both external and local signers.

After quickly looking at the nostr-sdk source I think subscriptions don't use the signer at all. The subscribe_with_id_to and related methods just send REQ messages to relays and there seems to be no signing involved. My understanding now is that the signer is only used for signing events, decrypting content, and NIP-42 auth.

The _with_signer subscription variants can likely be removed entirely.

Also, if NIP-42 relay auth becomes needed later, the current set→operate→unset pattern wouldn't work anyway because I think AUTH challenges would arrive asynchronously after the signer is already unset so we would need a different approach anyway.

src/nostr_manager/subscriptions.rs
dannym-arx: perhaps, yeah

src/nostr_manager/mod.rs
dannym-arx: I think that change should be a different PR, right? or would you prefer if I did it in this one?

src/whitenoise/accounts.rs
dannym-arx: I could build it either way, I think it's easier to do it on the flutter side

src/nostr_manager/mod.rs
jgmontoya: Yes we can tackle it in a different PR

src/whitenoise/accounts.rs
```diff
@@ -337,11 +422,205 @@ impl Whitenoise {
...
+    /// 2. Sets up relays (fetches existing from network or uses defaults)
+    /// 3. Registers the external signer for ongoing use (e.g., giftwrap decryption)
+    /// 4. Publishes relay lists if using defaults (using the signer)
+    /// 5. Publishes the MLS key package (using the signer)
+    ///
+    /// # Arguments
+    ///
+    /// * `pubkey` - The user's public key obtained from the external signer.
+    /// * `signer` - The external signer to use for signing operations. Must implement `Clone`.
+    pub async fn login_with_external_signer(
```
jgmontoya, line 440: I think this method is too long, could you extract some private methods with properly descriptive names from it to make it more readable / easier to follow?

dannym-arx: yep, will do

db_migrations/0023_add_account_type.sql
```diff
@@ -0,0 +1,8 @@
+-- Add account_type column to support external signers (NIP-55)
+-- account_type values:
+--   'local' - Private key is stored locally in secrets store
+--   'external' - Account uses external signer (e.g., Amber via NIP-55)
+
+ALTER TABLE accounts
```
erskingardner, line 6: @mubarakcoded has a 023 migration in his MDK branch. one of you will have to change.

jgmontoya: image

PR #419, opened 2025-12-22
feat: check for app updates

Quwaysim:
Addresses [#900](https://github.com/marmot-protocol/whitenoise/issues/900)

### New module: src/whitenoise/app_update.rs
- Queries `wss://relay.zapstore.dev` for Kind 1063 events from the Whitenoise pubkey.
- Filters events by created_at timestamp to get the latest release.
- Extracts the version tag from the event and compares against the current app version
- Returns AppUpdateInfo with the latest version and whether an update is available or not.

## Summary by CodeRabbit

* **New Features**
  * App update check added: queries a remote store for latest version and reports if an update is available.
  * Public configuration and result types exposed with a default configuration and an async API for programmatic checks.
  * New publicly exported module for app-update functionality.

* **Tests**
  * Unit tests for version parsing, comparison, and formatting.
  * New integration scenario and test case covering update-check flows (update available vs. no update).

### DIFF
--- a/src/whitenoise/app_update.rs
+++ b/src/whitenoise/app_update.rs
@@ -45,6 +45,7 @@ impl FromStr for Version {
     type Err = &'static str;
 
     fn from_str(s: &str) -> std::result::Result<Self, Self::Err> {
+        let s = s.trim();
         let parts: Vec<&str> = s.split('.').collect();
         if parts.len() != 3 {
             return Err("Invalid version format (expected x.y.z)");
@@ -90,6 +91,9 @@ fn extract_version_from_event(event: &Event) -> Option<String> {
             slice.get(1).and_then(|s| {
                 let value = s.as_str();
                 let (_, version) = value.split_once('@')?;
+                if version.is_empty() {
+                    return None;
+                }
                 Some(version.to_string())
             })
         } else {
@@ -251,6 +255,15 @@ mod tests {
         assert!(result.is_err());
     }
 
+    #[test]
+    fn test_version_parse_whitespace() {
+        // Test parsing with leading/trailing whitespace
+        let version = Version::from_str("  1.2.3  ").unwrap();
+        assert_eq!(version.major, 1);
+        assert_eq!(version.minor, 2);
+        assert_eq!(version.patch, 3);
+    }
+
     /// Version Comparison Tests (Ord and PartialOrd implementations)
 
     #[test]
@@ -382,5 +395,29 @@ mod tests {
 
         let version = extract_version_from_event(&event);
         assert!(version.is_none());
+    }
+
+    #[test]
+    fn test_extract_version_invalid_tags() {
+        let keys = Keys::generate();
+        let invalid_cases = vec![
+            vec![],
+            vec!["".to_string()],
+            vec!["invalid_format".to_string()],
+            vec!["app@".to_string()],
+        ];
+
+        for (i, tags) in invalid_cases.into_iter().enumerate() {
+            let event = EventBuilder::new(Kind::Custom(30063), "test content")
+                .tag(Tag::custom(TagKind::Custom("d".into()), tags))
+                .sign_with_keys(&keys)
+                .unwrap();
+
+            assert!(
+                extract_version_from_event(&event).is_none(),
+                "Failed at case index {}",
+                i
+            );
+        }
     }
 }


src/whitenoise/app_update.rs
```diff
@@ -0,0 +1,415 @@
+use nostr_sdk::prelude::*;
+use std::str::FromStr;
+use std::time::Duration;
+
+use crate::whitenoise::error::{Result, WhitenoiseError};
+
+const ZAPSTORE_RELAY_URL: &str = "wss://relay.zapstore.dev";
```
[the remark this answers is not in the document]
Quwaysim: No, the events needed are only published on that relay. We'll add fallback/replace this later when we publish our app update events @coderabbitai

src/whitenoise/app_update.rs
```diff
@@ -0,0 +1,531 @@
...
+
+    let pubkey = PublicKey::from_hex(WHITE_NOISE_PUBKEY)
+        .map_err(|e| WhitenoiseError::Other(anyhow::anyhow!(e)))?;
+
+    let filter = Filter::new().author(pubkey).kind(Kind::Custom(1063));
+
+    let events = client
+        .fetch_events_from([relay_url.clone()], filter, TIMEOUT)
+        .await
+        .map_err(|e| WhitenoiseError::Other(anyhow::anyhow!(e)))?;
```
jgmontoya: If we publish other apps using the same pubkey this will grab those events as well. We should filter by app to make sure we're only grabbing the events we care about (for Whitenoise, content starts with "org.parres.whitenoise")

src/whitenoise/app_update.rs
```diff
@@ -0,0 +1,531 @@
+use nostr_sdk::prelude::*;
+use std::str::FromStr;
+use std::time::Duration;
+
+use crate::whitenoise::error::{Result, WhitenoiseError};
+
+const ZAPSTORE_RELAY_URL: &str = "wss://relay.zapstore.dev";
+const WHITE_NOISE_PUBKEY: &str = "75d737c3472471029c44876b330d2284288a42779b591a2ed4daa1c6c07efaf7";
```
jgmontoya: I don't like the way we're hardcoding these values. It makes this feature only work for Whitenoise, however this crate could be used by other mobile apps, desktop apps, etc... which could also be published under different identifiers/pubkeys.

Let's make the function generic by accepting a configuration as a parameter:

```rs
/// Configuration for checking app updates from Zapstore
pub struct AppUpdateConfig {
    /// The Zapstore relay URL (defaults to wss://relay.zapstore.dev)
    pub relay_url: String,
    /// The pubkey that publishes app releases
    pub publisher_pubkey: String,
    /// The app identifier (e.g., "org.parres.whitenoise")
    pub app_identifier: String,
}

impl Default for AppUpdateConfig {
    fn default() -> Self {
        Self {
            relay_url: "wss://relay.zapstore.dev".to_string(),
            publisher_pubkey: "75d737c3472471029c44876b330d2284288a42779b591a2ed4daa1c6c07efaf7".to_string(),
            app_identifier: "org.parres.whitenoise".to_string(),
        }
    }
}

pub async fn check_for_app_update(
    current_version: &str,
    config: &AppUpdateConfig,
) -> Result<AppUpdateInfo> {
    // Use config.relay_url, config.publisher_pubkey
    // Filter events by config.app_identifier
}
```

I think this will also make it easier to test as we can point to our own relay, use a new set of keys to publish whatever events we want to make sure the core logic is correct.

Also, the Zapstore events include architecture info (`["f", "android-arm64-v8a"]`), should we consider it to filter relevant updates?

src/whitenoise/app_update.rs
```diff
@@ -0,0 +1,531 @@
...
+    pub version: String,
+    pub update_available: bool,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+struct Version {
+    major: u32,
+    minor: u32,
+    patch: u32,
+}
```
jgmontoya, line 42: Version only handles strict `x.y.z` format, let's make sure / validate maybe with Jeff that we don't need to support things like:

- Pre-release: `1.0.0-alpha`, `1.0.0-rc.1`
- Build metadata: `1.0.0+build.123`

There are some options here, ranging from fully supporting semver to leaving it as is and committing to strict `x.y.z` version format. Perhaps just stripping suffixes could be enough.

src/whitenoise/app_update.rs
```diff
@@ -0,0 +1,531 @@
...
+    }
+
+    #[test]
+    fn test_whitenoise_pubkey_is_valid_hex() {
+        // Test that the hardcoded pubkey is valid 64-character hex
+        assert_eq!(WHITE_NOISE_PUBKEY.len(), 64);
+
+        // Verify all characters are valid hex digits
+        assert!(WHITE_NOISE_PUBKEY.chars().all(|c| c.is_ascii_hexdigit()));
+    }
```
jgmontoya: This test only validates that its hex, but it can be better and validate that we can actually parse it as a pubkey:

```rs
PublicKey::from_hex(&config.publisher_pubkey).expect("should be valid public key");
```

src/whitenoise/app_update.rs
```diff
@@ -0,0 +1,531 @@
...
+
+const ZAPSTORE_RELAY_URL: &str = "wss://relay.zapstore.dev";
+const WHITE_NOISE_PUBKEY: &str = "75d737c3472471029c44876b330d2284288a42779b591a2ed4daa1c6c07efaf7";
+const TIMEOUT: Duration = Duration::from_secs(10);
+
+#[derive(Debug, Clone)]
+pub struct AppUpdateInfo {
+    pub version: String,
+    pub update_available: bool,
+}
```
jgmontoya: Should this include the url?

```rs
pub download_url: Option<String>,  // From the "url" tag
```

src/whitenoise/app_update.rs
```diff
@@ -0,0 +1,531 @@
...
+
+    #[test]
+    fn test_version_clone() {
+        // Test that Clone works correctly (using explicit clone for trait verification)
+        let v1 = Version::from_str("1.2.3").unwrap();
+        #[allow(clippy::clone_on_copy)]
+        let v2 = v1.clone();
+
+        assert_eq!(v1, v2);
+    }
```
jgmontoya: I think this test provides little value and should be removed.
`Copy` implies `Clone`; testing derived trait with `#[allow(clippy::clone_on_copy)]` is a code smell

src/whitenoise/app_update.rs
```diff
@@ -0,0 +1,531 @@
...
+
+    #[test]
+    fn test_version_copy() {
+        // Test that Copy works correctly (Version implements Copy)
+        let v1 = Version::from_str("1.2.3").unwrap();
+        let v2 = v1;
+
+        assert_eq!(v1, v2);
+        assert_eq!(v1.major, 1);
+    }
```
jgmontoya: Testing a trivially-derived `Copy` trait adds no value, we should remove this test.

src/whitenoise/app_update.rs
```diff
@@ -0,0 +1,531 @@
...
+    fn test_app_update_info_creation() {
+        // Test creating AppUpdateInfo struct
+        let info = AppUpdateInfo {
+            version: "1.2.3".to_string(),
+            update_available: true,
+        };
+
+        assert_eq!(info.version, "1.2.3");
+        assert!(info.update_available);
+    }
```
jgmontoya: I think the compiler already guarantees field assignment. We should remove this test.

src/whitenoise/app_update.rs
```diff
@@ -0,0 +1,531 @@
...
+        // Test that AppUpdateInfo can be cloned
+        let info1 = AppUpdateInfo {
+            version: "2.0.0".to_string(),
+            update_available: false,
+        };
+        let info2 = info1.clone();
+
+        assert_eq!(info1.version, info2.version);
+        assert_eq!(info1.update_available, info2.update_available);
+    }
```
jgmontoya: Tests derived `Clone`, no custom logic to verify so we should remove this test.

src/whitenoise/app_update.rs
```diff
@@ -0,0 +1,531 @@
...
+        // Test that Debug formatting works (doesn't panic)
+        let info = AppUpdateInfo {
+            version: "1.0.0".to_string(),
+            update_available: true,
+        };
+
+        let debug_str = format!("{:?}", info);
+        assert!(debug_str.contains("AppUpdateInfo"));
+        assert!(debug_str.contains("1.0.0"));
+    }
```
jgmontoya: Same with this one

src/whitenoise/app_update.rs
```diff
@@ -0,0 +1,531 @@
...
+
+    /// Constants Tests
+
+    #[test]
+    fn test_zapstore_relay_url_is_valid() {
+        // Test that the hardcoded relay URL is a valid WebSocket URL and contains zapstore
+        assert!(ZAPSTORE_RELAY_URL.starts_with("wss://"));
+        assert!(ZAPSTORE_RELAY_URL.contains("zapstore"));
+    }
```
jgmontoya: I think this is a weak validation (just checks prefix and substring). I believe we can just remove it as well.

src/whitenoise/app_update.rs
```diff
@@ -0,0 +1,531 @@
...
+        assert_eq!(version.to_string(), "0.0.0");
+    }
+
+    #[test]
+    fn test_version_display_large_numbers() {
+        // Test Display with large numbers
+        let version = Version::from_str("100.200.300").unwrap();
+
+        assert_eq!(version.to_string(), "100.200.300");
+    }
```
jgmontoya: We can consolidate these tests:

```rs
```rust
#[test]
fn test_version_display() {
    for input in ["1.2.3", "0.0.0", "100.200.300"] {
        let version = Version::from_str(input).unwrap();
        assert_eq!(version.to_string(), input);
    }
}
```
```

File: src/whitenoise/app_update.rs [hunk 14]
```diff
@@ -0,0 +1,531 @@
...
+        let v3 = Version::from_str("1.0.1").unwrap();
+        let v4 = Version::from_str("1.1.0").unwrap();
+        let v5 = Version::from_str("2.0.0").unwrap();
+
+        assert!(v0 < v1);
+        assert!(v1 < v2);
+        assert!(v2 < v3);
+        assert!(v3 < v4);
+        assert!(v4 < v5);
+    }
```
jgmontoya, line 294: These tests can be consolidated in a single test that's also clearer on what we're doing:

```rust
#[test]
fn test_version_ordering() {
    let mut versions: Vec<_> = ["2.0.0", "1.2.3", "1.3.0", "1.9.9", "1.2.4", "1.2.9"]
        .into_iter()
        .map(|s| Version::from_str(s).unwrap())
        .collect();

    versions.sort();

    let sorted: Vec<_> = versions.iter().map(|v| v.to_string()).collect();
    assert_eq!(sorted, ["1.2.3", "1.2.4", "1.2.9", "1.3.0", "1.9.9", "2.0.0"]);
}
```

src/whitenoise/app_update.rs
```diff
@@ -0,0 +1,531 @@
...
+
+    #[test]
+    fn test_version_parse_invalid_patch() {
+        // Test that parsing fails when patch version is not a valid number
+        let result = Version::from_str("1.2.abc");
+
+        assert!(result.is_err());
+        assert_eq!(result.unwrap_err(), "Invalid patch");
+    }
```
jgmontoya, line 240: Here we could consolidate as well:

```rust
#[test]
fn test_version_parse_invalid_components() {
    let cases = [
        ("abc.2.3", "Invalid major"),
        ("1.xyz.3", "Invalid minor"),
        ("1.2.abc", "Invalid patch"),
    ];

    for (input, expected_err) in cases {
        let result = Version::from_str(input);
        assert_eq!(result.unwrap_err(), expected_err, "input: {input}");
    }
}
```

src/whitenoise/app_update.rs
```diff
@@ -0,0 +1,531 @@
...
+use std::time::Duration;
+
+use crate::whitenoise::error::{Result, WhitenoiseError};
+
+const ZAPSTORE_RELAY_URL: &str = "wss://relay.zapstore.dev";
+const WHITE_NOISE_PUBKEY: &str = "75d737c3472471029c44876b330d2284288a42779b591a2ed4daa1c6c07efaf7";
+const TIMEOUT: Duration = Duration::from_secs(10);
+
+#[derive(Debug, Clone)]
+pub struct AppUpdateInfo {
```
jgmontoya: We should derive `PartialEq`, `Eq` as well as this struct is public

src/whitenoise/app_update.rs
```diff
@@ -0,0 +1,531 @@
...
+pub async fn check_for_app_update(current_version: &str) -> Result<AppUpdateInfo> {
+    let client = Client::default();
+
+    let relay_url = RelayUrl::parse(ZAPSTORE_RELAY_URL)
+        .map_err(|e| WhitenoiseError::Other(anyhow::anyhow!(e)))?;
+
+    client
+        .add_relay(relay_url.clone())
+        .await
+        .map_err(|e| WhitenoiseError::Other(anyhow::anyhow!(e)))?;
```
jgmontoya: I think you can just use `.await?` without manually doing the `map_err`

src/whitenoise/app_update.rs
```diff
@@ -0,0 +1,531 @@
...
+
+    tokio::time::timeout(TIMEOUT, client.connect())
+        .await
+        .map_err(|_| {
+            WhitenoiseError::Other(anyhow::anyhow!("Timeout connecting to Zapstore relay"))
+        })?;
+
+    let pubkey = PublicKey::from_hex(WHITE_NOISE_PUBKEY)
+        .map_err(|e| WhitenoiseError::Other(anyhow::anyhow!(e)))?;
```
jgmontoya: Same here, `PublicKey::from_hex(...)?`

jgmontoya (changes requested):
Your code looks good in general.

I think we can do better to make this features less coupled with the actual frontend though.

src/whitenoise/app_update.rs
Quwaysim: Makes sense. Will update soon. Thanks

src/whitenoise/app_update.rs
Quwaysim: 💯 Makes sense.

Quwaysim: About using the arch in the event filter, I think we publish all the relevant arch installation packages to the stores so if user has WN installed, they should get their relevant update also as updates will most probably include the package for their device arch. We could also add this to future improvements.

src/whitenoise/app_update.rs
Quwaysim: I'm with you on doing full semver. However the goal right now is to make this work with what we have at the moment so I'll add this to the discussion on the issue as part of the future updates to this feature.

src/whitenoise/app_update.rs
Quwaysim: It should for when we are ready to give users a cta in the app update banner to redirect to their stores, not needed right now IMO.

src/whitenoise/app_update.rs
Quwaysim: Perks of the Rust Lang, eh! :D

src/whitenoise/app_update.rs
Quwaysim: Valid

src/whitenoise/app_update.rs
Quwaysim: Could you please explain a little bit why? @jgmontoya 
Esp this part "...as this struct is public"

Quwaysim: NVM, I read it as "remove" instead of "derive". I guess I was tired then :D

jgmontoya: We have it under our general rust-code guidelines:

```
Derive `Debug`, `Clone`, `Copy`, `PartialEq`, `Eq` and `Hash` for public types when possible (in this order).

**Rationale:** these traits can be useful for users and can be automatically implemented for most types.
```

Basically it's almost free to add and it's to give some basic ergonomics to the clients/users of the library (so they can for example directly compare and see if two instances of `AppUpdateInfo` are the same or not and stuff like that)

src/whitenoise/app_update.rs
Quwaysim: Didn't know WhiteNoiseError has a From impl for nostr_sdk::client::Error. Thanks.

PR #429, opened 2026-01-08
Fix/chat list sorting

josefinalliende:
When implementing chat list streams in sloth app, I was facing this issue: first chat list wasn't rendering in the expected order (considering last message timestamp and in its absence, chat created at timestamp). Second, when navigating through screens and going back to chat list, the order changed with no reason (no messages had arrived nor new invites)

After vibecoding this changes (I don't know rust 🦀 ) and only changing the rust crate in my local branch, the problem seemed to be fixed: the chat list appeared in correct order and when navigating, it did not change. 

It doesn't make sense to me that adding this tiebreak is the real fix cause in the videos you can see that the timestamps are not identical... but it works. 

Screenshots of the initial order before and after this changes. As you can see in the first picture, the third chat (bad bunny) is not in the correct position because the "hello" group has a created at timestamp that is later than the last "bad bunny" message. 

|Before| After|
|----|----|
|sorting-bug-photo|sorting-bug-fixed-picture|

**Video if the bug** : pay attention at the "bad bunny" avatar in the chat list. As you will notice, it changes it's position on the list every time we go back to the chat list

https://github.com/user-attachments/assets/c1bc08bb-5328-443d-98ce-d59956433eca

**Video of the bug fixed** : again pay attention to the "bad bunny" avatar in the chat list. It keeps always the same position in the chat list

https://github.com/user-attachments/assets/52bef1f4-2482-4510-800c-d648c9e2da9d

If you are curious and want to replicate the bug on sloth, I pushed a branch called `chat-list-order-issue` with 2 commits. In the first one I see the bug. In the second one (where I use the whitenoise rust local crate in this branch) the bug is gone.

## Summary by CodeRabbit

* **Bug Fixes**
  * Chat lists now sort stably and deterministically; when timestamps match, a consistent secondary identifier breaks ties so ordering is reproducible.
  * The initial chat list returned on subscribe is now reliably ordered newest-first.

* **Tests**
  * Added unit and async tests covering tiebreaking, ordering with no messages, identical timestamps, and repeated retrieval stability.

### DIFF
--- a/src/integration_tests/scenarios/chat_list_streaming.rs
+++ b/src/integration_tests/scenarios/chat_list_streaming.rs
@@ -13,6 +13,7 @@ pub struct ChatListStreamingScenario {
 
 impl ChatListStreamingScenario {
     const GROUP_NAME: &'static str = "chat_list_stream_group";
+    const INACTIVE_GROUP_NAME: &'static str = "inactive_group";
 
     pub fn new(whitenoise: &'static Whitenoise) -> Self {
         Self {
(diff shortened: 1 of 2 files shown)


josefinalliende:
> @josefinalliende We'll make a rust dev out of you yet! 😉
> 
> Looking at this a few things stand out.
> 
> 1. I don't think that we were doing any sorting in whitenoise-rs with the streams. Maybe we were in the older aggregated fetch methods.
> 2. Another thing is that when you call `into_values()` on a `HashMap` in Rust it basically randomizes the order of the items. This makes me surprised that the ordering was as stable as it seemed to be (so maybe I'm also wrong on point 1 and I missed the sort somewhere).
> 
> tbh, your solution is pretty sound, but one way to make it a little easier to understand (and potentially extend) would be to add a method that returns a comparable tuple:
> 
> ```rust
>   impl ChatListItem {
>       fn sort_key(&self) -> (Reverse<DateTime<Utc>>, &GroupId) {
>           let timestamp = self.last_message
>               .as_ref()
>               .map(|m| m.created_at)
>               .unwrap_or(self.created_at);
>           (Reverse(timestamp), &self.mls_group_id)
>       }
>   }
> ```
> 
> Then:
> 
> ```rust
>   fn sort_chat_list(items: &mut [ChatListItem]) {
>       items.sort_by(|a, b| a.sort_key().cmp(&b.sort_key()));
>   }
> ```
> 
> This is elegant because:
> 
> * `Reverse` handles descending order naturally
> * Tuple comparison handles the tiebreaker automatically (compares first element, then second if equal)
> * Logic is encapsulated but doesn't impose a global `Ord` on the type (which is what we'd have to do if we wanted to create a custom `Ord` trait for `ChatListItem`)
> * Easy to add more tiebreakers later — just extend the tuple

Thanks @erskingardner for taking a look, I will add your suggestion
1) LOL you are right. So I was wondering why by only adding the tie breaker it worked but actually the fix was in mod.rs calling the sort 😅 . 
2) It looked stable in sloth chat list because the get_chat_list method (without streams) that I was using before did call the sort, the sort was missing only in the subscribe_to_chat_list.

src/whitenoise/mod.rs
```diff
@@ -643,7 +643,8 @@ impl Whitenoise {
             }
         }
 
-        let initial_items: Vec<chat_list::ChatListItem> = items_map.into_values().collect();
+        let mut initial_items: Vec<chat_list::ChatListItem> = items_map.into_values().collect();
+        chat_list::sort_chat_list(&mut initial_items);
```
jgmontoya, line 647: Nice catch!

src/whitenoise/chat_list.rs
```diff
@@ -55,6 +56,17 @@ pub struct ChatListItem {
...
+    fn sort_key(&self) -> (Reverse<DateTime<Utc>>, &GroupId) {
+        let timestamp = self
+            .last_message
+            .as_ref()
+            .map(|m| m.created_at)
+            .unwrap_or(self.created_at);
+        (Reverse(timestamp), &self.mls_group_id)
+    }
+}
+
```
jgmontoya, line 69: Good idea to refactor this

jgmontoya:
> > Looks good, I'll give you a hand with the tests
> 
> @jgmontoya I've just added a unit test for the subscribe to chat list method... but if you think we also need integration tests for this, I'd love some help there

added the change on the integration test

PR #433, opened 2026-01-14
feat(read-tracking): implement read tracking with basic functionalities

jgmontoya:
## Summary by CodeRabbit

* **New Features**
  * Users can mark messages as read; the system tracks each account’s last-read message per chat.
  * Chat list items now show per-chat unread message counts.

* **Tests**
  * Added integration and unit tests covering marking-as-read, unread-count calculations, and related edge cases.

* **Chores**
  * Database migration added to persist per-chat last-read markers.

### DIFF
--- a/src/whitenoise/chat_list.rs
+++ b/src/whitenoise/chat_list.rs
@@ -54,6 +54,9 @@ pub struct ChatListItem {
 
     /// The public key of the user who invited the account to the group.
     pub welcomer_pubkey: Option<PublicKey>,
+
+    /// Number of unread messages in this chat
+    pub unread_count: usize,
 }
 
 impl ChatListItem {
@@ -128,6 +131,7 @@ fn assemble_chat_list_items(
     users_by_pubkey: &HashMap<PublicKey, User>,
     image_paths: &HashMap<GroupId, PathBuf>,
     membership_map: &HashMap<GroupId, AccountGroup>,
+    unread_counts: &HashMap<GroupId, usize>,
 ) -> Vec<ChatListItem> {
     groups
         .iter()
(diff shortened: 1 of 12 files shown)


src/whitenoise/chat_list.rs
```diff
@@ -215,6 +221,19 @@ impl Whitenoise {
...
 
+        // Compute unread counts for all groups
+        let mut unread_counts = HashMap::new();
+        for (group_id, account_group) in &membership_map {
+            let count = AggregatedMessage::count_unread_for_group(
+                group_id,
+                account_group.last_read_message_id.as_ref(),
+                &self.database,
+            )
+            .await
```
dannym-arx: rather than looping, could we make a single query that returns (group_id, count) for every group?

src/whitenoise/database/aggregated_messages.rs
```diff
@@ -479,6 +479,65 @@ impl AggregatedMessage {
...
+                     WHERE mls_group_id = ? AND kind = 9 AND deletion_event_id IS NULL",
+                )
+                .bind(group_id.as_slice())
+                .fetch_one(&database.pool)
+                .await?
+            }
+        };
+
+        Ok(count as usize)
+    }
```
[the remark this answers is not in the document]
jgmontoya: This is an acceptable edgecase

src/whitenoise/chat_list.rs
jgmontoya: I thought it might be too complex but I will give it a try

mubarakcoded: I agree with @dannym-arx, this creates  N+1 query problem. If a user has 50 groups, this means 50 separate database queries, that's going to be really slow. Can we batch this into a single query that returns all unread counts at once?

src/whitenoise/chat_list.rs
```diff
@@ -319,7 +339,16 @@ impl Whitenoise {
...
 
-        // 8. Assemble and return ChatListItem
+        // 8. Compute unread count
+        let unread_count = AggregatedMessage::count_unread_for_group(
+            group_id,
+            account_group.last_read_message_id.as_ref(),
+            &self.database,
+        )
+        .await
+        .unwrap_or(0);
```
mubarakcoded: We will be silently swallowing database errors here if the db query fails for some reason, user might see incorrect unread counts. Should we be propagating this error instead or at minimum logging it?

PR #439, opened 2026-01-19
fix: update derive_upload_keypair() with Secret type and version parameter

mubarakcoded:
## Summary

Updates `derive_upload_keypair()` calls to match new signature from MDK security improvements.

## Changes

- Add `mdk-storage-traits` dependency for `Secret` type
- Wrap `image_key` in `Secret::new()` for proper zeroization
- Add version parameter (v2 for current extension format)

## Files Changed

- `Cargo.toml`: Added `mdk-storage-traits` dependency
- `src/whitenoise/groups.rs`: Updated two `derive_upload_keypair()` calls

## Note

Build will have errors until remaining breaking changes are fixed in subsequent PRs.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,10 +31,11 @@ keyring = { version = "3.6", features = [
 ] }
 lightning-invoice = "0.33.1"
 
-mdk-core = { version = "0.5.1", git = "https://github.com/parres-hq/mdk", rev = "46738bd1e16e0aa78255c7f70b1d90698e968746", features = [
+mdk-core = { version = "0.5.1", git = "https://github.com/marmot-protocol/mdk", rev = "46738bd1e16e0aa78255c7f70b1d90698e968746", features = [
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -35,6 +35,7 @@ mdk-core = { version = "0.5.1", git = "https://github.com/parres-hq/mdk", rev =
     "mip04",
 ] }
 mdk-sqlite-storage = { version = "0.5.1", git = "https://github.com/parres-hq/mdk", rev = "46738bd1e16e0aa78255c7f70b1d90698e968746" }
+mdk-storage-traits = { version = "0.5.1", git = "https://github.com/parres-hq/mdk", rev = "46738bd1e16e0aa78255c7f70b1d90698e968746" }
```
erskingardner: we should update the URLs to be the correct `https://github.com/marmot-protocol/mdk`

PR #441, opened 2026-01-19
chore(deps): upgrade reqwest to 0.13 and fix cargo-audit

erskingardner:
## Summary
- Upgrade `reqwest` from 0.11 to 0.13 with updated `rustls` feature (renamed from `rustls-tls`)
- Add `RUSTSEC-2026-0002` (lru unsound advisory) to ignored advisories, awaiting upstream fix in nostr-sdk
- Add required `permissions` block to cargo-audit workflow job (`checks: write`, `contents: read`)
- Fix flaky integration tests by making cleanup more robust

## Test Infrastructure Fixes
The integration tests had a flaky failure where cleanup could leave the database in a corrupted state:

1. **Cleanup now always runs** - Previously, cleanup was skipped when scenarios failed, leaving state behind
2. **Cleanup failures now fail the scenario** - Prevents corrupted state from affecting subsequent scenarios  
3. **`delete_all_data` is now fully atomic** - Uses DELETE instead of DROP TABLE + migrate
   - Preserves schema even if interrupted
   - No risk of migration failures leaving database without tables
4. **Added retry logic** - Exponential backoff for database lock contention
5. **Better timing** - Increased delays to let background tasks settle before cleanup

## Summary by CodeRabbit

* **Chores**
  * Upgraded HTTP client and adjusted TLS configuration.
  * Tightened CI audit job permissions and expanded its ignore list.

* **Bug Fixes / Reliability**
  * Cleanup now always runs and surfaces failures to prevent silent resource issues.
  * Database reset is safer: preserves schema, retries on locks, and resets counters when applicable.

* **Tests**
  * Expanded and reorganized unit and integration tests; increased timing tolerance for message streaming.

### DIFF
--- a/src/whitenoise/database/mod.rs
+++ b/src/whitenoise/database/mod.rs
@@ -171,7 +171,6 @@ impl Database {
                             attempt,
                             max_retries
                         );
-                        // Wait before retrying, with exponential backoff
                         tokio::time::sleep(tokio::time::Duration::from_millis(
                             100 * attempt as u64,
                         ))


.github/workflows/ci.yml
```diff
@@ -78,7 +81,8 @@ jobs:
           # Skip advisories:
           # - RUSTSEC-2023-0071: RSA Marvin Attack (transitive via sqlx-mysql, not used in our SQLite-only app)
           # - RUSTSEC-2024-0384: instant unmaintained (transitive via rust-nostr, low risk)
-          ignore: RUSTSEC-2023-0071,RUSTSEC-2024-0384
+          # - RUSTSEC-2026-0002: lru unsound (transitive via nostr-sdk, awaiting upstream fix)
+          ignore: RUSTSEC-2023-0071,RUSTSEC-2024-0384,RUSTSEC-2026-0002
```
jgmontoya, line 85: Should we maybe create an issue for this just so we remember to check back again at some point in the future?

src/whitenoise/database/mod.rs
```diff
@@ -139,35 +139,89 @@ impl Database {
...
+                    if is_lock_error && attempt < max_retries {
+                        tracing::warn!(
+                            "Database locked during cleanup (attempt {}/{}), retrying...",
+                            attempt,
+                            max_retries
+                        );
+                        // Wait before retrying, with exponential backoff
+                        tokio::time::sleep(tokio::time::Duration::from_millis(
+                            100 * attempt as u64,
+                        ))
```
jgmontoya: This is linear backoff, not exponential. I'd just remove the comment.

.github/workflows/ci.yml
erskingardner: Good call. I'm usually good at remembering this sort of thing but issues are even better. :)

PR #442, opened 2026-01-20
feat(accounts_groups): add pin_order

jgmontoya:
## Summary by CodeRabbit

* **New Features**
  * Users can now pin chats to keep them at the top of the chat list
  * Pinned chats are organized by custom order, with unpinned chats appearing below sorted by recent activity

### DIFF
--- a/src/whitenoise/accounts_groups.rs
+++ b/src/whitenoise/accounts_groups.rs
@@ -25,6 +25,10 @@ pub struct AccountGroup {
     pub user_confirmation: Option<bool>,
     pub welcomer_pubkey: Option<PublicKey>,
     pub last_read_message_id: Option<EventId>,
+    /// Pin order for chat list sorting.
+    /// - `None` = not pinned (appears after pinned chats)
+    /// - `Some(n)` = pinned, lower values appear first
+    pub pin_order: Option<i64>,
     pub created_at: DateTime<Utc>,
     pub updated_at: DateTime<Utc>,
 }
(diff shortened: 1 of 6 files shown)


src/whitenoise/accounts_groups.rs
```diff
@@ -204,6 +208,27 @@ impl Whitenoise {
...
+        let account_group = AccountGroup::get(self, &account.pubkey, mls_group_id)
+            .await?
+            .ok_or(WhitenoiseError::GroupNotFound)?;
+
+        let updated = account_group
+            .update_pin_order(pin_order, &self.database)
+            .await?;
+
+        Ok(updated)
+    }
```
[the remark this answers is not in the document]
jgmontoya: This is not necessary because all pin order changes are triggered by the user/frontend itself.

src/whitenoise/accounts_groups.rs
```diff
@@ -204,6 +208,27 @@ impl Whitenoise {
         let account_group = AccountGroup::get(self, &account.pubkey, group_id).await?;
         Ok(account_group.and_then(|ag| ag.last_read_message_id))
     }
+
+    /// Sets the pin order for a chat.
+    ///
+    /// - `None` = unpin the chat
+    /// - `Some(n)` = pin the chat with order n (lower values appear first)
+    pub async fn set_chat_pin_order(
```
erskingardner, line 216: Do we need to have some sort of bulk update method that would let you reorder pinned chats? Right now you just set a number and then have to loop through and update the rest of the chats? Or are you just assuming that pinned chats are all given the same low number and then are re-ordered based on the last_message timestamp?

jgmontoya: The current design supports different uses:

1. We can have the 'pin chat' action just bring whatever its pin to the top (it can set pin_order to something like max_int - now() for example)
2. We can have a feature that allows for "drag to reorder" and allows the user to give a specific order for their chats (then the front just calls set_chat_pin_order to whatever needs to change to make that ordering)
3. We can have pinned chats be ordered among themselves according to the last message (by having the same pin_order)

jgmontoya: I don't think a bulk update is needed, at least for now

erskingardner: Okay, I think that's fine that we don't have any bulk update method for now. Um I definitely think people are gonna want to reorder these and have specific orderings. So I think uh eventually we'll need to have um some method to allow for that sort of drag and drop reordering. But we can certainly tackle that in the future

jgmontoya: > Um I definitely think people are gonna want to reorder these and have specific orderings. So I think uh eventually we'll need to have um some method to allow for that sort of drag and drop reordering. But we can certainly tackle that in the future

This can be done without any new method

PR #449, opened 2026-01-21
Remove default pet names

josefinalliende:
In the sloth flutter app we were having some issues with metadata on sign up: https://github.com/marmot-protocol/sloth/issues/41

In the whitenoise app, the identity was created before showing the singn up form, the metadata of the new identity was fetched (with the random pet name assigned by rust) and then the user could edit the metadata. The problem with this is that if the user closed the app before submitting the signup form of went back, the new identity was created anyways

In sloth we are showing the signup form with the metadata fields but only creating the identity when the user clicks the submit button, and then updating the metadata.

The problem with approach is that sometimes the metadata update with random pet names triggered here in the rust crate overrides the metadata update triggered from the flutter side on the form submission.  For this reason, we want to remove the initial metadata setup from rust and delegate that responsibility to the flutter side.

When removing the petname dependency, checks failed so I needed to add the "derive" feature in clap dependency to make them pass again.
 

## Summary by CodeRabbit

* **Chores**
  * Updated core dependencies for improved compatibility and feature support.
  * Removed unused dependencies to streamline the project.

* **Bug Fixes**
  * Corrected metadata initialization behavior during account creation.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ base64 = "0.22"
 blurhash = "0.2.3"
 chacha20poly1305 = "0.10"
 chrono = { version = "0.4.40", features = ["serde"] }
-clap = "4.5.37"
+clap = { version = "4.5.37", features = ["derive"] }
 dashmap = "6.1"
 hex = "0.4"
 image = "0.24"
@@ -61,7 +61,6 @@ nostr-sdk = { version = "0.43", features = [
 #     "nip59",
 # ] }
 
-petname = "2.0.2"
 rand = "0.9"
 reqwest = { version = "0.11", features = [
     "multipart",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2019,15 +2019,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7943c866cc5cd64cbc25b2e01621d07fa8eb2a1a23160ee81ce38704e97b8ecf"
 
 [[package]]
-name = "itertools"
-version = "0.14.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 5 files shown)


Cargo.toml
```diff
@@ -19,7 +19,7 @@ base64 = "0.22"
 blurhash = "0.2.3"
 chacha20poly1305 = "0.10"
 chrono = { version = "0.4.40", features = ["serde"] }
-clap = "4.5.37"
+clap = { version = "4.5.37", features = ["derive"] }
```
jgmontoya, line 22: why did this one change?

pepina-dev: So the thing is that when I removed the pet name dependency, tests were failing and I could only get them fixed after having adding this derive feature explicitly. 

Clap was a dependency of petname, so my guess is that petname had the clap derive feature ...

```
[[package]]
name = "petname"
version = "2.0.2"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "9cd31dcfdbbd7431a807ef4df6edd6473228e94d5c805e8cf671227a21bad068"
dependencies = [
 "anyhow",
 "clap",
 "itertools",
 "proc-macro2",
 "quote",
 "rand 0.8.5",
]
```

pepina-dev: Without that I saw errors like this: 
```
error: cannot find derive macro `Parser` in this scope
 --> src/bin/integration_test.rs:8:10
  |
8 | #[derive(Parser, Debug)]
  |          ^^^^^^
```
```
error: cannot find attribute `clap` in this scope
  --> src/bin/integration_test.rs:19:7
   |
19 |     #[clap(value_name = "SCENARIO")]
   |       ^^^^
```

PR #450, opened 2026-01-21
Add nonce and scheme_version to MediaReference for MIP-04 v2

mubarakcoded:
This PR adds database schema changes and code updates to support MDK's `MediaReference` requirements for `nonce` and `scheme_version` fields, which are mandatory for MIP-04 v2 encrypted media decryption.

### Database Migration
- **New migration 0021**: Adds `nonce` (TEXT) and `scheme_version` (TEXT) columns to `media_files` table
- Both fields are nullable for backward compatibility with existing records and group images
- Added partial index on `scheme_version` (WHERE NOT NULL) for query performance optimization

### Code Updates
- Updated `MediaFile`, `MediaFileRow`, and `MediaFileParams` structs with new optional fields
- Modified all SQL INSERT and SELECT queries to include new columns
- Extract and store `nonce` (hex-encoded) and `scheme_version` from MDK's `parse_imeta_tag()` when receiving chat media
- Retrieve and validate encryption metadata when decrypting chat media in `download_and_decrypt_chat_media_blob()`
- Updated all test cases (11 instances) to pass `None` for group images

### Technical Details
- **nonce**: Stored as hex-encoded string (24 chars for 12 bytes)
- **scheme_version**: Stored as string (e.g., "mip04-v2")
- Proper validation on retrieval: hex decode, length check (must be exactly 12 bytes)
- Clear separation: `None` for group images (use key/nonce encryption), required for chat_media

## Breaking Change ⚠️

**Old chat_media records** (created before this migration) cannot be decrypted because they lack the required encryption metadata. This is intentional - MDK removed support for MIP-04 v1 due to nonce reuse vulnerabilities.

**Group images are unaffected** as they use a different encryption scheme (key/nonce pairs, not MDK).

### DIFF
--- a/db_migrations/0021_add_encryption_metadata_to_media_files.sql
+++ b/db_migrations/0021_add_encryption_metadata_to_media_files.sql
@@ -1,0 +1,33 @@
+-- Migration 0021: Add encryption metadata to media_files for MIP-04 decryption
+--
+-- This migration adds nonce and scheme_version columns required for MDK's
+-- decrypt_from_download() function. These values are extracted from the IMETA tag
+-- when receiving chat media.
+--
+-- Both fields are required for decryption:
+-- - nonce: Encryption nonce (hex-encoded, 24 chars for 12 bytes)
+-- - scheme_version: Encryption version (e.g., "mip04-v2")
+--
+-- IMPORTANT: Old chat_media records (created before this migration) cannot be
+-- decrypted because they lack these required fields. This is a breaking change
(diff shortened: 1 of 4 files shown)


jgmontoya (commented):
I think we should just drop the table and recreate it to make the columns NOT NULL like the others

erskingardner:
> I think we should just drop the table and recreate it to make the columns NOT NULL like the others

Yeah I agree with @jgmontoya on the drop and recreate to get NOT NULL fields.

PR #460, opened 2026-01-26
feat(notification-streaming): add stream for notifications

jgmontoya:
## Summary by CodeRabbit

* **New Features**
  * Real-time notification streaming for NewMessage and GroupInvite events, including optional group name, DM detection, sender/receiver info, content, timestamp and trigger type.
  * Public subscription API and exported notification types; supports multiple subscribers and filters self-sent messages.
  * Notifications now emitted for group invites and additional incoming message cases.

* **Tests**
  * Unit tests for the streaming manager.
  * New end-to-end integration scenario and verification helpers for notification streaming.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -171,15 +171,6 @@ jobs:
         env:
           RUST_LOG: error
 
-      - name: Integration Test
-        if: matrix.run-integration
-        run: |
-          mkdir -p ./dev/data/integration_test/
-          RUST_LOG=warn,sqlx=info,refinery_core=error,keyring=info,nostr_relay_pool=error,mdk_sqlite_storage=error,tungstenite=error,integration_test=debug \
-          cargo run --bin integration_test --features integration-tests -- \
-          --data-dir ./dev/data/integration_test/ --logs-dir ./dev/data/integration_test/
-          rm -rf ./dev/data/integration_test/
-
       - name: Extract coverage percentage
         if: matrix.run-integration
         id: coverage
@@ -275,6 +266,15 @@ jobs:
               repo: context.repo.repo,
               body: body
             });
+
+      - name: Integration Test
+        if: matrix.run-integration
+        run: |
+          mkdir -p ./dev/data/integration_test/
+          RUST_LOG=warn,sqlx=info,refinery_core=error,keyring=info,nostr_relay_pool=error,mdk_sqlite_storage=error,tungstenite=error,integration_test=debug \
+          cargo run --bin integration_test --features integration-tests -- \
+          --data-dir ./dev/data/integration_test/ --logs-dir ./dev/data/integration_test/
+          rm -rf ./dev/data/integration_test/
 
       - name: Save coverage baseline (master only)
         if: matrix.run-integration && github.ref == 'refs/heads/master' && github.event_name == 'push'


src/whitenoise/notification_streaming/manager.rs
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    pub fn subscribe(&self) -> broadcast::Receiver<NotificationUpdate> {
+        self.sender.subscribe()
+    }
+
+    pub fn emit(&self, update: NotificationUpdate) {
+        // receiver_count() is O(1) - just reads an AtomicUsize
+        if self.sender.receiver_count() > 0 {
+            let _ = self.sender.send(update);
```
mubarakcoded, line 24: If a send fails for any reason, the error is silently discarded. Should we add tracing?

src/whitenoise/notification_streaming/mod.rs
```diff
@@ -0,0 +1,315 @@
...
+        message: &ChatMessage,
+        group_name: Option<String>,
+    ) {
+        if self.is_own_account(&message.author).await {
+            return;
+        }
+
+        if !self.notification_stream_manager.has_subscribers() {
+            return;
+        }
```
mubarakcoded, line 38: Here, you're doing a database query (is_own_account) before checking if there are subscribers(O(1) check). Can we flip the order to avoid unnecessary DB queries when no one is listening?

jgmontoya: yep, nice catch

src/whitenoise/notification_streaming/manager.rs
jgmontoya: The only way `send()` fails here is if all receivers dropped between our `receiver_count() > 0` check and the actual send, a benign race where the notification was headed to listeners that no longer exist. Not worth logging since it's expected and not actionable.

mubarakcoded: Agreed.

mubarakcoded (approved):
LGTM!

src/whitenoise/notification_streaming/manager.rs
```diff
@@ -0,0 +1,137 @@
...
+        Self { sender }
+    }
+
+    pub fn subscribe(&self) -> broadcast::Receiver<NotificationUpdate> {
+        self.sender.subscribe()
+    }
+
+    pub fn emit(&self, update: NotificationUpdate) {
+        // receiver_count() is O(1) - just reads an AtomicUsize
+        if self.sender.receiver_count() > 0 {
```
erskingardner: why not use the `has_subscribers` method here?

jgmontoya: no reason, I just missed it :P

src/integration_tests/test_cases/notification_streaming/verify_notification_update.rs
```diff
@@ -0,0 +1,180 @@
...
+        // Verify group name if expected
+        if let Some(expected_name) = &self.expected_group_name {
+            let expected_group = context.get_group(expected_name)?;
+            assert_eq!(
+                update.mls_group_id, expected_group.mls_group_id,
+                "Expected group '{}' but got different group ID",
+                expected_name
+            );
+            tracing::info!("Group ID matches expected group '{}'", expected_name);
+        }
```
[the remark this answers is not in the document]
jgmontoya: updated the name (now expect_group) to reduce confusion. We don't need to verify the group name here, checking the group id is enough

PR #464, opened 2026-01-28
chore: bump nostr-sdk to 0.44 and fix deprecations

mubarakcoded:
## Summary
- Bump nostr-sdk, nwc, and nostr-blossom from 0.43 to 0.44
- Update MDK to rev 4156904 (includes nostr 0.44 support)
- Remove `lmdb` feature (not available in nostr-sdk 0.44)
- Replace deprecated `Timestamp::as_u64()` with `as_secs()` throughout codebase
- Add `epoch: None` field to Message struct initializers (new MDK field)

### DIFF
--- a/src/whitenoise/event_processor/event_handlers/handle_mls_message.rs
+++ b/src/whitenoise/event_processor/event_handlers/handle_mls_message.rs
@@ -151,7 +151,7 @@ impl Whitenoise {
             mls_group_id: group_id.clone(),
             event: inner_event,
             wrapper_event_id: event_id, // Reuse event_id as placeholder, we don't need it
-            epoch: None,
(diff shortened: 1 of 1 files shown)


src/whitenoise/event_processor/event_handlers/handle_mls_message.rs
```diff
@@ -151,6 +151,7 @@ impl Whitenoise {
             mls_group_id: group_id.clone(),
             event: inner_event,
             wrapper_event_id: event_id, // Reuse event_id as placeholder, we don't need it
+            epoch: None,
```
jgmontoya: I'd maybe add a comment here explaining why we use this non-intuitive value (like we do for `wrapper_event_id`)

PR #467, opened 2026-01-30
feat: upgrade MDK and nostr-sdk to 0.44

mubarakcoded:
## Summary

Upgrades MDK to latest master (7c3157c) and nostr-sdk/nwc/nostr-blossom to 0.44, addressing all breaking changes.

### MDK Breaking Changes Addressed
- **Secret\<T\> type wrapper**: Updated `decrypt_group_image()`, `derive_upload_keypair()`, `resolve_group_image_path()`, and `GroupImageUpload` to use `Secret<[u8; 32]>` / `Secret<[u8; 12]>` for encryption keys and nonces
- **Pagination & scoped message lookup**: Updated MDK messages API for `Option<Pagination>` parameter and scoped `find_message_by_event_id()` requiring both `mls_group_id` and `event_id`
- **MdkSqliteStorage::new() with keyring**: Updated storage initialization with keyring parameters and added `keyring-core` as dev dependency with mock store for tests
- **MIP-04 v2 encrypted media**: Added `nonce` and `scheme_version` fields to `MediaReference` for MIP-04 v2 decryption
- **Key package tag format**: Updated key package tag extraction for MIP-00 compliant hex formats
- **Message epoch field**: Added `epoch` field to Message struct construction
- **RETURNING clause**: Added encryption metadata columns to RETURNING clauses

### nostr-sdk 0.44 Breaking Changes Addressed
- Replaced deprecated `Timestamp::as_u64()` with `as_secs()`
- Updated Kind matching to use enum variants directly (`Kind::ChatMessage`, `Kind::Reaction`, `Kind::EventDeletion`)
- Fixed giftwrap test to align rumor author with seal signer per NIP-59 validation
- Removed `lmdb` feature (no longer available in 0.44)

## Test plan
- [x] All 687 unit tests passing
- [x] cargo clippy passes with no warnings
- [x] cargo fmt applied

## Release Notes

* **New Features**
  * Added encryption metadata (nonce, scheme_version) to media files for improved decryption support
  * Implemented secure key wrapping for group image encryption
  * Enhanced message structure with epoch tracking

* **Bug Fixes**
  * Corrected timestamp calculations throughout the codebase

* **Chores**
  * Updated protocol and storage dependencies to latest versions
  * Enhanced database schema with encryption metadata support

* **Tests**
  * Added comprehensive app settings persistence test suite
  * Improved integration test infrastructure with keyring initialization

### DIFF
--- a/db_migrations/0024_add_encryption_metadata_to_media_files.sql
+++ b/db_migrations/0024_add_encryption_metadata_to_media_files.sql
@@ -1,4 +1,4 @@
--- Migration 0023: Add encryption metadata to media_files for MIP-04 decryption
+-- Migration 0024: Add encryption metadata to media_files for MIP-04 decryption
 --
 -- This migration adds nonce and scheme_version columns required for MDK's
 -- decrypt_from_download() function. These values are extracted from the IMETA tag


db_migrations/0024_add_encryption_metadata_to_media_files.sql
jgmontoya, line 1: Should we just delete all previous records here? wdyt @erskingardner

erskingardner: yeah, definitely we should

dannym-arx, line 1: Just add a delete here, like @jgmontoya proposed, and the rest looks good to me

src/whitenoise/accounts.rs
```diff
@@ -265,7 +265,8 @@ impl Account {
         data_dir: &Path,
     ) -> core::result::Result<MDK<MdkSqliteStorage>, AccountError> {
         let mls_storage_dir = data_dir.join("mls").join(pubkey.to_hex());
-        let storage = MdkSqliteStorage::new(mls_storage_dir)?;
+        let db_key_id = format!("mdk.db.key.{}", pubkey.to_hex());
+        let storage = MdkSqliteStorage::new(mls_storage_dir, "com.whitenoise.app", &db_key_id)?;
```
jgmontoya, line 269: should this "com.whitenoise.app" be configurable at the app settings level?

jgmontoya (commented):
Looks good! Just one question

src/whitenoise/accounts.rs
mubarakcoded: No, this shouldn't be configurable at the app settings level. "com.whitenoise.app" is a service identifier for the OS keychain, it tells the platform credential store (Keychain, Keystore, etc.) which app owns the database encryption key. Every consumer of MDK would use their own identifier (e.g. "com.otherapp.messenger"). Changing it at runtime would break the keychain lookup and lock users out of their encrypted database.

jgmontoya: But it seems like this could cause issues if there's multiple apps/frontends that use the whitenoise-rs internally. Particularly on desktop, no?

mubarakcoded: Good point. For the multi-frontend case, we could extract it to a const that each consumer passes in at initialization (e.g., Whitenoise::new(config) with a keyring_service field), rather than hardcoding it deep in accounts.rs. That way each frontend sets it once at startup like "com.whitenoise.desktop", "com.whitenoise.mobile", etc. But it's still not a runtime changeable user setting. It's a build/init-time constant per consumer, not an app settings toggle. We could track that as a separate follow-up issue. WDYT?

jgmontoya: Ah yes, that's what I meant when I was saying app settings, I was actually thinking of the config we pass upon initialization!

jgmontoya: Let's do that and have "com.whitenoise.app" be the default

mubarakcoded: Okay. I will handle this in separate PR. Good ?

erskingardner: I'm fine with it being something in a follow up as long as it's a quick follow up

db_migrations/0023_add_encryption_metadata_to_media_files.sql
```diff
@@ -0,0 +1,37 @@
+-- Migration 0023: Add encryption metadata to media_files for MIP-04 decryption
```
erskingardner: You and @dannym-arx both have 0023 migrations - someone is going to have to update.

src/whitenoise/accounts.rs
mubarakcoded: Yeah, It's going to be a quick one.

db_migrations/0023_add_encryption_metadata_to_media_files.sql
mubarakcoded: I have renamed mine.

db_migrations/0024_add_encryption_metadata_to_media_files.sql
```diff
@@ -0,0 +1,37 @@
...
+-- decrypted because they lack these required fields. This is a breaking change
+-- from MDK - the encryption parameters were previously not stored in IMETA tags.
+-- Group images are unaffected as they use a different encryption scheme (key/nonce).
+--
+-- See: MDK MediaReference struct requirements
+-- See: MIP-04 specification https://github.com/marmot-protocol/marmot/blob/master/04.md
+
+-- Delete legacy chat_media records that lack encryption metadata
+-- Group images are preserved as they use a different encryption scheme (key/nonce)
+DELETE FROM media_files WHERE media_type = 'chat_media';
```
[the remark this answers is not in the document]
mubarakcoded: There's no subset worth preserving, we are deleting all chat_media records anyway, moving the delete after the ALTER TABLE and adding the NULL check would produce the same result since every pre-migration chat_media row will have NULL nonce and scheme_version by definition.

PR #468, opened 2026-02-02
refactor: make keyring service identifier configurable via WhitenoiseConfig

mubarakcoded:
## Summary
- Extract hardcoded `"com.whitenoise.app"` keyring service identifier from `accounts.rs` into a configurable `keyring_service` field on `WhitenoiseConfig`
- Add `DEFAULT_KEYRING_SERVICE` constant with the existing value as the default
- Update `Account::create_mdk()` signature to accept `keyring_service` parameter and thread it through all 40 call sites

## Motivation
Follow-up from PR review on #mdk-upgrade — the keyring service identifier should be an init-time config parameter with a sensible default rather than hardcoded deep in the codebase. This allows consumers to override it if needed.

No behavioral change — all existing code paths use the same default value.

## Test plan
- [x] All 763 unit tests pass
- [x] Integration tests build cleanly (`cargo build --features integration-tests`)
- [x] No change in runtime behavior (same default value used everywhere)

## Summary by CodeRabbit

* **New Features**
  * Configurable keyring service identifier added to configuration; startup validates it.

* **Refactor**
  * Centralized per-account key manager initialization (new Whitenoise-backed factory) and propagated keyring selection across groups, messages, media, and subscription flows.

* **Tests**
  * Expanded tests: retry/backoff coverage, invalid-path handling, CI-friendly mock keyring support, and test updates to use the new MDK creation flow.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2613,9 +2613,9 @@ dependencies = [
 
 [[package]]
 name = "num-conv"
-version = "0.1.0"
+version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "51d515d32fb182ee37cda2ccdcb92950d6a3c2893aa280e540671c2cd0f3b1d9"
+checksum = "cf97ec579c3c42f953ef76dbf8d55ac91fb219dde70e49aa4a6b7d74e9919050"
 
 [[package]]
 name = "num-integer"
@@ -4164,30 +4164,30 @@ dependencies = [
 
 [[package]]
 name = "time"
-version = "0.3.44"
+version = "0.3.47"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "91e7d9e3bb61134e77bde20dd4825b97c010155709965fedf0f49bb138e52a9d"
+checksum = "743bd48c283afc0388f9b8827b976905fb217ad9e647fae3a379a9283c4def2c"
 dependencies = [
  "deranged",
  "itoa",
  "num-conv",
  "powerfmt",
- "serde",
+ "serde_core",
  "time-core",
  "time-macros",
 ]
 
 [[package]]
 name = "time-core"
-version = "0.1.6"
+version = "0.1.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "40868e7c1d2f0b8d73e4a8c7f0ff63af4f6d19be117e90bd73eb1d62cf831c6b"
+checksum = "7694e1cfe791f8d31026952abf09c69ca6f6fa4e1a1229e18988f06a04a12dca"
 
 [[package]]
 name = "time-macros"
-version = "0.2.24"
+version = "0.2.27"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "30cfb0125f12d9c277f35663a0a33f8c30190f4e4574868a330595412d34ebf3"
+checksum = "2e70e4c5a0e0a8a4823ad65dfe1a6930e4f4d756dcd9dd7939022b5e8c501215"
 dependencies = [
  "num-conv",
  "time-core",


src/whitenoise/mod.rs
```diff
@@ -43,6 +43,9 @@ use crate::init_tracing;
 use crate::nostr_manager::NostrManager;
 
 use crate::types::ProcessableEvent;
+
+/// Default keyring service identifier used by MDK for SQLCipher key management.
+pub const DEFAULT_KEYRING_SERVICE: &str = "com.whitenoise.app";
```
erskingardner: I don't know that we want this to have a default value. If someone is going to use this crate they should be forced to pass their own identifier. Having overlapping identifiers would cause failures in the secret storage right?

src/whitenoise/event_processor/event_handlers/handle_giftwrap.rs
```diff
@@ -281,9 +291,10 @@ impl Whitenoise {
         account: &Account,
         data_dir: &std::path::Path,
         keys: Keys,
+        keyring_service: &str,
```
erskingardner: I don't like the name `keyring_service` for the arg here. make me think of a service or class or something but then it's a `&str`. Even though it's super long, I think I would be explicit with something like `keyring_service_identifier`

src/whitenoise/accounts.rs
```diff
@@ -348,10 +348,26 @@ impl Account {
...
         let mls_storage_dir = data_dir.join("mls").join(pubkey.to_hex());
         let db_key_id = format!("mdk.db.key.{}", pubkey.to_hex());
-        let storage = MdkSqliteStorage::new(mls_storage_dir, "com.whitenoise.app", &db_key_id)?;
+        let storage = MdkSqliteStorage::new(mls_storage_dir, keyring_service, &db_key_id)?;
+        Ok(MDK::new(storage))
+    }
+
+    /// Create an MDK instance without encryption (for testing only).
+    ///
+    /// This bypasses the keyring requirement, allowing tests to run in CI environments
```
erskingardner: Do we know that running it with keyring won't work? Our CI is running on either linux or macos with github, both of which are supported by the keyring package. 

The reason I bring this up is that the more things that we have running differently on CI than on the real app, the more likely it is that we miss some issue that prevents the app from working.

dannym-arx: for linux we need to install a libsecret compatible keyring for it to work, I don't think that the CI runner has it by default, need to double check, but I completely agree

src/whitenoise/event_processor/event_handlers/handle_giftwrap.rs
jgmontoya: `keyring_service_id` ??

src/whitenoise/mod.rs
mubarakcoded: Good catch. Agreed. I will remove the default

src/whitenoise/event_processor/event_handlers/handle_giftwrap.rs
mubarakcoded: renamed to keyring_service_id, it's short and precise.

mubarakcoded:
> Not 100% convinced but maybe we should create a helper method
> 
> ```rust
> // src/whitenoise/mod.rs
> 
> impl Whitenoise {
>     pub(crate) fn create_mdk_for_account(&self, pubkey: PublicKey) -> Result<MDK<MdkSqliteStorage>, AccountError> {
>         Account::create_mdk(pubkey, &self.config.data_dir, &self.config.keyring_service)
>     }
> }
> ```
> 
> so we can in most places just call `self.create_mdk_for_account(account.pubkey)?`
> 
> what do you think?

Agreed

src/whitenoise/accounts.rs
mubarakcoded: @erskingardner your concern is valid, I removed create_mdk_for_tests and its test. All tests now go through the real create_mdk with mock keyring, which exercises the same encrypted storage code path as production. The only difference is an in-memory backend instead of the platform keyring. This way CI tests match production behavior as closely as possible.

src/whitenoise/accounts.rs
```diff
@@ -348,10 +348,11 @@ impl Account {
     pub(crate) fn create_mdk(
         pubkey: PublicKey,
         data_dir: &Path,
+        keyring_service_id: &str,
     ) -> core::result::Result<MDK<MdkSqliteStorage>, AccountError> {
         let mls_storage_dir = data_dir.join("mls").join(pubkey.to_hex());
         let db_key_id = format!("mdk.db.key.{}", pubkey.to_hex());
-        let storage = MdkSqliteStorage::new(mls_storage_dir, "com.whitenoise.app", &db_key_id)?;
+        let storage = MdkSqliteStorage::new(mls_storage_dir, keyring_service_id, &db_key_id)?;
         Ok(MDK::new(storage))
```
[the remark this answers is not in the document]
mubarakcoded: This is a diff rendering artifact, the two MdkSqliteStorage::new lines are the before/after, not two actual lines in the code. The current code correctly passes keyring_service_id.  Also with regards to NostrKeyring conventions, MdkSqliteStorage is from the mdk-sqlite-storage crate and handles its own keyring interactions internally using the service ID and entry name we provide. We don't construct NostrKeyring or Entry directly, that's MDK's internal concern. The naming is intentional and correct as-is.

src/whitenoise/event_processor/event_handlers/handle_giftwrap.rs
```diff
@@ -281,9 +287,10 @@ impl Whitenoise {
         account: &Account,
         data_dir: &std::path::Path,
         keys: Keys,
+        keyring_service_id: &str,
```
jgmontoya: Maybe instead of adding the extra param to the method we could access it from whitenoise?

`whitenoise.config.keyring_service_id`

src/whitenoise/accounts.rs
```diff
@@ -1392,7 +1393,14 @@ pub mod test_utils {
     }
 
     pub fn create_mdk(pubkey: PublicKey) -> MDK<MdkSqliteStorage> {
-        super::Account::create_mdk(pubkey, &data_dir()).unwrap()
+        // Initialize mock keyring store (only once per process)
+        use std::sync::OnceLock;
+        static MOCK_STORE_INIT: OnceLock<()> = OnceLock::new();
+        MOCK_STORE_INIT.get_or_init(|| {
+            keyring_core::set_default_store(keyring_core::mock::Store::new().unwrap());
+        });
```
jgmontoya: We should maybe move this to test_utils ,something like `init_mock_keyring` or something?

PR #475, opened 2026-02-10
refactor: unify keyring on keyring-core with platform-native stores

erskingardner:
## Summary

Fixes the startup crash where `mdk-sqlite-storage` fails with **"No default store has been set, so cannot search or create entries"** by initializing `keyring-core` with a platform-native credential store at startup. Also incorporates the `SecretsStore` refactor from #472, replacing the insecure XOR-obfuscated file fallback with a clean `keyring-core` implementation.

## Problem

Two separate keyring systems were in play:

- **`keyring` v3.6** (used by `SecretsStore`) — has built-in platform backends, works out of the box
- **`keyring-core` v0.7** (used by `mdk-sqlite-storage`) — requires an explicit `set_default_store()` call, which was never done in production

When `MdkSqliteStorage::new()` tried to create a keyring entry for the SQLCipher DB encryption key, it crashed because no credential store was registered.

## Changes

**`Cargo.toml`:**
- Removed `keyring` v3.6, `base64`, and `uuid` dependencies
- Made `keyring-core` a required (non-optional) dependency
- Added platform-native `keyring-core` store crates:
  - `apple-native-keyring-store` (macOS keychain / iOS protected data)
  - `linux-keyutils-keyring-store` (Linux kernel keyutils)
  - `windows-native-keyring-store` (Windows Credential Manager)
  - `android-native-keyring-store` (Android Keystore)

**`src/whitenoise/mod.rs`:**
- Added `initialize_keyring_store()` — registers the correct platform-native `keyring-core` store at startup (mock store in test/integration-test builds)
- Unified the separate `OnceLock` guards for platform and mock stores into a single one to prevent tests from accidentally touching the real platform keychain
- `initialize_mock_keyring_store()` is now a convenience alias

**`src/whitenoise/secrets_store.rs`:** (supersedes #472)
- Replaced the dual-path implementation (`keyring` for desktop + XOR-obfuscated JSON files for Android/test) with a single `keyring-core` implementation
- Uses configurable `service_name` from `WhitenoiseConfig.keyring_service_id`
- Better error handling: distinguishes `KeyringNotInitialized` from general `KeyringError`, maps `NoEntry` to `KeyNotFound`

## Key design decisions

- Both `SecretsStore` and `mdk-sqlite-storage` now share the same `keyring-core` default store — no namespace collision risk (MDK keys are prefixed `mdk.db.key.`, SecretsStore keys are bare hex pubkeys)
- Unsupported platforms produce a `compile_error!` rather than silently falling back to an in-memory mock
- Logout correctly removes only the Nostr secret key, not the MDK DB encryption key (deleting it would make the MLS database permanently unreadable)

## Testing

- `just precommit-quick` passes (868 tests, 0 failures)
- All existing `SecretsStore` and account tests pass with the mock keyring-core store

Closes #472

## Summary by CodeRabbit

* **Chores**
  * Switched to platform-native keyring implementations for macOS, iOS, Windows, Linux and Android.
  * Removed unused base64 and uuid dependencies; consolidated keyring dependency.

* **Refactor**
  * Secrets now persist in the system keyring via a service identifier instead of local files.
  * Simplified secret-storage error surface and removed file/encoding plumbing.

* **Bug Fixes**
  * Hardened initialization to ensure the credential store is available before use.

* **Tests**
  * Replaced heavy integration/network tests with focused unit tests; integration scenarios moved to the integration test suite.

### DIFF
--- a/src/whitenoise/secrets_store.rs
+++ b/src/whitenoise/secrets_store.rs
@@ -143,8 +143,9 @@ fn map_keyring_error(e: KeyringError) -> SecretsStoreError {
 
 #[cfg(test)]
 mod tests {
-    use super::*;
     use crate::whitenoise::Whitenoise;
+
+    use super::*;
 
     fn create_test_secrets_store() -> SecretsStore {
         Whitenoise::initialize_mock_keyring_store();


Cargo.toml
```diff
@@ -93,20 +86,33 @@ futures = "0.3"
...
+apple-native-keyring-store = { version = "0.2", features = ["keychain"] }
+
+[target.'cfg(target_os = "ios")'.dependencies]
+apple-native-keyring-store = { version = "0.2", features = ["protected"] }
+
+[target.'cfg(target_os = "windows")'.dependencies]
+windows-native-keyring-store = "0.5"
+
+[target.'cfg(target_os = "linux")'.dependencies]
+linux-keyutils-keyring-store = "0.2"
```
jgmontoya: I think there's an issue with this one. From the [docs](https://docs.rs/linux-keyutils-keyring-store/0.2.1/linux_keyutils_keyring_store/):


> The key management facility provided by the kernel is completely in-memory and will not persist across reboots. Consider the keyring a secure cache and plan for your application to handle cases where the entry is no longer available in-memory.

Issue #431, opened 2026-01-12
Implement KeyPackage lifecycle management with delayed cleanup


mubarakcoded:
## The Problem

We need to implement KeyPackage cleanup, but there's a timing issue:

**Scenario:**
1. User has KeyPackage KP1 on relay
2. Alice invites user to Group A using KP1
3. Bob invites user to Group B using KP1 (before user processes Alice's invite)
4. User processes Welcome A → needs KP1's key material → joins Group A
5. User processes Welcome B → needs KP1's key material → joins Group B
6. Current code: `rotate_key_package()` deletes KP1 from relay after first welcome
7. **Issue**: If we deleted KP1's local key material immediately, Welcome B would fail
8. **Need**: Keep key material around for some time window, then clean it up

## What Needs to Happen

whitenoise-rs needs to:
- Track which KeyPackages have been used to join groups
- Implement a retention policy (e.g., keep key material for 7 days)
- Clean up expired key material periodically
- Handle relay deletion for non-last-resort KeyPackages (via NIP-09)
- Preserve last-resort KeyPackages (can be reused per MIP-02)

## Why Here and Not MDK?

MDK can't determine when it's safe to clean up because it doesn't know if more welcomes are coming from relays or what retention policy makes sense. This needs application-level context that whitenoise-rs has.

MDK already provides everything needed - the KeyPackage event ID is available in the welcome's `e` tag for relay deletion.

---

PR #477, opened 2026-02-11
feat: add KeyPackage lifecycle management with delayed cleanup (#431)

mubarakcoded:
## Summary

- Track consumed KeyPackages via `consumed_key_packages` table with burst-aware cleanup (30s quiet period)
- On welcome processing, cache the KeyPackage's serialized `hash_ref` before relay deletion so local MLS key material can be cleaned up later
- Add Phase 2 to `key_package_maintenance` scheduled task to delete local key material for consumed KeyPackages after the quiet period elapses
- Bump MDK to `0.5.3` (feature branch `feature/key-package-hash-ref-deletion`, PR marmot-protocol/mdk#176) for `compute_key_package_hash_ref` and `delete_key_package_from_storage_by_hash_ref` methods
- Fix message cache tests to not assume positional ordering for same-second messages (exposed by MDK sort order change)

## Changed files

- **`db_migrations/0026_consumed_key_packages.sql`** — new migration with FK to `accounts`, unique constraint, and indexes
- **`src/whitenoise/database/consumed_key_packages.rs`** — new DB module: `track`, `find_eligible_for_cleanup`, `delete`, `delete_all_for_account` with 8 tests
- **`src/whitenoise/key_packages.rs`** — add `track_consumed_key_package` method
- **`src/whitenoise/scheduled_tasks/tasks/key_package_maintenance.rs`** — add Phase 2 cleanup logic with `CONSUMED_KP_QUIET_PERIOD_SECS`; eliminate redundant `Account::all()` query
- **`src/whitenoise/event_processor/event_handlers/handle_giftwrap.rs`** — call `track_consumed_key_package` before relay deletion in `rotate_key_package`
- **`src/whitenoise/messages.rs`** — fix 2 tests that assumed positional message ordering
- **`Cargo.toml`** — bump MDK deps to git rev `8357281`

## Test plan

- [x] All 854 unit tests pass
- [x] `cargo fmt -- --check` clean
- [x] `cargo clippy --all-targets` no warnings
- [x] Verify consumed key packages are tracked when processing welcomes (integration)
- [x] Verify local key material is cleaned up after 30s quiet period (integration)
- [ ] Update MDK rev once marmot-protocol/mdk#176 is merged to master

## Summary by CodeRabbit

* **New Features**
  * Automatic tracking and scheduled cleanup of consumed key packages with a short quiet period, improving local key material lifecycle and reducing stale data.

* **Chores**
  * Updated project dependencies and added database migration to support consumed-key-package tracking.


src/whitenoise/database/consumed_key_packages.rs
```diff
@@ -0,0 +1,332 @@
+use nostr_sdk::PublicKey;
+
+use super::{Database, DatabaseError};
+
+/// Represents a consumed key package tracked for delayed cleanup.
+#[derive(Debug, PartialEq, Eq, Clone, Hash)]
```
erskingardner: nit: can we move `Clone` to just after `Debug` (from the style guide)

src/whitenoise/message_aggregator/processor.rs
```diff
@@ -352,4 +352,180 @@ mod tests {
...
+
+        let mut del_tags = Tags::new();
+        del_tags.push(Tag::parse(vec!["e", "nonexistent"]).unwrap());
+        let del_msg = create_test_message(&keys, Kind::EventDeletion, del_tags, "");
+
+        assert!(!try_process_deletion(&del_msg, &mut processed));
+    }
+
+    fn create_test_media_file(original_hash: Option<Vec<u8>>) -> MediaFile {
+        use chrono::Utc;
```
erskingardner: can we move all these use statements to the top of the test mod?

src/whitenoise/message_aggregator/processor.rs
```diff
@@ -352,4 +352,180 @@ mod tests {
...
+            nostr_key: None,
+            file_metadata: None,
+            nonce: None,
+            scheme_version: None,
+            created_at: Utc::now(),
+        }
+    }
+
+    fn create_test_message(keys: &Keys, kind: Kind, tags: Tags, content: &str) -> Message {
+        use mdk_core::prelude::GroupId;
```
erskingardner: Here too.

src/whitenoise/scheduled_tasks/tasks/key_package_maintenance.rs
```diff
@@ -29,72 +35,59 @@ impl Task for KeyPackageMaintenance {
 
     async fn execute(&self, whitenoise: &'static Whitenoise) -> Result<(), WhitenoiseError> {
         tracing::debug!(
-            target: "whitenoise::scheduler::key_package_maintenance",
+            target: "whitenoise::scheduled_tasks::key_package_maintenance",
```
erskingardner: Can we leave these so they're consistent with the other scheduler tasks? That or we should rename the tracing targets to the same in the other scheduled_tasks/ files

src/whitenoise/database/consumed_key_packages.rs
```diff
@@ -0,0 +1,332 @@
...
+                    id,
+                    account_pubkey,
+                    key_package_hash_ref,
+                    consumed_at,
+                },
+            )
+            .collect())
+    }
+
+    /// Removes a consumed key package record after successful cleanup.
```
erskingardner: I'd remove "after a successful cleanup" - this module doesn't have (and doesn't need) any awareness about why/where this method will be used.

src/whitenoise/database/consumed_key_packages.rs
```diff
@@ -0,0 +1,332 @@
...
+
+        Ok(())
+    }
+
+    /// Returns all consumed key packages for an account that are eligible for cleanup.
+    ///
+    /// A package is eligible when all consumed packages for the account have a
+    /// `consumed_at` timestamp older than `quiet_period_secs` seconds ago.
+    /// This ensures we wait for the burst of welcomes to finish before cleaning up.
+    pub(crate) async fn find_eligible_for_cleanup(
```
erskingardner: I think you're coupling too much here. This method is fundamentally just getting `consumed_key_packages` where `consumed_at` is older than some arbitrary date. I think a better method name could be `where_consumed_at_lt` or `where_consumed_at_before`

src/whitenoise/database/consumed_key_packages.rs
```diff
@@ -0,0 +1,332 @@
...
+    pub key_package_hash_ref: Vec<u8>,
+    pub consumed_at: i64,
+}
+
+impl ConsumedKeyPackage {
+    /// Records a consumed key package for later cleanup.
+    ///
+    /// Uses INSERT OR REPLACE so that if the same key package is consumed by
+    /// multiple welcomes, the timestamp is updated to the latest consumption.
+    pub(crate) async fn track(
```
erskingardner: This should be `create` or `save` to be consistent with other database modules.

db_migrations/0027_published_key_packages.sql
```diff
@@ -0,0 +1,18 @@
+-- Track the full lifecycle of every published KeyPackage from creation through cleanup.
+-- Replaces the consumed_key_packages approach: instead of fetching the hash_ref
```
erskingardner: I'd remove these last two lines of the comment. Future us won't have any idea what this is referring to.

src/integration_tests/test_cases/scheduler/key_package_maintenance.rs
```diff
@@ -157,7 +157,7 @@ async fn publish_backdated_key_package(
     days_old: u64,
 ) -> Result<EventId, WhitenoiseError> {
     // Get the encoded key package and tags
-    let (encoded_key_package, tags) = context
+    let (encoded_key_package, tags, _hash_ref) = context
```
erskingardner, line 160: If possible, I think we should have some integration tests that go through the whole key_package lifecycle.

src/whitenoise/database/published_key_packages.rs
```diff
@@ -0,0 +1,617 @@
...
+    ) -> Result<(), DatabaseError> {
+        sqlx::query("UPDATE published_key_packages SET key_material_deleted = 1 WHERE id = ?")
+            .bind(id)
+            .execute(&database.pool)
+            .await?;
+        Ok(())
+    }
+
+    /// Removes all published key package records for an account.
+    #[allow(dead_code)]
```
erskingardner: why have this if we're not using it?

src/whitenoise/message_aggregator/processor.rs
```diff
@@ -352,4 +358,173 @@ mod tests {
         // Should handle gracefully - return None for malformed tags
         assert!(reply_to_id.is_none());
     }
+
+    #[test]
+    fn test_extract_media_hashes_valid_imeta() {
```
erskingardner, line 363: Is this stuff related to the key package work or just added to get some more test coverage? It's ok if it's just to get more coverage, just wanted to double check

mubarakcoded: This is not related to key package, I added it to recover the coverage

src/whitenoise/key_packages.rs
```diff
@@ -168,43 +210,57 @@ impl Whitenoise {
...
+                &account.pubkey,
+                &event_id.to_hex(),
+                &self.database,
+            )
+            .await
+            {
+                Ok(Some(pkg)) if !pkg.key_material_deleted => {
+                    let mdk = self.create_mdk_for_account(account.pubkey)?;
+                    mdk.delete_key_package_from_storage_by_hash_ref(&pkg.key_package_hash_ref)?;
+                }
```
jgmontoya, line 285: I think we're missing a call to `PublishedKeyPackage::mark_key_material_deleted` here

src/whitenoise/database/published_key_packages.rs
jgmontoya, line 1: Every other database module uses the `XxxRow` struct + manual `FromRow` impl +
`From<XxxRow>` conversion pattern. See `accounts_groups.rs:23-129`,
`cached_graph_users.rs:23-86`, `users.rs:27-74`, `media_files.rs:69-179`,
`relays.rs:50-79`. This module uses inline 7-element tuple destructuring repeated in two
methods.

This is:

- **Fragile:** Column reorder or addition requires updating multiple tuple patterns in lockstep.
- **Inconsistent:** The only database module not using the established pattern.
- **Repetitive:** The identical 7-field mapping appears at lines 72-91 and 152-173.

**Suggested fix:** Introduce a `PublishedKeyPackageRow` struct with manual `FromRow` impl
and `From<PublishedKeyPackageRow> for PublishedKeyPackage`, following the pattern in every
other module.

src/whitenoise/key_packages.rs
mubarakcoded: Good catch! added - Thank you.

src/whitenoise/database/published_key_packages.rs
mubarakcoded: Refactored to use PublishedKeyPackageRow with manual FromRow impl and From<PublishedKeyPackageRow> for PublishedKeyPackage, matching the existing pattern in the codebase. Thank you.

src/integration_tests/test_cases/scheduler/key_package_maintenance.rs
mubarakcoded: I added KeyPackageLifecycleTestCase that verifies the full lifecycle

src/integration_tests/test_cases/scheduler/key_package_lifecycle.rs
```diff
@@ -0,0 +1,159 @@
+use std::time::Duration;
+
+use crate::WhitenoiseError;
+use crate::integration_tests::core::*;
+use crate::integration_tests::test_cases::shared::{CreateGroupTestCase, WaitForWelcomeTestCase};
+use crate::whitenoise::scheduled_tasks::{KeyPackageMaintenance, Task};
+use async_trait::async_trait;
```
erskingardner: This should be before the `crate::` imports

src/integration_tests/test_cases/scheduler/key_package_lifecycle.rs
```diff
@@ -0,0 +1,159 @@
...
+        // ============================================================
+        // Phase 3: Run maintenance and verify cleanup
+        // ============================================================
+        tracing::info!("Waiting for quiet period before running maintenance cleanup...");
+
+        // The quiet period is 30 seconds in production. We need to wait for it
+        // to elapse so the maintenance task considers the KP eligible for cleanup.
+        // Use a shorter poll loop — the consumed_at was set moments ago, so we
+        // wait slightly beyond the quiet period.
+        tokio::time::sleep(Duration::from_secs(35)).await;
```
erskingardner: This is going to add 35 seconds to each run of the integration tests. we need to find a better way to test this without waiting the full time.

src/whitenoise/scheduled_tasks/tasks/key_package_maintenance.rs
jgmontoya, line 1: I think we should have two independent tasks for the different things we're doing. I don't think we need to couple both operations. Not a hard blocker on the PR though

mubarakcoded: Done. ConsumedKeyPackageCleanup is split into its own scheduled task.

db_migrations/0029_published_key_packages.sql
```diff
@@ -0,0 +1,16 @@
...
+    event_id TEXT NOT NULL,
+    consumed_at INTEGER,
+    key_material_deleted INTEGER NOT NULL DEFAULT 0,
+    created_at INTEGER NOT NULL DEFAULT (unixepoch()),
+    FOREIGN KEY (account_pubkey) REFERENCES accounts(pubkey) ON DELETE CASCADE,
+    UNIQUE(account_pubkey, event_id)
+);
+
+CREATE INDEX idx_published_kp_account ON published_key_packages(account_pubkey);
+CREATE INDEX idx_published_kp_event_id ON published_key_packages(event_id);
```
jgmontoya, line 16: Should we also add an index on consumed_at? or perhaps composite index:

```sql
CREATE INDEX idx_published_kp_cleanup ON published_key_packages(account_pubkey, consumed_at, key_material_deleted);
```

src/whitenoise/scheduled_tasks/tasks/consumed_key_package_cleanup.rs
```diff
@@ -0,0 +1,225 @@
...
+    let mut cleaned = 0usize;
+
+    for consumed in &eligible {
+        match mdk.delete_key_package_from_storage_by_hash_ref(&consumed.key_package_hash_ref) {
+            Ok(()) => {
+                if let Err(e) = PublishedKeyPackage::mark_key_material_deleted(
+                    consumed.id,
+                    &whitenoise.database,
+                )
+                .await
```
jgmontoya, line 146: If `mdk.delete_key_package_from_storage_by_hash_ref()` succeeds but `PublishedKeyPackage::mark_key_material_deleted()` fails, the database will be in an inconsistent state (key material is gone but the record still shows `key_material_deleted = false`).

Perhaps we should first mark_key_material_deleted and then do the mdk.delete..., if the mdk operation fails we can rollback the mark

src/whitenoise/key_packages.rs
```diff
@@ -168,43 +210,67 @@ impl Whitenoise {
...
+                &self.database,
+            )
+            .await
+            {
+                Ok(Some(pkg)) if !pkg.key_material_deleted => {
+                    let mdk = self.create_mdk_for_account(account.pubkey)?;
+                    mdk.delete_key_package_from_storage_by_hash_ref(&pkg.key_package_hash_ref)?;
+                    if let Err(e) =
+                        PublishedKeyPackage::mark_key_material_deleted(pkg.id, &self.database).await
+                    {
```
jgmontoya, line 277: Same here

src/whitenoise/scheduled_tasks/tasks/consumed_key_package_cleanup.rs
mubarakcoded: Good concern, but the inconsistent state is completely self-healing. The delete_key_package_from_storage_by_hash_ref on MDK is idempotent; calling it on already-deleted key material returns Ok(()). So in the delete-first order, if mark fails, the next scheduler run picks it up again (since key_material_deleted is still 0), calls delete again (no-op, returns Ok), and marks it successfully. 
The mark-first approach actually introduces a worse failure mode — if both the delete and the rollback(eg., unmark_key_material_deleted()) fail, the record says deleted but key material still exists, and it'll never be retried.
WDYT?

db_migrations/0029_published_key_packages.sql
mubarakcoded: Good call, I have added it.

src/whitenoise/key_packages.rs
```diff
@@ -698,6 +764,54 @@ impl Whitenoise {
...
+    #[cfg(feature = "integration-tests")]
+    pub async fn find_published_key_package_for_testing(
+        &self,
+        account_pubkey: &nostr_sdk::PublicKey,
+        event_id: &str,
+    ) -> Result<Option<PublishedKeyPackage>> {
+        PublishedKeyPackage::find_by_event_id(account_pubkey, event_id, &self.database)
+            .await
+            .map_err(|e| WhitenoiseError::Other(e.into()))
+    }
```
jgmontoya, line 751: We could make the `PublishedKeyPackage::find_by_event_id` part of the public interface, what do you think?

mubarakcoded: Making find_by_event_id public would leak the Database abstraction without adding value. If a production use case emerges, we can add a new public method on Whitenoise (without the feature gate) rather than widening the database layer's visibility, and that will be a bigger refactor than the current approach.

jgmontoya: I think it can be a useful method to have (proved by the fact that we found it useful ourselves for tests) but ok. I don't feel too strongly about this.

PR #482, opened 2026-02-12
perf: fix slow login for accounts with large follow lists (#480)

### LINKED ISSUE #480: Login became slower
**The bug**
After upgrading the rust crate on this PR https://github.com/marmot-protocol/whitenoise/pull/209 from commit `a75ebdf711cd59cf4214c3d0aba2627a5adaf489` to `614f9d6bfb8353afce9806db70b6212fb39ebfe7` login became slower. For my personal account it increased from aprox 2s to 6s. @erskingardner reports times way higher for his account (over 30s 🫠)

Maybe it's related to follow counts? My account is very small (I follow 150 npubs) but jeff has a lot more, and he follows over 1000 npubs.

**Screen recordings**
Here are some screenshot recordings to compare before and after the upgrade

Before (2s aprox)

https://github.com/user-attachments/assets/8ff7c819-44a7-427b-aa47-f930191451ee

After (6s aprox)

https://github.com/user-attachments/assets/4915e23a-8ea9-4d94-93fd-e5fc3132afee

erskingardner:
## Summary

Fixes #480 — login performance regression for accounts with large follow lists (750+ follows).

Addresses the core bottlenecks causing slow login by:

- **Parallelizing relay fetches** — Inbox and KeyPackage relay list fetches now run concurrently after NIP-65 completes, as do the conditional relay list publishes
- **Eliminating N+1 database query** — `all_users_with_relay_urls()` replaced with a single SQL JOIN instead of 1 query per user (1001 queries → 1 for 1000 follows)
- **Throttling background user fetches** — Contact list processing now uses `buffer_unordered(35)` instead of spawning 1000+ unbounded concurrent tasks that overwhelm the relay pool
- **Deduplicating subscription rebuilds** — `refresh_global_subscription_for_user()` called once after the entire contact list batch completes instead of once per user
- **Adding cancellation on logout** — Per-account `watch` channels signal background fetch tasks to stop when the account logs out, eliminating the flood of `can't send message to 'nostr' channel` errors

## Benchmark Results

Added a new `login-performance` benchmark (`just benchmark login-performance`) that simulates login with ~1000 follows, each with published metadata and relay lists.

| Metric | Before | After | Improvement |
|--------|--------|-------|-------------|
| **Mean** | 4.36s | **1.15s** | **-74%** |
| **Median** | 1.35s | **1.11s** | **-18%** |
| **Max** | 19.31s | **1.57s** | **-92%** |
| **P95** | 19.31s | **1.57s** | **-92%** |
| **Std Dev** | 6.46s | **0.19s** | **-97%** |
| **Throughput** | 0.21 ops/s | **0.61 ops/s** | **+190%** |

### New files
- `src/integration_tests/benchmarks/scenarios/login_performance.rs` — Benchmark scenario that pre-publishes metadata + relay lists for 1000 follow users to local test relays
- `src/integration_tests/benchmarks/test_cases/accounts/login_benchmark.rs` — Test case that times `login()` → `logout()` per iteration

### Modified files
- `src/whitenoise/accounts.rs` — Parallel relay fetches/publishes, cancellation channel lifecycle
- `src/whitenoise/database/users.rs` — New `all_with_nip65_relay_urls()` single JOIN query
- `src/whitenoise/users.rs` — Delegate to new JOIN query
- `src/whitenoise/event_processor/event_handlers/handle_contact_list.rs` — Bounded concurrency, single subscription rebuild, cancellation support
- `src/whitenoise/mod.rs` — `background_task_cancellation` DashMap on Whitenoise struct

## Future work

Additional optimizations identified but deferred to keep this PR focused:
- Defer/paginate `sync_message_cache_on_startup()` for accounts with many groups
- Parallelize `activate_account()` subscription + key package setup
- Cache MDK instances to avoid redundant keyring access

## Summary by CodeRabbit

* **New Features**
  * Added a login performance benchmark to measure login times with large follow lists.

* **Performance Improvements**
  * Relay setup and relay-list/key-package publishing run in parallel to speed initialization.
  * User data fetching moved to batched, bounded-concurrency processing with cancellation.
  * Database retrieval now returns users with related relay URLs in a single call.
  * Login short-circuits when an account is already active for faster responses.

* **Bug Fixes**
  * Background tasks are cancelled and cleaned up when accounts log out.

* **Tests**
  * Added tests covering login benchmark, account lifecycle, relay publishing, and user-relay retrieval.

### DIFF
--- a/src/whitenoise/accounts.rs
+++ b/src/whitenoise/accounts.rs
@@ -2578,6 +2578,32 @@ mod tests {
         );
     }
 
+    #[tokio::test]
+    async fn test_login_with_external_signer_double_login_returns_existing_account() {
+        let (whitenoise, _data_temp, _logs_temp) = create_mock_whitenoise().await;
+        let keys = Keys::generate();
+        let pubkey = keys.public_key();
+
+        // First login via test helper (sets up account in DB without needing relays)
+        let first_account = whitenoise
+            .login_with_external_signer_for_test(pubkey)
+            .await
+            .unwrap();
+
+        // Second login via the real method — should hit the double-login guard
+        // and return early without doing any relay work
+        let second_account = whitenoise
+            .login_with_external_signer(pubkey, keys)
+            .await
+            .unwrap();
+
+        assert_eq!(first_account.id, second_account.id);
+        assert_eq!(first_account.pubkey, second_account.pubkey);
+
+        let count = whitenoise.get_accounts_count().await.unwrap();
+        assert_eq!(count, 1, "Double login should not create a second account");
+    }
+
     /// Test that Account helper methods work correctly for external accounts
     #[tokio::test]
     async fn test_external_account_helper_methods() {
--- a/src/whitenoise/mod.rs
+++ b/src/whitenoise/mod.rs
@@ -2268,4 +2268,20 @@ mod tests {
             assert!(whitenoise.get_external_signer(&pubkey2).is_some());
         }
     }
+
+    mod subscription_tests {
+        use super::*;
+
+        #[tokio::test]
+        async fn test_refresh_all_global_subscriptions_no_account_returns_ok() {
+            let (whitenoise, _data_temp, _logs_temp) = create_mock_whitenoise().await;
+
+            // With no accounts in the database, should return Ok (early return)
+            let result = whitenoise.refresh_all_global_subscriptions().await;
+            assert!(
+                result.is_ok(),
+                "refresh_all_global_subscriptions should succeed with no accounts"
+            );
+        }
+    }
 }


src/whitenoise/accounts.rs
```diff
@@ -736,6 +742,13 @@ impl Whitenoise {
         inbox_relays: &[Relay],
         key_package_relays: &[Relay],
     ) -> Result<()> {
+        // Create a fresh cancellation channel for this account's background tasks.
+        // Any previous channel (from a prior login) is replaced, which also drops
+        // any stale receivers.
+        let (cancel_tx, _) = tokio::sync::watch::channel(false);
+        self.background_task_cancellation
+            .insert(account.pubkey, cancel_tx);
+
```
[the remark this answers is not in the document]
erskingardner: I'm unclear how a user would re-login without having first logged out?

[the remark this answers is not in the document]
erskingardner: OK - we're just going to return early with the Account when a double login happens and the pubkey is already logged in. This avoids the duplicate setup work.

src/whitenoise/database/users.rs
```diff
@@ -84,6 +87,55 @@ impl User {
...
+        database: &Database,
+    ) -> Result<Vec<(PublicKey, Vec<RelayUrl>)>, WhitenoiseError> {
+        let relay_type_str = String::from(RelayType::Nip65);
+
+        let rows: Vec<(String, Option<String>)> = sqlx::query_as(
+            "SELECT u.pubkey, r.url
+             FROM users u
+             LEFT JOIN user_relays ur ON u.id = ur.user_id AND ur.relay_type = ?
+             LEFT JOIN relays r ON ur.relay_id = r.id
+             ORDER BY u.pubkey",
```
jgmontoya, line 103: Do we actually need this ordering?

src/whitenoise/event_processor/event_handlers/handle_contact_list.rs
```diff
@@ -97,19 +105,150 @@ impl Whitenoise {
...
+            );
+
+            // Refresh the global subscription once after all users are processed,
+            // rather than once per user. We use a dummy user since the method
+            // rebuilds the subscription for all known users anyway.
+            if fetched > 0
+                && let Ok((user, _)) =
+                    User::find_or_create_by_pubkey(&pubkeys[0], &whitenoise.database).await
+                && let Err(e) = whitenoise.refresh_global_subscription_for_user(&user).await
             {
```
jgmontoya, line 168: I don't think this is accurate: "the method rebuilds the subscription for all known users anyway."

The method only does for the batches that contain the user. If discovered users land on different batches, some will not be properly refreshed.

src/whitenoise/accounts.rs
```diff
@@ -396,17 +396,31 @@ impl Whitenoise {
...
-        tracing::debug!(target: "whitenoise::login", "Logging in with pubkey: {}", pubkey.to_hex());
+        tracing::debug!(target: "whitenoise::accounts", "Logging in with pubkey: {}", pubkey.to_hex());
+
+        // If this account is already logged in, return it as-is. The session
+        // (relay connections, subscriptions, cancellation channel, background
+        // tasks) was set up during the original login and is still active —
+        // re-running activate_account would create duplicate subscriptions and
+        // needlessly kill in-progress background tasks.
+        if let Ok(existing) = Account::find_by_pubkey(&pubkey, &self.database).await {
+            tracing::debug!(
```
jgmontoya, line 407: Nice catch

src/whitenoise/database/users.rs
```diff
@@ -84,6 +87,55 @@ impl User {
...
+        }
+
+        let mut result = Vec::with_capacity(order.len());
+        for pubkey_hex in order {
+            let pubkey =
+                PublicKey::parse(&pubkey_hex).map_err(|e| WhitenoiseError::Other(e.into()))?;
+            let relay_urls = map.remove(&pubkey_hex).unwrap_or_default();
+            result.push((pubkey, relay_urls));
+        }
+
```
jgmontoya: My agent suggests this:

> 
> This is a manual ordered-grouping implementation. The SQL already uses `ORDER BY u.pubkey`, so the rows arrive pre-sorted. You can exploit this with a simpler approach that doesn't need a HashMap at all:
> 
> ```rust
> let mut result: Vec<(PublicKey, Vec<RelayUrl>)> = Vec::new();
> let mut current_pubkey: Option<String> = None;
> 
> for (pubkey_hex, relay_url) in rows {
>     if current_pubkey.as_ref() != Some(&pubkey_hex) {
>         let pk = PublicKey::parse(&pubkey_hex)
>             .map_err(|e| WhitenoiseError::Other(e.into()))?;
>         result.push((pk, Vec::new()));
>         current_pubkey = Some(pubkey_hex);
>     }
>     if let Some(url_str) = relay_url
>         && let Ok(url) = RelayUrl::parse(&url_str)
>     {
>         result.last_mut().unwrap().1.push(url);
>     }
> }
> ```
> 
> This eliminates the HashMap, the order Vec, and all the cloning of pubkey strings. Single pass, zero extra allocations beyond the result.

src/whitenoise/event_processor/event_handlers/handle_contact_list.rs
```diff
@@ -97,19 +105,150 @@ impl Whitenoise {
...
+    /// concurrency. Returns the number of users successfully fetched.
+    ///
+    /// Checks the cancellation signal before starting each user fetch and
+    /// stops early if cancelled (e.g. on logout).
+    async fn fetch_users_batch(
+        whitenoise: &Whitenoise,
+        pubkeys: &[PublicKey],
+        cancel_rx: Option<watch::Receiver<bool>>,
+    ) -> usize {
+        let cancel_rx = Arc::new(tokio::sync::Mutex::new(cancel_rx));
```
jgmontoya: Do we actually need the `tokio::sync::Mutex` wrap here?

src/integration_tests/benchmarks/scenarios/login_performance.rs
```diff
@@ -0,0 +1,216 @@
...
+            self.follow_count,
+            SETUP_CONCURRENCY
+        );
+
+        let follow_keys: Vec<Keys> = (0..self.follow_count).map(|_| Keys::generate()).collect();
+        let follow_pubkeys: Vec<PublicKey> = follow_keys.iter().map(|k| k.public_key()).collect();
+
+        // Publish metadata + relay lists for each follow user concurrently
+        let dev_relays = context.dev_relays.clone();
+        let relay_urls_for_follows = relay_urls.clone();
```
jgmontoya: Why do we need an additional `relay_urls_for_follow` __alias__ for `relay_urls`? Can't we just clone directly within the async move block?

src/whitenoise/database/users.rs
```diff
@@ -75,6 +77,7 @@ impl From<UserRow> for User {
 }
 
 impl User {
+    #[cfg_attr(not(test), allow(dead_code))]
```
jgmontoya: Perhaps this should just be `[cfg(test)]` ?

src/integration_tests/benchmarks/scenarios/login_performance.rs
```diff
@@ -0,0 +1,216 @@
...
+        let total_follows = self.follow_count;
```
jgmontoya: Do we actually need this variable?

src/integration_tests/benchmarks/scenarios/login_performance.rs
```diff
@@ -0,0 +1,216 @@
...
+                })
+                .collect()
+                .await;
+
+        if !errors.is_empty() {
+            tracing::warn!(
+                "Failed to publish data for {} follow users (continuing with partial setup)",
+                errors.len()
+            );
+        }
```
jgmontoya: We're capturing the index above (`i`) but we're only using the length of errors, we can probably simplify 

```rs
 Err(e) => (i, Some(e.to_string())),
```

to

```
 Err(e) => Some(e.to_string()),
```

src/whitenoise/database/users.rs
erskingardner: No - it's not needed. Removing.

src/whitenoise/accounts.rs
erskingardner: CodeRabbit 🐰

src/whitenoise/database/users.rs
erskingardner: Nice! I'll add this.

src/whitenoise/event_processor/event_handlers/handle_contact_list.rs
```diff
@@ -97,19 +105,150 @@ impl Whitenoise {
...
+                        if let Some(rx) = guard.as_ref()
+                            && *rx.borrow()
+                        {
+                            tracing::debug!(
+                                target: "whitenoise::handle_contact_list",
+                                "Background fetch cancelled, stopping"
+                            );
+                            return false;
+                        }
+                    }
```
jgmontoya, line 202: Check out this one:

> 
>   The current design checks cancellation inside each future. That
>   means the iterator keeps yielding all 1000 pubkeys into the stream,
>    and each one spawns a future that checks "am I cancelled?" at the
>   start. When cancelled, you still create and poll ~965 remaining
>   futures (each immediately returning false).
> 
>   With .take_while() before .buffer_unordered(), the iterator itself
>   stops yielding items:
> 
> ```rs
>   let cancel_rx_for_stream: Arc<Option<watch::Receiver<bool>>> =
>   Arc::new(cancel_rx);
>   let cancel_rx_for_check = cancel_rx_for_stream.clone();
> 
>   let results: Vec<bool> = stream::iter(pubkeys.iter().copied())
>       .take_while(move |_| {
>           let cancelled = cancel_rx_for_check
>               .as_ref()
>               .map(|rx| *rx.borrow())
>               .unwrap_or(false);
>           async move { !cancelled }
>       })
>       .map(|pubkey| {
>           let cancel_rx = cancel_rx_for_stream.clone();
>           async move {
>               // ... fetch user data (no cancellation check needed
>   here) ...
>               true
>           }
>       })
>       .buffer_unordered(CONTACT_LIST_FETCH_CONCURRENCY)
>       .collect()
>       .await;
> ```
> 
>   What changes:
>   - take_while runs synchronously on each item before it enters the
>   buffer
>   - The moment cancellation is signaled, the iterator produces no
>   more items
>   - The 35 already-in-flight fetches finish (they're already
>   running), then the stream ends
>   - No 965 spurious futures created, no cancellation checks inside
>   each future
> 
>   The difference is subtle but clean: cancellation logic lives in one
>    place (the stream adapter) instead of being duplicated inside
>   every future.

PR #486, opened 2026-02-14
fix: detect and retry failed key package relay publishes

### LINKED ISSUE #371: Rotation of key packages fails sometimes
Something I noticed today. I was getting lots of new conversation requests and then they stopped and I started getting people pinging me on other channels saying they were getting the invite screen when trying to message me. 

Turns out my key packages were gone from relays. looks like the app tried to rotate my key package at some point today and succeeded in deleting the current one but not publishing the replacement. I think we should implement some retry logic there to ensure that there is always a key package published for users.

mubarakcoded:
## Summary

- Key package publishes previously returned `Ok(())` even when no relay accepted the event, silently leaving accounts unable to receive group invites
- Added `result.success.is_empty()` check after every relay publish, returning `KeyPackagePublishFailed` error when no relay accepts
- Added retry with exponential backoff (3 attempts, 2s/4s delays) for the main `publish_key_package_for_account` path
- Separated key package creation (`encoded_key_package`) from relay publishing (`send_key_package_to_relays`) so retries reuse the same MLS key material instead of orphaning unused key packages in local storage

## Test plan

- [x] `just precommit-quick` passes (fmt, docs, clippy, unit tests)
- [x] New unit tests for `KeyPackagePublishFailed` error variant and external signer relay validation
- [x] All existing key package tests pass unchanged

Closes #371

## Summary by CodeRabbit

* **New Features**
  * Automatic retry with exponential backoff for publishing key packages, capped attempts to improve delivery reliability.
  * Safer key-package rotation: publishes replacement before cleanup and logs outcomes.

* **Bug Fixes**
  * Clearer publish-failure reporting and non-fatal handling when cleanup fails (scheduler will retry).

* **Tests**
  * Expanded coverage for missing relays, publish failures, retry flows, signer scenarios, and deletion outcomes.

* **Documentation**
  * Updated comments/docs reflecting retry, publish, and rotation behavior.

### DIFF
--- a/src/whitenoise/accounts.rs
+++ b/src/whitenoise/accounts.rs
@@ -873,9 +873,21 @@ impl Whitenoise {
                 .await?;
         }
         if key_package_event.is_none() {
-            self.publish_key_package_to_relays(account, key_package_relays)
-                .await?;
-            tracing::debug!(target: "whitenoise::accounts", "Published key package");
+            match self
+                .create_and_publish_key_package(account, key_package_relays)
+                .await
+            {
+                Ok(()) => {
+                    tracing::debug!(target: "whitenoise::accounts", "Published key package");
+                }
+                Err(e) => {
+                    tracing::warn!(
+                        target: "whitenoise::accounts",
+                        "Initial key package publish failed, scheduler will retry: {}",
+                        e
+                    );
+                }
+            }
         }
         Ok(())
     }
--- a/src/whitenoise/key_packages.rs
+++ b/src/whitenoise/key_packages.rs
@@ -1,10 +1,15 @@
+use std::sync::Arc;
+use std::time::Duration;
+
+use nostr_sdk::prelude::*;
+
 use crate::whitenoise::Whitenoise;
 use crate::whitenoise::accounts::Account;
 use crate::whitenoise::error::{Result, WhitenoiseError};
 use crate::whitenoise::relays::Relay;
-use nostr_sdk::prelude::*;
-use std::sync::Arc;
-use std::time::Duration;
+
+/// Maximum number of relay publish attempts before giving up.
+const MAX_PUBLISH_ATTEMPTS: u32 = 3;
 
 impl Whitenoise {
     /// Gets the appropriate signer for an account.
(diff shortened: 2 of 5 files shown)


src/whitenoise/key_packages.rs
```diff
@@ -81,36 +135,56 @@ impl Whitenoise {
...
+        let relay_urls = Relay::urls(&relays);
+        self.send_key_package_to_relays(&encoded_key_package, &relay_urls, &tags, signer)
+            .await
     }
 
+    /// Creates a new MLS key package for the account and publishes it to the given relays.
+    ///
+    /// This is a convenience wrapper that calls [`Self::encoded_key_package`] followed by
+    /// [`Self::send_key_package_to_relays`]. If you need retry semantics, prefer calling
+    /// those two methods separately so the key package is only created once.
```
jgmontoya, line 147: Perhaps rename the method to something like `create_and_publsh_key_package` ?

src/whitenoise/key_packages.rs
```diff
@@ -81,36 +135,56 @@ impl Whitenoise {
...
+        self.send_key_package_to_relays(&encoded_key_package, &relay_urls, &tags, signer)
+            .await
+    }
+
+    /// Publishes an already-encoded key package event to the given relays.
+    ///
+    /// Returns an error if no relay accepted the event. This method is
+    /// intentionally separated from key package creation so callers can retry
+    /// the relay publish without generating additional MLS key material.
+    async fn send_key_package_to_relays(
```
jgmontoya: And then make this one `publish_key_package_to_relays`

src/whitenoise/key_packages.rs
jgmontoya, line 1: Should we also add retries on `publish_key_package_for_account_with_signer` ?? cc. @dannym-arx

src/whitenoise/key_packages.rs
```diff
@@ -601,6 +675,26 @@ mod tests {
...
+            &RelayUrl::parse("wss://unreachable.test.relay").unwrap(),
+            &whitenoise.database,
+        )
+        .await
+        .unwrap();
+        user.add_relay(&relay, crate::RelayType::KeyPackage, &whitenoise.database)
+            .await
+            .unwrap();
+        (account, keys)
+    }
```
jgmontoya, line 697: The keys seem to never be used, we could simplify this to return just the account

src/whitenoise/key_packages.rs
mubarakcoded: I'd say no for this PR. It's called only in one place, i.e., login_with_external_signer, which is an interactive flow where the user is present and can retry the login if something goes wrong. The signer also takes impl NostrSigner + 'static which isn't cloneable, so adding retries would mean changing the public API to Arc<dyn NostrSigner>

src/whitenoise/key_packages.rs
```diff
@@ -81,36 +135,56 @@ impl Whitenoise {
...
+    /// Returns an error if no relay accepted the event. This method is
+    /// intentionally separated from key package creation so callers can retry
+    /// the relay publish without generating additional MLS key material.
+    async fn publish_key_package_to_relays(
+        &self,
+        encoded_key_package: &str,
+        relay_urls: &[RelayUrl],
+        tags: &[Tag],
+        signer: impl NostrSigner + 'static,
+    ) -> Result<()> {
```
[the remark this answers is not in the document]
mubarakcoded: this pattern (impl NostrSigner + 'static) is used 7 times in just this file and consistently throughout the existing codebase. Changing only our methods would be inconsistent with the rest of the file.

src/whitenoise/key_packages.rs
dannym-arx: I kinda agree with mubarak, should be a new pr

src/whitenoise/accounts.rs
```diff
@@ -873,9 +873,21 @@ impl Whitenoise {
                 .await?;
         }
         if key_package_event.is_none() {
-            self.publish_key_package_to_relays(account, key_package_relays)
-                .await?;
-            tracing::debug!(target: "whitenoise::accounts", "Published key package");
+            match self
```
erskingardner, line 876: This will pick up key packages published by any client won't it. That was fine when white noise was the only one around but I think we need ensure that we have a KP published by white noise here. That might be something we address in a follow up issue though.

src/whitenoise/key_packages.rs
erskingardner: agree too

src/whitenoise/accounts.rs
mubarakcoded: Yes, it will pick up key packages published by any client. Definitely worth a follow-up issue to address. I'll create one.

src/whitenoise/key_packages.rs
jgmontoya: Ok, makes sense to me 👍

PR #487, opened 2026-02-14
fix(ci): improve coverage reporting accuracy and reduce PR noise

erskingardner:
## Summary

Coverage reports have been noisy and unreliable. This PR improves coverage accuracy by excluding test infrastructure from metrics and improves the reporting UX to reduce PR noise.

## Changes

- **Exclude test infrastructure from coverage** — Added `--ignore-filename-regex` to filter out `integration_tests/`, `bin/integration_test.rs`, and `bin/benchmark_test.rs`. These files inflate the "lines found" denominator without reflecting library code quality.

- **Single updating PR comment with history** — Instead of creating a new comment on every push (spamming the PR), the workflow now finds and updates a single comment using an HTML marker. A collapsible "History" section tracks coverage across commits so you can see how it evolved.

- **Align local recipes with CI** — `just coverage` and `just coverage-html` now use the same flags as CI (`--all-targets`, `--ignore-filename-regex`) so local numbers match what CI reports.

## Example PR Comment

> ## :white_check_mark: Coverage: 42.57% → 42.80% (+0.23%)
>
> <details>
> <summary>History</summary>
>
> - `a1b2c3d` 2026-02-14 19:30:12 UTC — **42.60%** (+0.03% vs base)
> - `e4f5g6h` 2026-02-14 20:15:44 UTC — **42.75%** (+0.18% vs base)
> - `i7j8k9l` 2026-02-14 21:02:31 UTC — **42.80%** (+0.23% vs base)
>
> </details>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -239,29 +239,18 @@ jobs:
           CURRENT=${{ steps.coverage.outputs.percentage }}
           BASELINE=${{ steps.master_coverage.outputs.baseline }}
 
-          # Allow small fluctuations (up to 0.5%) to avoid false failures from
-          # non-deterministic LLVM instrumentation and async test timing.
-          THRESHOLD=0.5
-
           echo "=== Coverage Comparison ==="
           echo "Master branch: $BASELINE%"
           echo "This PR:       $CURRENT%"
-          echo "Threshold:     -${THRESHOLD}% allowed"
 
-          # Rounded values for display only
           DIFF=$(awk "BEGIN {printf \"%.2f\", $CURRENT - $BASELINE}")
-          DROP=$(awk "BEGIN {printf \"%.2f\", $BASELINE - $CURRENT}")
 
-          # Threshold checks use full-precision awk expressions to avoid
-          # rounding masking small regressions (e.g. 0.504% rounding to 0.50%).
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -231,42 +239,135 @@ jobs:
           CURRENT=${{ steps.coverage.outputs.percentage }}
           BASELINE=${{ steps.master_coverage.outputs.baseline }}
 
+          # Allow small fluctuations (up to 0.5%) to avoid false failures from
+          # non-deterministic LLVM instrumentation and async test timing.
+          THRESHOLD=0.5
+
```
jgmontoya: I don't like this. There's no way to tell if the drop was due to 'false failures from non-deterministic LLVM instrumentation and async test timing' or because there's new code that's not actually covered.

I'd remove this and continue dealing with those on a case-by-case basis.

.github/workflows/ci.yml
```diff
@@ -231,42 +239,135 @@ jobs:
...
+              '<summary>History</summary>',
+              '',
+              historyStart,
+              ...historyEntries,
+              historyEnd,
+              '',
+              '</details>',
+              '',
+              marker,
+            ].join('\n');
```
jgmontoya, line 319: The table and collapsible details is cool but I also liked our previous format of

```${baseline}% → ${current}% (${diff > 0 ? '+' : ''}${diff}%)```

Can we perhaps keep both?

PR #493, opened 2026-02-17
Feat/drafts

jgmontoya:
## Summary by CodeRabbit

* **New Features**
  * Message draft support: save, retrieve, and delete unsent messages.
  * Drafts scoped per account and group; one draft per account/group enforced.
  * Drafts can include reply references and media attachments, and track created/updated timestamps.
  * Draft updates preserve original creation time while updating modification time.

* **Tests**
  * End-to-end and integration tests covering save, load, update, deletion, and no-op deletion cases.

### DIFF
--- a/src/whitenoise/database/drafts.rs
+++ b/src/whitenoise/database/drafts.rs
@@ -4,8 +4,9 @@ use chrono::{DateTime, Utc};
 use mdk_core::prelude::GroupId;
 use nostr_sdk::{EventId, PublicKey};
 
-use super::{Database, utils::parse_timestamp};
+use super::{Database, DatabaseError, utils::parse_timestamp};
 use crate::whitenoise::drafts::Draft;
+use crate::whitenoise::error::WhitenoiseError;
 use crate::whitenoise::media_files::MediaFile;
 
 /// Internal database row representation for the drafts table.
@@ -106,7 +107,7 @@ impl Draft {
         reply_to_id: Option<&EventId>,
         media_attachments: &[MediaFile],
         database: &Database,
-    ) -> Result<Self, sqlx::Error> {
+    ) -> Result<Self, WhitenoiseError> {
         let now = Utc::now().timestamp_millis();
 
         let media_json =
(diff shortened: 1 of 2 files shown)


db_migrations/0028_add_drafts.sql
```diff
@@ -0,0 +1,19 @@
...
+        CHECK (reply_to_id IS NULL OR (length(reply_to_id) = 64 AND reply_to_id NOT GLOB '*[^0-9a-fA-F]*')),
+    media_attachments JSONB NOT NULL DEFAULT '[]',
+    created_at INTEGER NOT NULL,  -- Unix timestamp in milliseconds
+    updated_at INTEGER NOT NULL,  -- Unix timestamp in milliseconds
+    FOREIGN KEY (account_pubkey) REFERENCES accounts(pubkey) ON DELETE CASCADE,
+    FOREIGN KEY (mls_group_id) REFERENCES group_information(mls_group_id) ON DELETE CASCADE,
+    UNIQUE(account_pubkey, mls_group_id)
+);
+
+CREATE INDEX idx_drafts_group ON drafts(mls_group_id);
```
erskingardner, line 19: does this need a compound index with the `account_pubkey` as well?

src/whitenoise/database/drafts.rs
```diff
@@ -0,0 +1,389 @@
...
+    /// `media_attachments`, and `updated_at` are replaced while
+    /// `created_at` is preserved.
+    pub(crate) async fn save(
+        account_pubkey: &PublicKey,
+        mls_group_id: &GroupId,
+        content: &str,
+        reply_to_id: Option<&EventId>,
+        media_attachments: &[MediaFile],
+        database: &Database,
+    ) -> Result<Self, sqlx::Error> {
```
erskingardner: I think we want to return `Result<Self, WhitenoiseError>` here. That's the pattern that we have everywhere else in the code.

src/whitenoise/database/drafts.rs
```diff
@@ -0,0 +1,389 @@
...
+
+        Ok(row.into())
+    }
+
+    /// Fetches the draft for (account, group), returning `None` if absent.
+    pub(crate) async fn find(
+        account_pubkey: &PublicKey,
+        mls_group_id: &GroupId,
+        database: &Database,
+    ) -> Result<Option<Self>, sqlx::Error> {
```
erskingardner: Ah - seems like the wrong error type is on all the methods.

db_migrations/0028_add_drafts.sql
jgmontoya: I had that initially but it's redundant with the UNIQUE(account_pubkey, mls_group_id)

src/whitenoise/database/drafts.rs
jgmontoya: You're absolutely right!

...

busted hehe

jgmontoya: fixed

src/whitenoise/database/drafts.rs
jgmontoya: fixed

PR #505, opened 2026-02-19
fix: publish evolution events before merging pending commits, with retry

### LINKED ISSUE #500: Fix commit ordering: publish evolution event before merge_pending_commit
## Summary

Per MIP-03, evolution events should be published to relays **before** calling `merge_pending_commit()` locally. This ensures local MLS state only advances after confirming the relay accepted the event. If publishing fails with the current ordering, local state has already advanced to an epoch that no other group member knows about.

## Affected Methods

All three are in `src/whitenoise/groups.rs` and follow the same merge-then-publish pattern:

### 1. `add_members_to_group` (line ~483)
```rust
let update_result = mdk.add_members(group_id, &key_package_events)?;
mdk.merge_pending_commit(group_id)?;  // ← merges before publish
// ...
self.nostr.publish_event_to(evolution_event, ...).await?;
```
**Note:** This one is more complex because welcome rumors are generated during the commit and the local state may need to be consistent for the welcome fan-out. Needs careful analysis.

### 2. `remove_members_from_group` (line ~583)
```rust
let update_result = mdk.remove_members(group_id, &members)?;
mdk.merge_pending_commit(group_id)?;  // ← merges before publish
// ...
self.nostr.publish_event_to(evolution_event, ...).await?;
```

### 3. `update_group_data` (line ~613)
```rust
let update_result = mdk.update_group_data(group_id, group_data)?;
mdk.merge_pending_commit(group_id)?;  // ← merges before publish
// ...
self.nostr.publish_event_to(evolution_event, ...).await?;
```

## Correct Ordering

The new `perform_self_update` method (added in the self-update-after-join PR) demonstrates the correct pattern:

1. Call the MDK operation (e.g., `self_update`, `add_members`) to create the pending commit
2. Publish the `evolution_event` to relays
3. Only after relay acceptance, call `merge_pending_commit()` to advance local state

## Context

- MIP-03 specifies publish-then-merge ordering
- MDK's epoch snapshot rollback (MIP-03 commit race resolution) can recover from stranded epochs, but it's better to avoid them
- Each method should be fixed in a separate PR to keep changes reviewable

### LINKED ISSUE #503: Add retry logic for publish_event_to failures in MLS evolution event methods
## Problem

Currently, methods that publish MLS evolution events (`add_members_to_group`, `remove_members_from_group`, `update_group_data`, `leave_group`, `perform_self_update`) do not implement retry logic for `publish_event_to` failures.

When `publish_event_to` fails after a pending commit has been created (e.g., via `mdk.self_update()`), the commit remains persisted in local state with no automatic cleanup or abort API. This blocks subsequent MLS operations, which will fail with `OwnCommitPending` errors.

## Scope

To maintain consistency across the codebase, retry logic (with exponential backoff) should be added to all of these methods in a single PR:

- `add_members_to_group`
- `remove_members_from_group`
- `update_group_data`
- `leave_group`
- `perform_self_update`

## Context

- Raised by: @erskingardner
- Related PR: #501
- Discussion: https://github.com/marmot-protocol/whitenoise-rs/pull/501#discussion_r2823399786

## Next Steps

1. Design a consistent retry strategy (e.g., 3 attempts with exponential backoff)
2. Implement retry logic in all affected methods
3. Consider error handling for exhausted retries (e.g., logging, alerting)
4. Update tests to verify retry behavior

erskingardner:
## Summary

- **Fix MIP-03 commit ordering** (#500): Reorder `add_members_to_group`, `remove_members_from_group`, and `update_group_data` to publish the evolution event to relays *before* calling `merge_pending_commit()`. This prevents local MLS state from advancing to an epoch that no other group member knows about when publishing fails.
- **Add publish retry with exponential backoff** (#503): Introduce `publish_event_with_retry()` as the single entry-point for all MLS protocol event publishing (3 attempts, exponential backoff: 2s, 4s). All five affected methods now use this helper.

## Detail

Per MIP-03, the correct ordering for MLS state evolution is:

1. Create the MDK operation (pending commit)
2. Publish the evolution event to relays
3. Only after relay acceptance, merge the pending commit locally

The previous code merged before publishing in three methods, which meant a publish failure would leave local state out of sync with the rest of the group.

### Methods changed

| Method | File | Ordering fix | Retry added |
|--------|------|:---:|:---:|
| `add_members_to_group` | `groups.rs` | Yes | Yes |
| `remove_members_from_group` | `groups.rs` | Yes | Yes |
| `update_group_data` | `groups.rs` | Yes | Yes |
| `leave_group` | `groups.rs` | N/A (proposal, no merge) | Yes |
| `perform_self_update` | `handle_giftwrap.rs` | N/A (already correct) | Yes |

### Retry helper

`publish_event_with_retry` is a `pub(crate)` method on `Whitenoise` designed as a clean seam for future replacement by a durable publish queue/ledger system. The key package publish retry in `key_packages.rs` is left as-is because it uses a different publish path (event builder + signer) and needs the returned `EventId` for lifecycle tracking.

## Testing

- `just precommit-quick` passes (fmt, docs, clippy, unit tests)
- Integration tests should be run with Docker (`just precommit`) to verify end-to-end

Closes #500, closes #503

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of group member and data changes by requiring relay acceptance before applying local state; added automatic retry with exponential backoff for event publishing.
* **Tests**
  * Added tests validating publish-retry behavior and ensuring failed publishes do not advance local group state.

### DIFF
--- a/src/whitenoise/groups.rs
+++ b/src/whitenoise/groups.rs
@@ -206,6 +206,41 @@ impl Whitenoise {
         Err(last_error.expect("loop ran at least once"))
     }
 
+    /// Publishes an evolution event and merges the pending commit on success.
+    ///
+    /// Per MIP-03 this is the canonical ordering for MLS state evolution:
+    /// 1. Caller creates the pending commit via an MDK operation
+    /// 2. This method publishes the evolution event (with retry)
+    /// 3. Only after at least one relay accepts, the pending commit is merged
+    ///
+    /// # Dangling pending commits on failure
+    ///
+    /// If all publish attempts fail, the pending commit created by the MDK
+    /// operation remains in the MLS group's internal state. MDK does not
+    /// currently expose `clear_pending_commit` (OpenMLS has it, but MDK
+    /// does not wrap it). This means subsequent MLS operations on the same
+    /// group may fail until the pending commit is resolved. Tracked
+    /// upstream — once MDK exposes a clear/abort API, the error path here
+    /// should call it.
+    // TODO(mdk): call mdk.clear_pending_commit(group_id) on failure once
+    // the MDK exposes it, so a failed publish doesn't block the group.
+    pub(crate) async fn publish_and_merge_commit(
+        &self,
+        evolution_event: Event,
+        account_pubkey: &PublicKey,
+        group_id: &GroupId,
+        relay_urls: &[RelayUrl],
+    ) -> Result<()> {
+        self.publish_event_with_retry(evolution_event, account_pubkey, relay_urls)
+            .await?;
+
+        // Relay accepted — now safe to advance local MLS state
+        let mdk = self.create_mdk_for_account(*account_pubkey)?;
+        mdk.merge_pending_commit(group_id)?;
+
+        Ok(())
+    }
+
     /// Creates a new MLS group with the specified members and settings
     ///
     /// # Arguments
@@ -573,19 +608,10 @@ impl Whitenoise {
             )));
         }
 
-        // Per MIP-03, publish the evolution event to relays *before* merging
-        // the pending commit. This ensures local state only advances after
-        // confirming the relay accepted the event.
-        self.publish_event_with_retry(evolution_event, &account.pubkey, &relay_urls)
+        self.publish_and_merge_commit(evolution_event, &account.pubkey, group_id, &relay_urls)
             .await?;
 
-        // Relay accepted the event — now safe to advance local MLS state
-        {
-            let mdk = self.create_mdk_for_account(account.pubkey)?;
-            mdk.merge_pending_commit(group_id)?;
-        }
-
-        // Evolution event published successfully
+        // Evolution event published and commit merged successfully
         // Fan out the welcome message to all members
         for (welcome_rumor, user) in welcome_rumors.iter().zip(users) {
             // Get the public key of the member from the key package event
@@ -665,16 +691,8 @@ impl Whitenoise {
             (relay_urls, update_result.evolution_event)
         };
 
-        self.publish_event_with_retry(evolution_event, &account.pubkey, &relay_urls)
-            .await?;
-
-        // Relay accepted the event — now safe to advance local MLS state
-        {
-            let mdk = self.create_mdk_for_account(account.pubkey)?;
-            mdk.merge_pending_commit(group_id)?;
-        }
-
-        Ok(())
+        self.publish_and_merge_commit(evolution_event, &account.pubkey, group_id, &relay_urls)
+            .await
     }
 
     /// Updates group metadata and publishes the change to group relays.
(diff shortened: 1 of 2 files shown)


src/whitenoise/groups.rs
```diff
@@ -2920,4 +3017,212 @@ mod tests {
...
+            members_before.len(),
+            members_after.len(),
+            "Member count should be unchanged when publish fails \
+             (pending commit must not be merged)"
+        );
+        assert!(
+            !members_after.contains(&new_pk),
+            "New member should not appear in the group"
+        );
+    }
```
jgmontoya, line 3165: From my agent:

> These tests manually replicate the production flow step-by-step, then assert that state didn't change because `merge_pending_commit` was never called. But this is self-evident — of course state doesn't change if you don't call merge. What we actually need to verify is that `add_members_to_group()` itself doesn't merge when publish fails.
> 
> The tests should call the actual production method (`add_members_to_group`, `remove_members_from_group`, `update_group_data`) with unreachable relays and verify the group state is unchanged. This would catch regressions if someone reorders the code in the future. As written, these tests will pass regardless of the publish/merge ordering in production code — they test their own local logic, not the system's.
> 
> **Suggestion:** Rewrite to call the actual production methods. You'd need a way to make the group's relay URLs point to unreachable ports. Perhaps override `ensure_group_relays` or create a group with unreachable relay URLs in its config.

src/whitenoise/groups.rs
```diff
@@ -504,10 +573,18 @@ impl Whitenoise {
             )));
         }
 
-        self.nostr
-            .publish_event_to(evolution_event, &account.pubkey, &relay_urls)
+        // Per MIP-03, publish the evolution event to relays *before* merging
+        // the pending commit. This ensures local state only advances after
+        // confirming the relay accepted the event.
+        self.publish_event_with_retry(evolution_event, &account.pubkey, &relay_urls)
```
jgmontoya: From my agent:

> When `mdk.add_members()` (or `remove_members`, `update_group_data`, `self_update`) is called, the MDK creates a pending commit inside the MLS group. If `publish_event_with_retry` then fails after all retries, the function returns `Err` — but the pending commit remains in the MLS group's internal state.
> 
> This is actually a regression from the old code. The old code merged immediately, so there was never a dangling pending commit. The new code can leave the MLS group in a state where a pending commit exists but was never merged. Depending on how the MLS library handles this, subsequent operations on the group could:
> 
> - Fail because a pending commit already exists
> - Silently overwrite the pending commit (data loss of the operation)
> - Behave unpredictably
> 
> **Suggestion:** Investigate whether `mdk` exposes a `clear_pending_commit()` or equivalent. If so, call it in the error path. If not, this may need to be added to `mdk-core`. At minimum, document this known limitation with a `// TODO` so it's tracked.

src/whitenoise/groups.rs
jgmontoya, line 1: From my agent:

> The same structural pattern repeats four times:
> 
> ```rust
> let (relay_urls, evolution_event, ...) = {
>     let mdk = self.create_mdk_for_account(account.pubkey)?;
>     let relay_urls = Self::ensure_group_relays(&mdk, group_id)?;
>     let update_result = mdk.some_operation(group_id, ...)?;
>     (relay_urls, update_result.evolution_event, ...)
> };
> 
> self.publish_event_with_retry(evolution_event, &account.pubkey, &relay_urls).await?;
> 
> {
>     let mdk = self.create_mdk_for_account(account.pubkey)?;
>     mdk.merge_pending_commit(group_id)?;
> }
> ```
> 
> The merge-after-publish half is identical in all four call sites. Consider extracting at least the post-publish merge (and future rollback-on-error) into a helper like `publish_and_merge_commit()`. This would:
> 
> - Ensure the rollback fix from issue #1 is applied consistently
> - Make the invariant ("publish before merge") structural rather than convention-based
> - Reduce the surface area for ordering bugs
> 
> `add_members_to_group` is slightly different (it has welcome_rumors), but the post-publish merge + the publish call itself could still be shared.
>

PR #515, opened 2026-02-23
fix(login): block login when any of 10002/10050/10051 relay lists are missing

erskingardner:
## Summary

- **Bug fix**: A user with a 10002 (NIP-65) relay list but no 10050 (InboxRelays) or 10051 (MlsKeyPackageRelays) would silently complete login. `try_discover_relay_lists` fell back to default relays for any missing 10050/10051, so `LoginStatus::Complete` was always returned as long as 10002 was present.
- **Selective publish**: `login_publish_default_relays` (and the external-signer equivalent) now reads the partial discovery results stashed during step 1 and only publishes/assigns defaults for the kinds that were actually missing — existing lists are preserved unchanged.
- **25 new tests** covering all 7 partial-discovery combinations, `pending_logins` stash mechanics, selective publish behavior for both local-key and external-signer paths, cancel idempotency, and an explicit regression test for the original bug.

### `src/whitenoise/accounts.rs`

- `DiscoveredRelayLists` now carries empty `Vec`s for lists not found on the network and exposes `is_complete()` to check all three are present.
- `try_discover_relay_lists` always returns `DiscoveredRelayLists` (no longer `Option`); callers branch on `is_complete()`. When 10002 is absent, 10050/10051 are still searched using the original source relays so partial state is captured accurately.
- All four login-start methods (`login_start`, `login_with_custom_relay`, `login_external_signer_start`, `login_external_signer_with_custom_relay`) store the full `DiscoveredRelayLists` in `pending_logins` on incomplete discovery.
- `login_publish_default_relays` and `login_external_signer_publish_default_relays` retrieve the stash and only publish+assign defaults for the missing kinds; found lists are persisted to the DB without re-publishing.

### `src/whitenoise/mod.rs`

- `pending_logins` changed from `DashSet<PublicKey>` to `DashMap<PublicKey, DiscoveredRelayLists>` so partial discovery results survive from step 1 into step 2a.

## Testing

All new tests are pure unit tests or use direct state injection into `pending_logins` to avoid relay-network round trips, keeping them fast and deterministic. Tests that require Docker relays (the existing Docker-dependent suite) are unchanged.

## Summary by CodeRabbit

* **New Features**
  * Added DiscoveredRelayLists (with completeness check, merge, and relay/fallback accessors) to track discovered relays.

* **Refactor**
  * Login flow now preserves and merges per-session partial relay discoveries and only publishes defaults for missing lists.
  * Pending-login state now stores per-key discovered-relay data to support selective publishing.

* **Tests**
  * Expanded tests covering completeness, merging/stashing, selective publishing, and various login flows.

### DIFF
--- a/src/whitenoise/accounts.rs
+++ b/src/whitenoise/accounts.rs
@@ -639,6 +639,10 @@ impl Whitenoise {
 
     // -----------------------------------------------------------------------
     // Multi-step login API
+    //
+    // Callers must not invoke step-2 methods concurrently for the same pubkey.
+    // The pending-login stash is not guarded beyond DashMap's per-key lock, so
+    // overlapping calls can race between the merge and complete_login phases.
     // -----------------------------------------------------------------------
 
     /// Step 1 of the multi-step login flow.
@@ -867,19 +871,10 @@ impl Whitenoise {
         //     that try_discover_relay_lists already wrote to the DB; and
         // (b) the is_complete check below correctly accounts for lists found
         //     across multiple relays (e.g. nip65 from login_start, inbox+kp here).
-        let mut stash = self
-            .pending_logins
-            .get_mut(pubkey)
-            .ok_or(LoginError::NoLoginInProgress)?;
-        stash.merge(discovered);
-        let is_complete = stash.is_complete();
-        let merged_nip65 = stash.nip65.clone();
-        let merged_inbox = stash.inbox.clone();
-        let merged_key_package = stash.key_package.clone();
-        drop(stash);
+        let merged = self.merge_into_stash(pubkey, discovered)?;
 
-        if is_complete {
-            self.complete_login(&account, &merged_nip65, &merged_inbox, &merged_key_package)
+        if merged.is_complete() {
+            self.complete_login(&account, &merged.nip65, &merged.inbox, &merged.key_package)
                 .await?;
             self.pending_logins.remove(pubkey);
             tracing::info!(
@@ -898,9 +893,9 @@ impl Whitenoise {
                 "Relay lists still incomplete after {} for {} (nip65={}, inbox={}, key_package={})",
                 relay_url,
                 pubkey.to_hex(),
-                !merged_nip65.is_empty(),
-                !merged_inbox.is_empty(),
-                !merged_key_package.is_empty(),
+                !merged.nip65.is_empty(),
+                !merged.inbox.is_empty(),
+                !merged.key_package.is_empty(),
             );
             Ok(LoginResult {
                 account,
@@ -1165,23 +1160,14 @@ impl Whitenoise {
         // Same upfront-merge pattern as login_with_custom_relay: update the
         // stash before attempting completion so a failed complete_external_signer_login
         // leaves the stash accurate for any subsequent retry.
-        let mut stash = self
-            .pending_logins
-            .get_mut(pubkey)
-            .ok_or(LoginError::NoLoginInProgress)?;
-        stash.merge(discovered);
-        let is_complete = stash.is_complete();
-        let merged_nip65 = stash.nip65.clone();
-        let merged_inbox = stash.inbox.clone();
-        let merged_key_package = stash.key_package.clone();
-        drop(stash);
+        let merged = self.merge_into_stash(pubkey, discovered)?;
 
-        if is_complete {
+        if merged.is_complete() {
             self.complete_external_signer_login(
                 &account,
-                &merged_nip65,
-                &merged_inbox,
-                &merged_key_package,
+                &merged.nip65,
+                &merged.inbox,
+                &merged.key_package,
                 signer,
             )
             .await?;
@@ -1202,9 +1188,9 @@ impl Whitenoise {
                 "Relay lists still incomplete after {} for {} (nip65={}, inbox={}, key_package={})",
                 relay_url,
                 pubkey.to_hex(),
-                !merged_nip65.is_empty(),
-                !merged_inbox.is_empty(),
-                !merged_key_package.is_empty(),
+                !merged.nip65.is_empty(),
+                !merged.inbox.is_empty(),
+                !merged.key_package.is_empty(),
             );
             Ok(LoginResult {
                 account,
@@ -1283,6 +1269,24 @@ impl Whitenoise {
             inbox: inbox_relays,
             key_package: key_package_relays,
         })
+    }
+
+    /// Merge newly-discovered relay lists into the pending-login stash and
+    /// return a snapshot.  The DashMap lock is released before returning so
+    /// the caller is free to do async work with the result.
+    fn merge_into_stash(
+        &self,
+        pubkey: &PublicKey,
+        discovered: DiscoveredRelayLists,
+    ) -> core::result::Result<DiscoveredRelayLists, LoginError> {
+        let mut stash = self
+            .pending_logins
+            .get_mut(pubkey)
+            .ok_or(LoginError::NoLoginInProgress)?;
+        stash.merge(discovered);
+        let snapshot = stash.clone();
+        drop(stash);
+        Ok(snapshot)
     }
 
     /// Activate a local-key account after relay lists have been resolved.


src/whitenoise/accounts.rs
```diff
@@ -702,46 +757,55 @@ impl Whitenoise {
...
+                // account.nip65_relays() etc. after login.
+                tracing::debug!(
+                    target: "whitenoise::accounts",
+                    "Skipping publish for {:?} — already exists on network; persisting to DB",
+                    relay_type,
+                );
+                user.add_relays(found, relay_type, &self.database)
+                    .await
+                    .map_err(LoginError::from)?;
+            }
```
jgmontoya, line 798: On step 1 we call `try_discover_relay_lists` which already does `user.add_relays`, those calls seem redundant here

src/whitenoise/accounts.rs
```diff
@@ -1078,82 +1198,72 @@ impl Whitenoise {
     /// NIP-65 relays to fetch Inbox and KeyPackage lists. Saves all found
     /// relays to the database.
     ///
-    /// Returns `Ok(Some((nip65, inbox, key_package)))` when the NIP-65 list is
-    /// found. Inbox and KeyPackage fall back to defaults if not found
-    /// independently.
+    /// Attempt to fetch all three relay lists from the network.
```
jgmontoya, line 1206: Should this replace the comment above it? This doc comment looks weird

src/whitenoise/accounts.rs
```diff
@@ -790,37 +854,50 @@ impl Whitenoise {
...
+                relay_url,
+                pubkey.to_hex(),
+                !discovered.nip65.is_empty(),
+                !discovered.inbox.is_empty(),
+                !discovered.key_package.is_empty(),
+            );
+            let mut stash = self
+                .pending_logins
+                .get_mut(pubkey)
+                .expect("pending_logins entry must exist when login_with_custom_relay is called");
```
jgmontoya: Perhaps instead of `expect` we can do `.ok_or(LoginError::NoLoginInProgress)?` ? not 100% convinced though

src/whitenoise/accounts.rs
```diff
@@ -1028,43 +1140,51 @@ impl Whitenoise {
...
+                relay_url,
+                pubkey.to_hex(),
+                !discovered.nip65.is_empty(),
+                !discovered.inbox.is_empty(),
+                !discovered.key_package.is_empty(),
+            );
+            let mut stash = self.pending_logins.get_mut(pubkey).expect(
+                "pending_logins entry must exist when login_external_signer_with_custom_relay \
+                     is called",
+            );
```
jgmontoya: same here

src/whitenoise/accounts.rs
```diff
@@ -702,46 +757,55 @@ impl Whitenoise {
...
+        let publish_to_relays = discovered
+            .relays_or(RelayType::Nip65, &default_relays)
+            .to_vec();
+
+        // For each relay type: persist found relays to the DB and publish
+        // defaults only for the ones that were missing.
         for relay_type in [RelayType::Nip65, RelayType::Inbox, RelayType::KeyPackage] {
-            user.add_relays(&default_relays, relay_type, &self.database)
+            let found = discovered.relays_or(relay_type, &[]);
+            if found.is_empty() {
```
jgmontoya: It's a bit odd to call `relays_or` with an empty fallback and then check `is_empty` right away, couldn't we check directly without doing `relays_or`?

Maybe defina a `relays` method:

```rs
pub fn relays(&self, relay_type: RelayType) -> &[Relay] {
    match relay_type {
        RelayType::Nip65 => &self.nip65,
        RelayType::Inbox => &self.inbox,
        RelayType::KeyPackage => &self.key_package,
    }
}
```

then modify the `relays_or` such that it uses it:

```rs
/// Returns the relays for `relay_type` when non-empty, otherwise `fallback`.
pub fn relays_or<'a>(&'a self, relay_type: RelayType, fallback: &'a [Relay]) -> &'a [Relay] {
    let found = self.relays(relay_type);
    if found.is_empty() { fallback } else { found }
}
```

Then this becomes more natural:

```rs
let found = discovered.relays(relay_type);
if found.is_empty() {
```

src/whitenoise/accounts.rs
```diff
@@ -4855,4 +4994,1036 @@ mod tests {
...
+
+        let stashed = pending.get(&pubkey).unwrap();
+        assert!(
+            !stashed.is_complete(),
+            "Stashed partial must not be complete"
+        );
+        assert_eq!(stashed.nip65.len(), 1);
+        assert!(stashed.inbox.is_empty());
+        assert!(stashed.key_package.is_empty());
+    }
```
jgmontoya, line 5296: According to my agent:

> `test_pending_logins_stash_is_incomplete_when_only_nip65_found` is redundant
> 
> **File:** `src/whitenoise/accounts.rs:5139-5175`
> 
> This test creates a `DiscoveredRelayLists` with only nip65, checks `is_complete()` is false, inserts it into a local `DashMap`, retrieves it, and checks `is_complete()` again. But:
> 
> - The `is_complete()` behavior is already exhaustively covered by the 7 `is_complete_*` tests (lines 5013-5112)
> - The `DashMap` round-trip doesn't test any application logic — it tests that `DashMap::get` returns what `DashMap::insert` stored
> - The comments say "This is a pure unit test with no relay or DB involvement" — but it's testing `DashMap`, not our code
> 
> **Suggestion:** Remove this test. Its assertions are fully covered by `test_discovered_relay_lists_is_complete_only_nip65` (line 5036) and `test_regression_nip65_only_user_is_blocked_from_login` (line 5962).

src/whitenoise/accounts.rs
```diff
@@ -4855,4 +4994,1036 @@ mod tests {
...
+
+        // Set up the external-signer account path (no relays pre-discovered).
+        let (account, _) = whitenoise
+            .setup_external_signer_account_without_relays(pubkey)
+            .await
+            .unwrap();
+        whitenoise
+            .insert_external_signer(pubkey, keys.clone())
+            .await
+            .unwrap();
```
jgmontoya: Non blocker but this might be worth it to extract to a `setup_partial_pending_login_external_signer` helper

src/whitenoise/accounts.rs
jgmontoya, line 1: From my agent:

> 
> Stale doc comment on `login_with_custom_relay`
> 
> **File:** `src/whitenoise/accounts.rs:822-830`
> 
> ```rust
> /// Step 2b: search a user-provided relay for existing relay lists.
> ///
> /// Called after [`Whitenoise::login_start`] returned [`LoginStatus::NeedsRelayLists`].
> /// Connects to `relay_url`, fetches the NIP-65 list, and (if found) fetches
> /// Inbox and KeyPackage lists from the discovered NIP-65 relays.
> ///
> /// If relay lists are found the account is activated and
> /// [`LoginStatus::Complete`] is returned. If nothing is found,
> /// [`LoginStatus::NeedsRelayLists`] is returned so the caller can re-prompt.
> ```
> 
> The doc says "fetches the NIP-65 list, and (if found) fetches Inbox and KeyPackage lists from the discovered NIP-65 relays." This describes the _old_ behavior where 10050/10051 were only fetched when 10002 was present. The new `try_discover_relay_lists` always attempts all three (falling back to `source_relays` when 10002 is absent). The doc should reflect this: all three lists are always attempted.

src/whitenoise/accounts.rs
```diff
@@ -790,37 +855,74 @@ impl Whitenoise {
...
-        match self
+        let discovered = self
             .try_discover_relay_lists(&mut account, &source_relays)
-            .await?
-        {
-            Some(relays) => {
-                self.complete_login(&account, &relays.nip65, &relays.inbox, &relays.key_package)
+            .await?;
+
+        if discovered.is_complete() {
```
mubarakcoded: Can we update the stash before attempting completion here? If complete_login fails, the stash still holds the old partial from login_start, so a retry via login_publish_default_relays would publish defaults over relays that try_discover_relay_lists already found and wrote to the DB.

jgmontoya: nice catch

src/whitenoise/accounts.rs
```diff
@@ -790,37 +855,57 @@ impl Whitenoise {
...
+        let mut stash = self
+            .pending_logins
+            .get_mut(pubkey)
+            .ok_or(LoginError::NoLoginInProgress)?;
+        stash.merge(discovered);
+        let is_complete = stash.is_complete();
+        let merged_nip65 = stash.nip65.clone();
+        let merged_inbox = stash.inbox.clone();
+        let merged_key_package = stash.key_package.clone();
+        drop(stash);
```
jgmontoya: From my agent:

> While DashMap provides interior mutability, the sequence "get_mut → merge → clone fields → drop → complete_login → remove" creates a window where another concurrent call to `login_with_custom_relay` for the same pubkey could interleave. Consider this race:
> 
> 1. Thread A: `get_mut`, `merge`, `clone`, `drop(stash)` → releases lock
> 2. Thread B: `get_mut`, `merge`, `clone`, `drop(stash)` → releases lock
> 3. Thread A: `complete_login` succeeds, `remove(pubkey)`
> 4. Thread B: `complete_login` (same account, may fail or create duplicate state)
> 
> **Recommendation:** Either:
> 
> - Document that the public API methods (`login_with_custom_relay`, etc.) must not be called concurrently for the same pubkey (seems reasonable — who would tap "login" twice?), OR
> - Use a per-pubkey lock/semaphore to serialize the entire "merge → check → complete → remove" sequence
> 
> The same issue exists in `login_external_signer_with_custom_relay` (lines 1158-1191).

I think both recommendations are reasonable, documenting or adding some lock/semaphore

jgmontoya: This can be extracted to a method, something like 

```rs
fn merge_and_snapshot_stash(
    &self,
    pubkey: &PublicKey,
    discovered: DiscoveredRelayLists,
) -> Result<(bool, Vec<Relay>, Vec<Relay>, Vec<Relay>), LoginError> {
    let mut stash = self.pending_logins.get_mut(pubkey).ok_or(...)?;
    stash.merge(discovered);
    let snapshot = (
        stash.is_complete(),
        stash.nip65.clone(),
        stash.inbox.clone(),
        stash.key_package.clone(),
    );
    drop(stash);
    Ok(snapshot)
}
```

Usage: `let (is_complete, nip65, inbox, kp) = self.merge_and_snapshot_stash(&pubkey, discovered)?;`

src/whitenoise/accounts.rs
```diff
@@ -1028,132 +1158,131 @@ impl Whitenoise {
...
```
jgmontoya: and used here as well

src/whitenoise/accounts.rs
jgmontoya: documented

jgmontoya: extracted (slightly differently but better)

Issue #511, opened 2026-02-21
Messages silently fail for specific npub (no send error, no receive)


vladimir-krstic:
# Summary
**npub1t3gd5yefglarhar4n6uh34uymvft4tgu8edk5465zzhtv4rrnd9sg7upxq** appears “broken” in White Noise. Any message sent to this account shows no error on the sender side, but the recipient never receives it. Same in reverse: messages sent from this account show no error, but the other party never receives them.

This pubkey historically worked on the original Svelte app (published key packages and sent messages). It has been misbehaving since the Flutter switch.

# Affected account
	•	npub: *npub1t3gd5yefglarhar4n6uh34uymvft4tgu8edk5465zzhtv4rrnd9sg7upxq*

# Steps to reproduce
	•	From any normal account A, open chat with the affected npub and send a message
	•	Observe: send reports success (no error)
	•	Check on affected npub side: message never arrives

# Also:
	•	From the affected npub, send a message to any normal account A
	•	Observe: send reports success (no error)
	•	Check on A side: message never arrives

# Expected
	•	Recipient receives messages
	•	Sender gets an error if publish or delivery fails

# Actual
	•	Silent failure: no visible error, no delivery in either direction

# Notes / observations
	•	Primal does not show a correct relay list for this account, it only shows relay.primal.net
	•	This suggests something off with the account’s relay list (kind 10002) or how clients interpret it

# Cleanup attempts already done
	•	Tried nuking events across ~500 relays
	•	Kept kinds: 0, 1, 3, 6, 7, 9734, 9735
	•	Some events could not be deleted
	•	About 30 event IDs still present

# Remaining kinds still seen:
	•	10002 (relay list)
	•	4 (DM)
	•	10115
	•	30818
	•	10050
	•	10040
	•	10000
	•	443
	•	10051
	•	30078

jgmontoya:
Did some digging on the npub and found a couple of interesting things:

| Kind      | Name              | damus.io     | nos.lol      | primal.net   |
| --------- | ----------------- | ------------ | ------------ | ------------ |
| 0         | Metadata          | found        | found        | found        |
| 3         | Contacts          | found        | found        | found        |
| **10002** | **NIP-65 Relays** | **2 relays** | **2 relays** | **2 relays** |
| **10050** | **Inbox Relays**  | **MISSING**  | **MISSING**  | **MISSING**  |
| 10051     | KP Relays         | found        | found        | found        |
[...]

Issue #447, opened 2026-01-20
Track RUSTSEC-2026-0002: lru crate unsoundness


erskingardner:
## Advisory Details

- **RUSTSEC ID**: [RUSTSEC-2026-0002](https://rustsec.org/advisories/RUSTSEC-2026-0002)
- **Component**: `lru` crate (transitive dependency via nostr-sdk)
- **Severity**: TBD
- **Issue**: Unsoundness in lru crate

## Current Status

This advisory is currently **ignored** in our cargo audit checks (added in CI workflow).

## Justification

This is a transitive dependency from nostr-sdk. Awaiting upstream fix.

## Risk Assessment

TBD - needs evaluation of the specific unsoundness issue.

## Mitigation

- Monitor nostr-sdk for updates that address this dependency
- Evaluate the specific unsoundness to determine actual risk

## Action Items

- [ ] Review the specific unsoundness details
- [ ] Document justification in `docs/SECURITY.md`
- [ ] Monitor nostr-sdk for dependency updates
- [ ] Remove from ignore list when fixed

Issue #446, opened 2026-01-20
Track RUSTSEC-2024-0384: instant crate unmaintained


erskingardner:
## Advisory Details

- **RUSTSEC ID**: [RUSTSEC-2024-0384](https://rustsec.org/advisories/RUSTSEC-2024-0384)
- **Component**: `instant` crate (transitive dependency via rust-nostr)
- **Severity**: Warning (unmaintained)
- **Issue**: The `instant` crate is no longer actively maintained

## Current Status

This advisory is currently **ignored** in our cargo audit checks.

## Justification

This is a transitive dependency from the rust-nostr ecosystem. The crate provides basic cross-platform time functionality with minimal attack surface.

## Risk Assessment

**Low** - basic functionality that is unlikely to introduce security issues.

## Mitigation

- Will be resolved when rust-nostr updates to an alternative time library
- Monitor rust-nostr releases for updates

## Action Items

- [ ] Monitor rust-nostr for migration away from `instant` crate
- [ ] Remove from ignore list when upstream updates dependency
- [ ] Update `docs/SECURITY.md` when resolved

Issue #445, opened 2026-01-20
Track RUSTSEC-2023-0071: RSA Marvin Attack in sqlx-mysql dependency


erskingardner:
## Advisory Details

- **RUSTSEC ID**: [RUSTSEC-2023-0071](https://rustsec.org/advisories/RUSTSEC-2023-0071)
- **Component**: `rsa` crate (transitive dependency via `sqlx-mysql`)
- **Severity**: Medium (5.9 CVSS)
- **Issue**: Potential key recovery through timing side-channels in RSA operations

## Current Status

This advisory is currently **ignored** in our cargo audit checks.

## Justification

This is a transitive dependency through `sqlx-mysql` which is included by sqlx's macro system at compile time. White Noise only uses SQLite for data storage, not MySQL. The vulnerable RSA code is used for MySQL authentication and is never executed in our application's code paths.

## Risk Assessment

**Low** - the vulnerable code path is never active in our application.

## Mitigation

- Monitor for updates to the `rsa` crate and sqlx dependencies
- Remove this ignore once the upstream vulnerability is patched

## Action Items

- [ ] Monitor `rsa` crate for security patch
- [ ] Monitor sqlx releases for dependency updates
- [ ] Remove from ignore list when fixed
- [ ] Update `docs/SECURITY.md` when resolved

erskingardner:
This isn't actually an issue for us, we don't use the rsa part of this crate

PR #519, opened 2026-02-27
feat(messages): replace fire-and-forget publishing with tracked delivery status (#511)

mubarakcoded:
## Summary

- Replaces fire-and-forget message publishing with optimistic UI + async delivery tracking, fixing the root cause of #511 (messages silently fail)
- Messages appear instantly with `Sending` status, then transition to `Sent(N relays)` or `Failed(reason)` after background publish with 3 automatic retries and exponential backoff (2s, 4s)
- Adds `retry_message_publish()` for manual retry of persistent failures (tap-to-retry from Flutter)
- Removes the old `background_publish_event_to()` fire-and-forget function from `publisher.rs` — fully replaced by `publish_with_retries` in `messages.rs`
- Ignores RUSTSEC-2026-0023 in CI security audit (unrelated advisory blocking CI)

## Test plan

- [x] `just precommit-quick` passes (fmt, docs, clippy, 1147 tests)
- [x] Flutter bridge codegen regenerated and compiles
- [ ] Send message with working relays → status transitions Sending → Sent
- [ ] Send message with no relays → immediate GroupMissingRelays error
- [ ] Send message offline → status transitions Sending → Failed after retries
- [ ] Retry failed message → status transitions Failed → Sending → Sent/Failed

## Summary by CodeRabbit

* **New Features**
  * Message delivery status for outgoing messages (Sending, Sent with relay count, Failed with reason) with live UI updates.
  * Automatic publish retries with exponential backoff plus a manual retry action; optimistic caching shows messages immediately and updates status in-place.

* **Chores**
  * Database migration to persist delivery status and support status updates.

### DIFF
--- a/src/whitenoise/messages.rs
+++ b/src/whitenoise/messages.rs
@@ -2331,4 +2331,122 @@ mod tests {
             "Error should mention unsupported kind, got: {err_msg}"
         );
     }
+
+    /// Test cascade_delivery_failure for kind 7 when parent message is not cached.
+    /// Should handle gracefully (warn + return) without panic.
+    #[tokio::test]
+    async fn test_cascade_reaction_failure_parent_not_found() {
+        let (whitenoise, _data_temp, _logs_temp) = create_mock_whitenoise().await;
+        let group_id = GroupId::from_slice(&[200; 32]);
+        let author = Keys::generate().public_key();
+        let stream_manager = MessageStreamManager::new();
+
+        // Point reaction at a target that doesn't exist in cache
+        let nonexistent_target = format!("{:0>64x}", 0xdeadbeefu64);
+        let tags = Tags::from_list(vec![Tag::parse(vec!["e", &nonexistent_target]).unwrap()]);
+
+        // Should return cleanly without panic — parent not found is handled
+        Whitenoise::cascade_delivery_failure(
+            7,
+            "some_reaction_id",
+            &tags,
+            &author,
+            "+",
+            &group_id,
+            &whitenoise.database,
+            &stream_manager,
+        )
+        .await;
+    }
+
+    /// Test cascade_delivery_failure for kind 7 with no e-tag at all.
+    /// Should handle the extract_reaction_target_id error gracefully.
+    #[tokio::test]
+    async fn test_cascade_reaction_failure_missing_etag() {
+        let (whitenoise, _data_temp, _logs_temp) = create_mock_whitenoise().await;
+        let group_id = GroupId::from_slice(&[201; 32]);
+        let author = Keys::generate().public_key();
+        let stream_manager = MessageStreamManager::new();
+
+        // Empty tags — no e-tag to extract
+        Whitenoise::cascade_delivery_failure(
+            7,
+            "some_reaction_id",
+            &Tags::new(),
+            &author,
+            "+",
+            &group_id,
+            &whitenoise.database,
+            &stream_manager,
+        )
+        .await;
+    }
+
+    /// Test that deleting the last message in a group exercises the
+    /// find_last_by_group_ids path in cache_and_apply_outgoing_deletion.
+    #[tokio::test]
+    async fn test_send_deletion_of_last_message_exercises_last_check() {
+        let (whitenoise, _data_temp, _logs_temp) = create_mock_whitenoise().await;
+        let creator = whitenoise.create_identity().await.unwrap();
+        let members = setup_multiple_test_accounts(&whitenoise, 1).await;
+        let member_pubkey = members[0].0.pubkey;
+
+        tokio::time::sleep(Duration::from_millis(200)).await;
+
+        let config = create_nostr_group_config_data(vec![creator.pubkey]);
+        let group = whitenoise
+            .create_group(&creator, vec![member_pubkey], config, None)
+            .await
+            .unwrap();
+
+        // Send two messages so we can delete the last one
+        whitenoise
+            .send_message_to_group(
+                &creator,
+                &group.mls_group_id,
+                "First message".to_string(),
+                9,
+                None,
+            )
+            .await
+            .unwrap();
+
+        tokio::time::sleep(Duration::from_millis(100)).await;
+
+        let second = whitenoise
+            .send_message_to_group(
+                &creator,
+                &group.mls_group_id,
+                "Second (last) message".to_string(),
+                9,
+                None,
+            )
+            .await
+            .unwrap();
+
+        tokio::time::sleep(Duration::from_millis(100)).await;
+
+        let last_id = second.message.id.to_hex();
+
+        // Delete the last message — exercises find_last_by_group_ids + LastMessageDeleted path
+        let deletion_tags = Some(vec![Tag::parse(vec!["e", &last_id]).unwrap()]);
+        let del_result = whitenoise
+            .send_message_to_group(
+                &creator,
+                &group.mls_group_id,
+                String::new(),
+                5,
+                deletion_tags,
+            )
+            .await;
+        assert!(del_result.is_ok(), "Deletion should succeed");
+
+        // The second message should be marked deleted
+        let msg =
+            AggregatedMessage::find_by_id(&last_id, &group.mls_group_id, &whitenoise.database)
+                .await
+                .unwrap()
+                .unwrap();
+        assert!(msg.is_deleted, "Last message should be marked deleted");
+    }
 }


.github/workflows/ci.yml
```diff
@@ -82,7 +82,8 @@ jobs:
           # - RUSTSEC-2023-0071: RSA Marvin Attack (transitive via sqlx-mysql, not used in our SQLite-only app)
           # - RUSTSEC-2024-0384: instant unmaintained (transitive via rust-nostr, low risk)
           # - RUSTSEC-2026-0002: lru unsound (transitive via nostr-sdk, awaiting upstream fix)
-          ignore: RUSTSEC-2023-0071,RUSTSEC-2024-0384,RUSTSEC-2026-0002
+          # - RUSTSEC-2026-0023: libcrux-ecdh (transitive via mdk-core → openmls_rust_crypto, awaiting upstream fix)
+          ignore: RUSTSEC-2023-0071,RUSTSEC-2024-0384,RUSTSEC-2026-0002,RUSTSEC-2026-0023
```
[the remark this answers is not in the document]
mubarakcoded: The repo doesn't use audit.toml; CI uses rustsec/audit-check@v2 with inline ignore:. The existing pattern is tracking issues (#445, #446, #447, now #527) with action items. Adding audit.toml governance would introduce a new pattern inconsistent with the established approach.

src/whitenoise/messages.rs
```diff
@@ -42,24 +48,342 @@ impl Whitenoise {
...
+                })?;
+
+            // Emit NewMessage so the UI shows it immediately (optimistic)
+            self.message_stream_manager.emit(
+                group_id,
+                MessageUpdate {
+                    trigger: UpdateTrigger::NewMessage,
+                    message: chat_message,
+                },
+            );
```
jgmontoya: This should be extracted to a method

src/whitenoise/messages.rs
```diff
@@ -42,24 +48,342 @@ impl Whitenoise {
...
+            }
+        }
+    }
+
+    /// Publish an event with exponential backoff retries.
+    ///
+    /// Attempts up to 3 publishes with delays of 2s and 4s between retries.
+    /// On success, updates the delivery status to `Sent(relay_count)`.
+    /// On failure (all retries exhausted), updates to `Failed(reason)`.
+    async fn publish_with_retries(
```
jgmontoya: I think this method belongs to nostr_manager

src/whitenoise/messages.rs
```diff
@@ -42,24 +48,342 @@ impl Whitenoise {
...
+                    .await
+                {
+                    tracing::warn!(
+                        target: "whitenoise::messages::delivery",
+                        "Failed to publish kind {} event: {e}",
+                        kind,
+                    );
+                }
+            }
+        });
```
jgmontoya: I think this was cleaner when just calling the method from nostr_manager

src/whitenoise/messages.rs
jgmontoya, line 1: From my agent:

> **src/whitenoise/messages.rs:815-858**
> 
> ```rust
> async fn update_and_emit_delivery_status(
>     event_id: &str,
>     group_id: &GroupId,
>     status: &DeliveryStatus,
>     database: &Database,
>     stream_manager: &MessageStreamManager,
> ) {
>     if let Err(e) =
>         AggregatedMessage::update_delivery_status(event_id, group_id, status, database).await
>     {
>         // ... error handling
>         return;
>     }
> 
>     // Re-fetch the full message to emit with updated status
>     match AggregatedMessage::find_by_id(event_id, group_id, database).await {
> ```
> 
> There's a gap between updating the delivery status (line 823) and re-fetching the message (line 834). If `retry_message_publish` is called during this window, it could:
> 1. See the old status in the database
> 2. Pass the Failed state guard check
> 3. Start a competing publish task
> 
> While both publishes would eventually succeed (they're idempotent), you'd have two parallel retry attempts racing. Consider one of:
> - Add a version/generation counter to delivery status updates to detect concurrent modifications
> - Use optimistic locking (update WHERE status = expected_old_value)
> - Document this as acceptable behavior if parallel retries are harmless

mubarakcoded: Documented as acceptable behavior. Retries are user-initiated so the overlap window is practically unreachable, and even if two publishes did race they're idempotent — the UI self-corrects on the next DeliveryStatusChanged emission. If this ever becomes a real concern, we can add optimistic locking (UPDATE WHERE delivery_status = 'Failed') to make the guard atomic with the status transition.

src/whitenoise/messages.rs
```diff
@@ -42,24 +43,230 @@ impl Whitenoise {
...
+        let mdk_message =
+            mdk.get_message(group_id, event_id)?
+                .ok_or(WhitenoiseError::MdkCoreError(
+                    mdk_core::error::Error::MessageNotFound,
+                ))?;
+
+        // Re-create the MLS message event for publishing
+        let inner_event = UnsignedEvent::new(
+            mdk_message.pubkey,
+            mdk_message.created_at,
```
erskingardner: Shouldn't the `created_at` time here be the time now since we're retrying sending and we want it to show the actual current time?

src/nostr_manager/publisher.rs
```diff
@@ -229,6 +200,142 @@ impl NostrManager {
...
+    async fn update_and_emit_delivery_status(
+        event_id: &str,
+        group_id: &GroupId,
+        status: &DeliveryStatus,
+        database: &Database,
+        stream_manager: &MessageStreamManager,
+    ) {
+        if let Err(e) =
+            AggregatedMessage::update_delivery_status(event_id, group_id, status, database).await
+        {
```
jgmontoya, line 300: Could we avoid the race condition by having update_delivery_status return the updated aggregated message?

src/whitenoise/messages.rs
jgmontoya: We should use now instead of the previous created_at

erskingardner: actually - we want this `UnsignedEvent` to be treated as a brand new message in the database. the original should remain in the MDK messages table as `Created` and we should reuse only the `tags`, `kind`, and `content` here to create a brand new message. That will create a new aggregated_messages row and new MDK message - which leaves the original messages in the UX making it clear to users what has happened. 

Does that make sense?

src/whitenoise/messages.rs
```diff
@@ -42,24 +43,230 @@ impl Whitenoise {
...
+            mdk_message.kind,
+            mdk_message.tags.clone(),
+            &mdk_message.content,
+        );
+        let message_event = mdk.create_message(group_id, inner_event)?;
+
+        // Update status to Sending and emit MessageRetried (includes reposition)
+        let sending_status = DeliveryStatus::Sending;
+        AggregatedMessage::update_delivery_status(
```
erskingardner: We want to leave the original AggregatedMessage in place and just simply recreate a new one.

src/whitenoise/messages.rs
```diff
@@ -42,24 +43,230 @@ impl Whitenoise {
...
+        let group_id_clone = group_id.clone();
+        let account_pubkey = account.pubkey;
+
+        tokio::spawn(async move {
+            nostr
+                .publish_with_retries(
+                    message_event,
+                    &account_pubkey,
+                    &group_relays,
+                    &event_id_str,
```
erskingardner: this should be the new `message_event.id` now since we'll keep both the messages.

src/whitenoise/messages.rs
mubarakcoded: @erskingardner Just to clarify the intended UX — with this approach, the user would see two copies of the same message on screen: the original in its Failed state (staying in place) and the new retry as a fresh Sending message at the bottom. Is that the intended behavior, or should the original failed message be hidden/deleted when the retry is created?

src/whitenoise/messages.rs
```diff
@@ -42,24 +43,200 @@ impl Whitenoise {
...
+            };
+
+        // Build tags from original, adding a nonce to guarantee a unique event ID
+        // (without a nonce, retries within the same second produce identical IDs
+        // because Nostr event IDs are deterministic over pubkey+timestamp+kind+tags+content)
+        let mut tags = original.tags.to_vec();
+        tags.push(Tag::custom(
+            TagKind::custom("nonce"),
+            [event_id_str.clone()],
+        ));
```
jgmontoya: I don't think this is needed, the message should be using a new set of keys, no?

mubarakcoded: The outer MLS event uses different keys, but we track messages by the inner Nostr event ID, which is deterministic over pubkey+timestamp+kind+tags+content. A same-second retry would collide without the nonce.

erskingardner: but the exponential back off on the original event means that this isn't a possible scenario. 

This is the flow as I see it: 

1. You send a message. it isn't successful, and retries in the publish with retries loop which takes several seconds as it retries a few times.
2. We display that the message is failed in the UI and then the user retries
3. The user then retries and we create a new messages.

src/whitenoise/messages.rs
```diff
@@ -42,24 +43,200 @@ impl Whitenoise {
...
+            self.process_and_emit_outgoing_message(&mdk_message, group_id)
+                .await?;
+        }
+
+        // Spawn background publish task
+        let nostr = self.nostr.clone();
+        let event_id_str = event_id.to_string();
+        let account_pubkey = account.pubkey;
+
+        if kind == 9 {
```
erskingardner: I don't like that we're only doing this for kind 9s. The publish retry should apply to all the event kinds and whether we show the ability to retry is a UI question.

src/whitenoise/messages.rs
mubarakcoded: You're all right. I have removed the nonce.

db_migrations/0032_message_delivery_status.sql
jgmontoya, line 1: We'll probably need an index on status or `(mls_group_id, status)` as we'll be filtering by status frequently

src/whitenoise/database/aggregated_messages.rs
```diff
@@ -452,6 +491,60 @@ impl AggregatedMessage {
...
+        )
+        .bind(message_id)
+        .bind(group_id.as_slice())
+        .bind(serde_json::to_string(status)?)
+        .execute(&mut *tx)
+        .await
+        .map_err(|e| match e {
+            // FK violation means the message doesn't exist
+            sqlx::Error::Database(ref db_err) if db_err.code().as_deref() == Some("787") => {
+                DatabaseError::Sqlx(sqlx::Error::RowNotFound)
```
jgmontoya: Is this the best way to check for this? I'm not sure but the `== Some("787")` check looks weird/fragile

src/whitenoise/messages.rs
mubarakcoded: > I don't like that we're only doing this for kind 9s. The publish retry should apply to all the event kinds and whether we show the ability to retry is a UI question.

Yes, this makes sense, delivery tracking shouldn't be limited to kind 9. For this PR, though, reactions and deletions are aggregated onto parent messages rather than stored as standalone rows, so the delivery status table doesn't have a natural FK target for them. Happy to tackle this as a follow-up — would probably need to rethink how we track non-message events before we can attach delivery status to them. Does that work, or do you want it in this PR?

src/whitenoise/messages.rs
```diff
@@ -42,24 +45,519 @@ impl Whitenoise {
...
+
+        Ok(())
+    }
+
+    /// Reverse optimistic aggregated effects when a reaction/deletion fails to publish.
+    ///
+    /// - **Kind 7 (reaction)**: Remove the reaction from the parent message's summary.
+    /// - **Kind 5 (deletion)**: Clear `deletion_event_id` on targets so they reappear.
+    /// - **Kind 9 / other**: No cascade needed — the message already shows Failed status.
+    async fn cascade_delivery_failure(
```
jgmontoya, line 378: When cascade operations fail (e.g., database update errors), they log warnings but don't propagate errors or retry. This means a failed reaction could permanently persist in the parent's reaction summary even after delivery failure.

I'm not sure exactly  what would be the best way to handle this though, so we might push it to the future (we should create an issue so we don't forget)

jgmontoya (approved):

PR #524, opened 2026-02-27
fix(subscriptions): remove unsubscribe-first blind window and anchor replay to account sync state

erskingardner:
## Summary

Replace the hardcoded `Timestamp::now() - 10s` replay anchor in the account subscription refresh path with `account.since_timestamp(10)`, which anchors replay to the account's actual last-synced state.

On master, every refresh replays from 10 seconds ago regardless of when the account last synced. This means if the account was last synced an hour ago, only the last 10 seconds are replayed on refresh — events in between could be missed. With this change, the replay window correctly covers the full gap since the last successful sync.

Gift-wrap backdating (nostr-sdk timestamps gift-wraps up to 48 hours in the past) is already handled inside `setup_giftwrap_subscription` via `adjust_since_for_giftwrap`, which subtracts an additional 7-day `GIFTWRAP_LOOKBACK_BUFFER` from the `since` value. No extra buffer is needed at this layer.

Also ignores `RUSTSEC-2026-0023` in the security audit (libcrux-ecdh X25519 validation, transitive via openmls_rust_crypto → hpke-rs → hpke-rs-libcrux; pre-existing on master, fix tracked separately).

### `src/nostr_manager/mod.rs`
- `update_account_subscriptions_with_signer`: accepts `since: Option<Timestamp>` from caller instead of computing `Timestamp::now() - 10s` internally
- `update_account_subscriptions` (test-only): same signature change
- Updated `test_update_account_subscriptions_empty_relays` for new signature
- Added `test_update_account_subscriptions_does_not_unsubscribe_first`
- Added `test_update_account_subscriptions_replay_anchor_forwarded`

### `src/whitenoise/accounts/setup.rs`
- `refresh_account_subscriptions`: passes `account.since_timestamp(10)` as `since` to `update_account_subscriptions_with_signer`; doc comment explains the giftwrap lookback is handled at a lower layer

### `.github/workflows/ci.yml`
- Add `RUSTSEC-2026-0023` to the security audit ignore list

## Out of scope

- Global batched subscription redesign
- Welcome/self-update subscription timing
- Message delivery status features

## Acceptance

- Replay anchor uses `account.since_timestamp` path rather than a hardcoded wall-clock offset ✓
- Security audit passes ✓
- `just precommit` passes (fmt ✓, docs ✓, clippy ✓, tests ✓, integration tests ✓)

## Summary by CodeRabbit

* **Improvements**
  * More reliable subscription updates using replacement semantics to avoid unnecessary disconnects.
  * Optional "since" anchoring forwarded to subscription setup for tighter replay window control.
  * Cleaner resubscription flow that avoids forced unsubscribe delays.

* **Tests**
  * New tests covering zero-subscription cases, replay-anchor forwarding, and behavior with no relays.

### DIFF
--- a/src/nostr_manager/mod.rs
+++ b/src/nostr_manager/mod.rs
@@ -269,11 +269,14 @@ impl NostrManager {
     /// Updates account subscriptions by re-issuing them with the same stable subscription IDs.
     ///
     /// NIP-01 guarantees that a `REQ` with an existing subscription ID replaces the old subscription
-    /// atomically on each relay. Callers that want to clean up stale relay state first (e.g.
-    /// `refresh_account_subscriptions`) should call `unsubscribe_account_subscriptions` before
-    /// invoking this method.
+    /// atomically on each relay. This method does **not** unsubscribe first; it relies on the
+    /// NIP-01 replacement semantics to overwrite existing subscriptions on each relay.
     ///
-    /// The `since` parameter anchors the replay window.  Pass `account.since_timestamp(buffer)` from
+    /// Callers that need a clean relay state (e.g. `refresh_account_subscriptions`) must call
+    /// `unsubscribe_account_subscriptions` before invoking this method. Skipping that step leaves
+    /// stale subscriptions open on relays that are no longer in the updated relay set.
+    ///
+    /// The `since` parameter anchors the replay window. Pass `account.since_timestamp(buffer)` from
     /// the caller so the anchor reflects the account's actual sync state rather than a hardcoded
     /// wall-clock offset.
     pub(crate) async fn update_account_subscriptions_with_signer(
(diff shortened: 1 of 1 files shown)


jgmontoya (dismissed):

dannym-arx (approved):
lgtm

src/whitenoise/accounts/setup.rs
```diff
@@ -714,13 +722,22 @@ impl Whitenoise {
             account.pubkey
         );
 
+        // Clean up existing subscriptions before resubscribing.
+        self.nostr
+            .unsubscribe_account_subscriptions(&account.pubkey)
+            .await?;
+
         let user_relays: Vec<RelayUrl> = Relay::urls(&account.nip65_relays(self).await?);
```
jgmontoya, line 731: From my agent:

> The `refresh_account_subscriptions` function calls `unsubscribe_account_subscriptions` before calling `update_account_subscriptions_with_signer`. However, the documentation for `update_account_subscriptions_with_signer` (line 272-274 in nostr_manager/mod.rs) states:
> 
> > "Callers that want to clean up stale relay state first (e.g. `refresh_account_subscriptions`) should call `unsubscribe_account_subscriptions` before invoking this method."
> 
> This is correct as written, but creates a subtle coupling between these functions. Consider:
> 
> 1. The function name `update_account_subscriptions_with_signer` suggests it will handle the update completely, but it relies on the caller to handle cleanup.
> 2. If another caller uses `update_account_subscriptions_with_signer` without reading the docs carefully, they might not realize they need to unsubscribe first for relay changes.
> 3. The comment on line 725 says "Clean up existing subscriptions before resubscribing" but the actual resubscribe happens through `update_account_subscriptions_with_signer`, which is non-obvious.
> 
> **Recommendation**: Consider one of these approaches:
> - Add a boolean parameter `unsubscribe_first: bool` to `update_account_subscriptions_with_signer` to make the behavior explicit and remove the coupling
> - Rename the function to something like `resubscribe_account_subscriptions_with_signer` to make it clearer that it doesn't handle cleanup
> - Add a debug assertion or warning log if subscriptions exist when `update_account_subscriptions_with_signer` is called, to catch misuse

jgmontoya (approved):

PR #526, opened 2026-02-27
fix(welcome): replace fixed sleep with subscription-gated catch-up before self-update

erskingardner:
## Summary

- Removes the unconditional 5-second `tokio::time::sleep` that previously served as the only sequencing mechanism between subscription setup and the post-welcome self-update
- Replaces it with an explicit 4-step flow: (1) subscription setup gating steps 2 and 4, (2) bounded catch-up fetch queuing recent group messages before epoch advance, (3) independent ops concurrently, (4) self-update only when subscriptions are live
- Adds `join-time-race` integration scenario verifying that a message sent immediately after the invite (before the member processes the Welcome) is visible to the member after join completes

### `handle_giftwrap.rs` — core sequencing fix

`finalize_welcome_with_instance` now follows an explicit sequence:

1. **Subscription setup** — awaited; its result (`subscription_ok`) gates steps 2 and 4
2. **Catch-up fetch** — only when subscriptions are live: pulls up to 200 `MlsGroupMessage` events from the last 120 s, queues them through `event_sender`, waits 500 ms for the queue to drain (replacing the 5 s sleep with a purposeful, shorter wait)
3. **Independent ops** — group info, key rotation, image sync, welcomer lookup run concurrently regardless of subscription status
4. **Self-update** — only when `subscription_ok`; if subscription setup failed, a structured log entry with `reason = "subscription_setup_failed"` is emitted and self-update is skipped

### `query.rs` — bounded catch-up fetch

Adds `fetch_group_messages_catchup` with:
- `CATCHUP_EVENT_LIMIT = 200` — prevents unbounded relay backlog replay
- `CATCHUP_LOOKBACK = 120 s` — covers the Welcome dispatch → subscription live window
- `CATCHUP_TIMEOUT = 2 s` — dedicated shorter timeout so the catch-up cannot stall the join flow the way the 5 s default client timeout could

### `join_time_race.rs` — new integration scenario

Exercises the exact race: creator sends a message immediately after group creation (before invitee processes Welcome), then verifies the member sees both the pre-join and post-join messages after the full join-time flow (catch-up + self-update) completes.

### Integration scenario fixes

Three existing scenarios were relying on the implicit 5 s sleep as an undocumented synchronization barrier. Now that the sleep is gone, those scenarios needed explicit `WaitForWelcome` + `VerifySelfUpdate` guards before the member sends messages:

| Scenario | Fix |
|----------|-----|
| `advanced_messaging` | Add `WaitForWelcome` + `VerifySelfUpdate` for reactor before any messages are sent |
| `message_streaming` | Add `WaitForWelcome` + `VerifySelfUpdate` in phase1 before `stream_member` sends in phase2 |
| `chat_list_streaming` | Add `VerifySelfUpdate` after `WaitForWelcome` in phase3 so epoch advance completes before phases 4–5 |

## Acceptance criteria

- [x] Self-update is no longer unconditional after a fixed delay
- [x] Self-update is gated on successful subscription setup; failure is logged with structured reason
- [x] Bounded catch-up fetch queues in-flight epoch-N messages before epoch advance
- [x] Join-time race scenario passes (`join-time-race` 7/7)
- [x] `just precommit` passes (fmt, docs, clippy, unit tests, all 19 integration scenarios)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved join synchronization so members fully finalize their join before sending or receiving messages, preventing decryption/epoch mismatch errors.

* **New Features**
  * Automatic catch-up of recent group messages on join so new members see recent history (short lookback window) immediately.

* **Tests**
  * Added end-to-end and unit tests covering join sequencing, catch-up behavior, and edge cases.

### DIFF
--- a/src/integration_tests/scenarios/advanced_messaging.rs
+++ b/src/integration_tests/scenarios/advanced_messaging.rs
@@ -36,18 +36,6 @@ impl Scenario for AdvancedMessagingScenario {
             .execute(&mut self.context)
             .await?;
 
-        // Wait for the reactor to receive and process the welcome before anyone
-        // sends messages. The reactor's join-time self-update advances the group
-        // epoch; messages sent before that completes would be encrypted at epoch
-        // N-1 and could not be decrypted by other members at epoch N.
-        WaitForWelcomeTestCase::for_account("adv_msg_reactor", "advanced_messaging_group")
-            .execute(&mut self.context)
-            .await?;
-
-        VerifySelfUpdateTestCase::for_account("adv_msg_reactor", "advanced_messaging_group")
-            .execute(&mut self.context)
-            .await?;
-
         // Send initial message that will receive reactions
         SendMessageTestCase::basic()
             .with_sender("adv_msg_sender")
--- a/src/integration_tests/scenarios/chat_list.rs
+++ b/src/integration_tests/scenarios/chat_list.rs
@@ -104,19 +104,6 @@ impl Scenario for ChatListScenario {
         // ============================================================
         tracing::info!("Test 4: Verifying last message in chat list...");
 
-        // Wait for bob to complete the post-welcome self-update before alice
-        // sends.  Bob's self-update commits advance the group epoch; if alice
-        // sends at epoch N while bob's self-update is in flight, alice's
-        // message comes back from the relay at what is now epoch N+1 and is
-        // rejected as Wrong Epoch, so alice never sees her own last message.
-        WaitForWelcomeTestCase::for_account("chat_list_bob", "chat_list_group")
-            .execute(&mut self.context)
-            .await?;
-
-        VerifySelfUpdateTestCase::for_account("chat_list_bob", "chat_list_group")
-            .execute(&mut self.context)
-            .await?;
-
         SendMessageTestCase::basic()
             .with_sender("chat_list_alice")
             .with_group("chat_list_group")
--- a/src/integration_tests/scenarios/chat_list_streaming.rs
+++ b/src/integration_tests/scenarios/chat_list_streaming.rs
@@ -63,15 +63,6 @@ impl ChatListStreamingScenario {
             .execute(&mut self.context)
             .await?;
 
-        // Wait for the member's post-welcome self-update to complete before
-        // proceeding. The self-update runs as a background task that advances
-        // the group epoch; any messages sent while it is still running will be
-        // at a different epoch, which can interfere with later phases of this
-        // scenario that check the "last message" in the chat list.
-        VerifySelfUpdateTestCase::for_account("stream_member", Self::GROUP_NAME)
-            .execute(&mut self.context)
-            .await?;
-
         // Verify the NewGroup update was received
         new_group_verifier.execute(&mut self.context).await?;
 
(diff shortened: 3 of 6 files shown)


src/integration_tests/scenarios/advanced_messaging.rs
jgmontoya, line 1: nice

src/whitenoise/event_processor/event_handlers/handle_giftwrap.rs
```diff
@@ -268,72 +305,173 @@ impl Whitenoise {
...
+        // retry logic, and MLS state management apply correctly.
+        let pubkey_hash = whitenoise.nostr.create_pubkey_hash(&account.pubkey);
+        let subscription_id = format!("{}_mls_messages", pubkey_hash);
+
+        for event in events {
+            use crate::types::ProcessableEvent;
+            let processable =
+                ProcessableEvent::new_nostr_event(event, Some(subscription_id.clone()));
+            match tokio::time::timeout(
+                std::time::Duration::from_millis(100),
```
jgmontoya: This seems a bit too aggresive maybe? What if the event_processor is busy due to other groups having a lot of activity?

src/whitenoise/event_processor/event_handlers/handle_giftwrap.rs
```diff
@@ -223,36 +238,58 @@ impl Whitenoise {
...
+                    account = %account.pubkey.to_hex(),
+                    group = %hex::encode(group_id.as_slice()),
+                    error = %e,
+                    reason = "subscription_setup_failed",
+                    "Subscription setup failed; skipping catch-up and self-update to avoid epoch mismatch"
+                );
+                false
+            }
+        };
+
```
jgmontoya, line 270: Skipping the self-update here seems fine, but I think we need a background task to handle self-updates automatically or something?

Can be deferred, just pointing it out so we don't forget

src/nostr_manager/query.rs
```diff
@@ -1,12 +1,35 @@
...
+/// Maximum time to wait for the relay to respond during catch-up.
+///
+/// Keep this well below the default client timeout (5 s) so the catch-up
+/// step does not block the join-time flow for too long when the relay is
+/// slow or returns no results.  2 s is plenty for both local (sub-50 ms)
+/// and remote (sub-500 ms) relays while still leaving headroom before the
+/// WaitForWelcome test helper times out.
+const CATCHUP_TIMEOUT: Duration = Duration::from_secs(2);
+
 impl NostrManager {
```
jgmontoya, line 10: These seem fine to me, although arbitrary I think they will do just fine.

The only thing that worries me a bit is that if one of these limits is reached we might face silent failures:

- if there are more than CATCHUP_EVENT_LIMIT events, we might miss commits and be left on a broken/forked state
- if the welcome was sent more than CATCHUP_LOOKBACK seconds ago I think we could also be missing commits

src/whitenoise/event_processor/event_handlers/handle_giftwrap.rs
erskingardner: yes, we do need one - we have the data being tracked now from MDK but we don't yet have the maintenance/background task

PR #537, opened 2026-03-02
Feat/cli

jgmontoya:
## Summary by CodeRabbit

* **New Features**
  * New CLI client and daemon with rich commands for identity, accounts, groups, chats, messages, follows, users, profile, relays, settings, notifications, streaming IPC, and daemon lifecycle controls; platform-aware config and optional CLI feature set.
* **Tests**
  * Comprehensive end-to-end CLI test suite, new end-to-end test recipe, and a benchmark test binary.
* **Documentation**
  * Detailed CLI README with architecture, commands, and extension guidance.

### DIFF
--- a/src/cli/dispatch.rs
+++ b/src/cli/dispatch.rs
@@ -435,6 +435,11 @@ async fn messages_subscribe<W>(
             Err(tokio::sync::broadcast::error::RecvError::Closed) => break,
             Err(tokio::sync::broadcast::error::RecvError::Lagged(n)) => {
                 tracing::warn!("message stream lagged by {n} messages");
+                let _ = write_response(
+                    writer,
+                    &Response::err(format!("stream lagged: {n} messages dropped")),
+                )
+                .await;
             }
         }
     }
@@ -493,6 +498,11 @@ where
             Err(tokio::sync::broadcast::error::RecvError::Closed) => break,
             Err(tokio::sync::broadcast::error::RecvError::Lagged(n)) => {
                 tracing::warn!("chat list stream lagged by {n} updates");
+                let _ = write_response(
+                    writer,
+                    &Response::err(format!("stream lagged: {n} updates dropped")),
+                )
+                .await;
             }
         }
     }
@@ -527,6 +537,11 @@ where
             Err(tokio::sync::broadcast::error::RecvError::Closed) => break,
             Err(tokio::sync::broadcast::error::RecvError::Lagged(n)) => {
                 tracing::warn!("notification stream lagged by {n} updates");
+                let _ = write_response(
+                    writer,
+                    &Response::err(format!("stream lagged: {n} notifications dropped")),
+                )
+                .await;
             }
         }
     }
--- a/src/cli/protocol.rs
+++ b/src/cli/protocol.rs
@@ -1,10 +1,12 @@
+use std::fmt;
+
 use serde::{Deserialize, Serialize};
 
 /// A request from the CLI client to the daemon.
 ///
 /// Each variant maps to one daemon method. The JSON wire format uses
 /// `{"method": "variant_name", "params": {...}}` via serde's tagged enum.
-#[derive(Debug, Deserialize, Serialize)]
+#[derive(Deserialize, Serialize)]
 #[serde(tag = "method", content = "params")]
 pub enum Request {
     // Daemon management
@@ -161,6 +163,19 @@ fn default_radius_start() -> u8 {
 
 fn default_radius_end() -> u8 {
     2
+}
+
+/// Manual `Debug` impl to prevent accidental nsec exposure in logs.
+impl fmt::Debug for Request {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::LoginStart { .. } => f
+                .debug_struct("LoginStart")
+                .field("nsec", &"[REDACTED]")
+                .finish(),
+            other => write!(f, "{}", serde_json::to_string(other).unwrap_or_default()),
+        }
+    }
 }
 
 impl Request {


src/bin/wn.rs
```diff
@@ -0,0 +1,127 @@
...
+
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+    let args = Args::parse();
+    let config = Config::resolve(None, None);
+    let socket = args.socket.unwrap_or_else(|| config.socket_path());
+
+    match args.command {
+        Cmd::Daemon(cmd) => cmd.run(&config).await,
+        Cmd::CreateIdentity => identity::create_identity(&socket, args.json).await,
```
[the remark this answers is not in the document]
jgmontoya: Good observation that --socket isn't used in the daemon branch, but this is by design rather than a bug.

  The --socket flag is a client connection parameter — it tells the CLI where to connect when sending requests to a running daemon (via client::send() / client::stream()). Every non-daemon command
  uses it for this purpose.

  Daemon subcommands don't connect to the daemon as a client — they manage the daemon process directly:

  - start — spawns wnd as a subprocess with --data-dir / --logs-dir. The daemon determines its own socket path from its data directory.
  - stop — reads the PID file and sends SIGTERM. No socket involved.
  - status — checks if the PID is alive and prints the socket path from config. No socket connection.

  Since none of these operations involve a socket connection, there's no misrouting risk. If someone wanted to target a specific daemon instance, the relevant override would be --data-dir (which
  determines both the PID file and socket path), not --socket.

src/cli/config.rs
```diff
@@ -0,0 +1,111 @@
...
+
+impl Config {
+    /// Build config by merging platform defaults with CLI overrides.
+    pub fn resolve(data_dir: Option<&PathBuf>, logs_dir: Option<&PathBuf>) -> Self {
+        let defaults = Self::platform_defaults();
+        Self {
+            data_dir: data_dir.cloned().unwrap_or(defaults.data_dir),
+            logs_dir: logs_dir.cloned().unwrap_or(defaults.logs_dir),
+        }
+    }
```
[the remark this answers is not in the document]
jgmontoya: The .expect() calls here are intentional and appropriate. They fire only when dirs::home_dir() / dirs::data_dir() return None, which happens exclusively when $HOME is unset — a fundamentally broken
  environment where the CLI cannot function at all (no place to store data, create sockets, write PID files, or access the keyring).

  Converting to Result would change the exit path from panic-with-message to error-with-message — same user experience, more code. The Result approach implies the caller could meaningfully recover,
  but there's no recovery from "this OS has no home directory."

src/cli/dispatch.rs
```diff
@@ -0,0 +1,1213 @@
...
+    }
+}
+
+/// Write a single response line to the writer. Returns false if the client disconnected.
+async fn write_response(writer: &mut (impl AsyncWriteExt + Unpin), response: &Response) -> bool {
+    let mut buf = match serde_json::to_vec(response) {
+        Ok(buf) => buf,
+        Err(e) => {
+            tracing::error!("failed to serialize response: {e}");
+            return false;
```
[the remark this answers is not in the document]
jgmontoya: The tracing crate automatically sets the target field to the current module path. Since this file lives at src/cli/dispatch.rs in the whitenoise crate, every tracing::error!(...) and
  tracing::warn!(...) already has target whitenoise::cli::dispatch by default — no explicit annotation needed.

  Adding target: "whitenoise::cli::dispatch" to each call would be redundant boilerplate that restates what the framework already provides.

src/cli/server.rs
```diff
@@ -0,0 +1,285 @@
...
+    }
+
+    clean_stale_socket(&socket_path, &pid_path)?;
+
+    let listener = UnixListener::bind(&socket_path)?;
+    set_socket_permissions(&socket_path)?;
+    write_pid_file(&pid_path)?;
+
+    tracing::info!("daemon listening on {}", socket_path.display());
+
```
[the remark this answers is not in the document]
jgmontoya: Explicit targets are used selectively when they differ from the module path (e.g., target: "whitenoise::cache" in messages.rs, which routes cache-related logs separately from whitenoise::messages)

tests/cli_e2e.rs
```diff
@@ -0,0 +1,475 @@
...
+
+    if let Some(err) = parsed.get("error") {
+        panic!("wn {args:?} returned error: {err}");
+    }
+
+    parsed
+        .get("result")
+        .cloned()
+        .unwrap_or(serde_json::Value::Null)
+}
```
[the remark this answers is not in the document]
jgmontoya: The wn() helper panics by design — it's a test helper, and errors from these commands indicate real failures
  (bad arguments, daemon crash), not transient timing issues. The polling predicates wait for data to arrive
  (group sync, message delivery), not for the daemon to become available.

  If wn() returned Result and poll_until swallowed errors, genuine bugs would be masked as timeouts — making
  failures harder to diagnose, not easier. A panic inside a poll gives you the exact failing command and error
  message immediately.

  The follows_lifecycle flakiness we've observed in this branch is a relay timing issue (the relay hasn't
  propagated the contact list), not a CLI error response. That test doesn't use poll_until at all.

src/cli/commands/identity.rs
```diff
@@ -0,0 +1,181 @@
...
+                if let Some(pubkey) = account.get("pubkey").and_then(|v| v.as_str()) {
+                    println!("{pubkey}");
+                }
+            }
+        }
+    } else {
+        output::print_response(&resp, false);
+    }
+    if resp.error.is_some() {
+        std::process::exit(1);
```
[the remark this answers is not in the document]
jgmontoya: Error responses have result: None (see Response::err() in protocol.rs), so the else if on line 127 evaluates to
   None and falls through to the else branch on line 137, which correctly prints the error to stderr. The
  pubkey-printing branch is structurally unreachable when resp.error is set.

  The current flow already handles errors correctly: error → falls to else → prints error → exits 1.

src/cli/server.rs
```diff
@@ -0,0 +1,293 @@
...
+}
+
+fn write_pid_file(path: &Path) -> anyhow::Result<()> {
+    fs::write(path, std::process::id().to_string())?;
+    Ok(())
+}
+
+#[cfg(unix)]
+fn set_socket_permissions(path: &Path) -> anyhow::Result<()> {
+    use std::os::unix::fs::PermissionsExt;
```
erskingardner: move this out to the top level

src/cli/commands/daemon.rs
```diff
@@ -0,0 +1,74 @@
...
+    // When invoked via `wn daemon start`, spawn `wnd` as a child process.
+    // When invoked directly as `wnd`, this path isn't used — wnd.rs calls
+    // server::run() directly.
+    let wnd = which_wnd()?;
+    let mut cmd = Command::new(wnd);
+    cmd.arg("--data-dir").arg(&config.data_dir);
+    cmd.arg("--logs-dir").arg(&config.logs_dir);
+
+    let status = cmd.status()?;
+    if !status.success() {
```
erskingardner: From my agent: 

The command uses cmd.status() to start wnd, which waits for child exit. Since daemon is long-running, wn daemon start hangs the caller.

```rust
let status = cmd.status().context("failed to launch daemon process")?;
if !status.success() {
    anyhow::bail!("daemon exited with status: {status}");
}
```

Suggestion: Spawn detached/background process (or explicit foreground mode + clear UX), and return immediately on successful launch.

src/cli/output.rs
```diff
@@ -0,0 +1,409 @@
...
+    }
+}
+
+/// Try to extract a byte array from a JSON value and return it as a hex string.
+///
+/// Handles two patterns:
+/// - Plain array: `[100, 88, ...]`
+/// - MLS wrapper: `{"value": {"vec": [92, 112, ...]}}`
+fn try_as_hex(value: &serde_json::Value) -> Option<String> {
+    let arr = if let Some(arr) = value.as_array() {
```
erskingardner: from my agent: 

Both arms have meaningful logic, which STYLE.md says should use match.

```rs
let arr = if let Some(arr) = value.as_array() {
    arr
} else {
    value.get("value")?.get("vec")?.as_array()?
};
```

Suggestion: Rewrite as match value.as_array() { Some(arr) => arr, None => ... }.

src/cli/commands/identity.rs
```diff
@@ -0,0 +1,188 @@
...
+    let pubkey = resp
+        .result
+        .as_ref()
+        .and_then(|v| v.get("account"))
+        .and_then(|v| v.get("pubkey"))
+        .and_then(|v| v.as_str())
+        .ok_or_else(|| anyhow::anyhow!("unexpected login response: missing pubkey"))?
+        .to_string();
+
+    let relay_resp = if let Some(url) = relay {
```
erskingardner: Agent: 

Large Some and None branches are implemented as if let ... else, conflicting with style preference for match.

```rs
let relay_resp = if let Some(url) = relay {
    client::send(...).await?
} else {
    eprintln!("Relay lists not found on the network.");
    // substantial fallback logic...
};
```

Suggestion: Replace with match relay { Some(url) => ..., None => ... }.

src/cli/dispatch.rs
```diff
@@ -0,0 +1,1295 @@
...
+    group_id_hex: &str,
+    message: String,
+) -> Result<Response, Response> {
+    let account = find_account(wn, account_str).await?;
+    let group_id = parse_group_id(group_id_hex)?;
+
+    // Replicate send_message_to_group steps inline so we can publish
+    // synchronously and report relay results back to the CLI user.
+    let (inner_event, event_id) = wn
+        .create_unsigned_nostr_event(&account.pubkey, &message, 9, None)
```
erskingardner: feels like this might drift if we're recreating functionality.

src/cli/protocol.rs
```diff
@@ -0,0 +1,801 @@
...
+pub enum Request {
+    // Daemon management
+    #[serde(rename = "ping")]
+    Ping,
+
+    // Identity & auth
+    #[serde(rename = "create_identity")]
+    CreateIdentity,
+    #[serde(rename = "login_start")]
+    LoginStart { nsec: String },
```
dannym-arx, line 20: `Request` derives `Debug` (line 7), so `format!("{req:?}")` dumps the nsec in plaintext. The moment anyone adds `tracing::debug!("received: {req:?}")` to the daemon secret keys appear in logs.
Additionally, the nsec lives in un-zeroized `String` allocations throughout the serialization chain (client `serde_json::to_vec`, socket buffer, server `serde_json::from_str`, the `Request` enum itself). Multiple copies exist on the heap with no cleanup.
Use zeroize::Zeroizing<String> for the nsec field.

src/cli/server.rs
```diff
@@ -0,0 +1,294 @@
...
+
+    let shutdown = tokio::signal::ctrl_c();
+    tokio::pin!(shutdown);
+
+    loop {
+        tokio::select! {
+            accept = listener.accept() => {
+                match accept {
+                    Ok((stream, _addr)) => {
+                        tokio::spawn(handle_connection(stream));
```
dannym-arx, line 44: When CTRL+C fires, the loop breaks immediately. Any handle_connection tasks spawned here via tokio::spawn are abandoned and get killed when the runtime shuts down. If a request is mid-dispatch (e.g., send_message that committed MLS state but hasn't published to relays), it gets terminated with potentially half-committed state.

src/cli/dispatch.rs
```diff
@@ -0,0 +1,1231 @@
...
+                        "message": formatted,
+                    }));
+                    if !write_response(writer, &resp).await {
+                        return; // Client disconnected
+                    }
+                }
+            }
+            Err(tokio::sync::broadcast::error::RecvError::Closed) => break,
+            Err(tokio::sync::broadcast::error::RecvError::Lagged(n)) => {
+                tracing::warn!("message stream lagged by {n} messages");
```
dannym-arx, line 437: When the broadcast receiver lags, `n` messages are permanently lost and the will client never know

src/cli/server.rs
jgmontoya: I think this is ok and equivalent to how it works with the flutter app

src/cli/protocol.rs
jgmontoya: Made it so it doesn't show on the debug

PR #538, opened 2026-03-03
fix: make NostrManager::with_signer cancellation-safe

### LINKED ISSUE #535: Fix cancellation safety in NostrManager::with_signer
`NostrManager::with_signer` is cancellation-unsafe:

- It sets a temporary signer, awaits `f().await`, then unsets the signer.
- If the task/future is canceled while waiting on `f().await`, execution does not continue to `unset_signer()`.
- That can leave the signer set on the shared client, so subsequent operations may run with the wrong signer.

Code reference:
- https://github.com/marmot-protocol/whitenoise-rs/blob/bd74532567cba53afe104cf19967350572ef65b7/src/nostr_manager/mod.rs#L206-L215

Why this matters:
- `with_signer` is intended to provide scoped signer usage, but cancellation breaks that guarantee.
- `signer_lock` is still dropped on cancellation, so other operations can proceed while the stale signer remains set.

Suggested fix direction:
- Use an RAII-style guard that unsets signer in `Drop` (or another cancellation-safe cleanup strategy) so cleanup runs when the future is dropped.
- Keep signer setup/unset serialized under `signer_lock`.

Potential verification:
- Add a test that starts `with_signer`, cancels during `f().await`, and verifies the signer is not left set.

erskingardner:
## Summary
- fix `NostrManager::with_signer` cancellation safety by introducing an RAII signer scope guard that unsets the signer in `Drop`
- keep signer setup/cleanup serialized with `signer_lock` so signer state cannot leak between operations
- add a cancellation regression test that aborts an in-flight `with_signer` call and verifies the signer is eventually unset

## Testing
- just test
- just precommit

Closes #535

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved signer lifecycle handling to ensure the signer is reliably unset and resources released even if an operation is cancelled or aborted.

* **Documentation**
  * Clarified that signer cleanup is guaranteed in cancellation scenarios.

* **Tests**
  * Added tests verifying signer is properly unset when operations are cancelled.

### DIFF
--- a/src/nostr_manager/mod.rs
+++ b/src/nostr_manager/mod.rs
@@ -80,11 +80,14 @@ struct SignerScopeGuard {
 }
 
 impl SignerScopeGuard {
-    async fn new(
+    async fn new<S>(
         client: Client,
         signer_lock: std::sync::Arc<tokio::sync::Mutex<()>>,
-        signer: impl NostrSigner + 'static,
-    ) -> Self {
+        signer: S,
+    ) -> Self
+    where
+        S: NostrSigner + 'static,
+    {
         let lock_guard = signer_lock.lock_owned().await;
         client.set_signer(signer).await;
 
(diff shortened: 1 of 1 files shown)


src/nostr_manager/mod.rs
```diff
@@ -74,6 +74,54 @@ pub struct NostrManager {
...
+
+impl Drop for SignerScopeGuard {
+    fn drop(&mut self) {
+        let Some(lock_guard) = self.lock_guard.take() else {
+            return;
+        };
+
+        let client = self.client.clone();
+        let Ok(runtime_handle) = tokio::runtime::Handle::try_current() else {
+            tracing::warn!(
```
dannym-arx, line 115: if tokio::runtime::Handle::try_current() fails, cleanup is skipped, which could leave the signer set in an off-runtime drop edge case.

I think this is fine, but just fyi

dannym-arx (approved):

PR #548, opened 2026-03-03
fix(groups): pre-validate key package compatibility before member adds

erskingardner:
## Summary
- add fail-fast validation for fetched key package events before creating a group or adding members
- enforce Marmot/MIP-01 compatibility checks (kind, base64 payload, ciphersuite, required MLS extensions)
- verify key package author integrity by requiring event pubkey to match the expected member pubkey

## Why
- prevent incompatible or malformed key packages from reaching MDK, so errors are clearer and failures happen earlier at the app boundary
- reduce trust on relay-provided events by validating protocol-level expectations before use

## Summary by CodeRabbit

* **Bug Fixes**
  * Stricter validation of fetched member key packages during group creation and member addition to reject incompatible or malformed packages early.
  * More specific, user-facing error responses when key package validation, encoding, or compatibility checks fail.

* **Performance**
  * Reused cryptographic/context setup across group operations to avoid redundant initialization.

* **Tests**
  * Added tests covering valid and invalid key package compatibility, encoding, and edge cases.

### DIFF
--- a/src/whitenoise/key_packages.rs
+++ b/src/whitenoise/key_packages.rs
@@ -49,45 +49,42 @@ pub(crate) fn has_encoding_tag(event: &Event) -> bool {
 pub(crate) fn validate_marmot_key_package_tags(
     event: &Event,
     expected_ciphersuite: &str,
-) -> core::result::Result<(), String> {
+) -> Result<()> {
     if event.kind != Kind::MlsKeyPackage {
-        return Err(format!(
-            "expected kind {}, got {}",
-            Kind::MlsKeyPackage,
-            event.kind
-        ));
+        return Err(WhitenoiseError::InvalidEventKind {
+            expected: Kind::MlsKeyPackage.to_string(),
+            got: event.kind.to_string(),
+        });
     }
 
     if !has_encoding_tag(event) {
-        return Err("missing required encoding tag ['encoding','base64']".to_string());
+        return Err(WhitenoiseError::MissingEncodingTag);
     }
 
-    Base64::decode_vec(&event.content)
-        .map_err(|e| format!("invalid base64 key package content: {}", e))?;
+    Base64::decode_vec(&event.content)?;
 
     let expected_ciphersuite = expected_ciphersuite.to_ascii_lowercase();
     let advertised_ciphersuites = normalized_tag_values(event, TagKind::MlsCiphersuite);
     if !advertised_ciphersuites.contains(&expected_ciphersuite) {
-        return Err(format!(
-            "incompatible mls_ciphersuite: expected {}, got [{}]",
-            expected_ciphersuite,
-            advertised_ciphersuites.join(", ")
-        ));
+        return Err(WhitenoiseError::IncompatibleMlsCiphersuite {
+            expected: expected_ciphersuite,
+            advertised: advertised_ciphersuites,
+        });
     }
 
     let extensions: HashSet<String> = normalized_tag_values(event, TagKind::MlsExtensions)
         .into_iter()
         .collect();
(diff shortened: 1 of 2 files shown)


src/whitenoise/groups.rs
```diff
@@ -134,6 +136,23 @@ impl Whitenoise {
             let event = some_event.ok_or(WhitenoiseError::MdkCoreError(
                 mdk_core::Error::KeyPackage("Does not exist".to_owned()),
             ))?;
+
+            if event.pubkey != *pk {
+                return Err(WhitenoiseError::InvalidInput(format!(
+                    "Fetched key package event {} signed by {} instead of expected {}",
+                    event.id, event.pubkey, pk
+                )));
+            }
```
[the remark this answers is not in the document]
erskingardner: Events are already validated by the nostr-sdk crate that manages all nostr operations for whitenoise-rs

src/whitenoise/key_packages.rs
```diff
@@ -34,6 +41,71 @@ pub(crate) fn has_encoding_tag(event: &Event) -> bool {
...
+    expected_ciphersuite: &str,
+) -> core::result::Result<(), String> {
+    if event.kind != Kind::MlsKeyPackage {
+        return Err(format!(
+            "expected kind {}, got {}",
+            Kind::MlsKeyPackage,
+            event.kind
+        ));
+    }
+
```
jgmontoya, line 59: I think we're missing a test for this case

src/whitenoise/key_packages.rs
```diff
@@ -34,6 +41,71 @@ pub(crate) fn has_encoding_tag(event: &Event) -> bool {
...
+    if !has_encoding_tag(event) {
+        return Err("missing required encoding tag ['encoding','base64']".to_string());
+    }
```
jgmontoya, line 62: Same here, regarding the tests

src/whitenoise/groups.rs
```diff
@@ -81,6 +82,7 @@ impl Whitenoise {
         group_type: Option<GroupType>,
     ) -> Result<group_types::Group> {
         let signer = self.get_signer_for_account(creator_account)?;
+        let mdk = self.create_mdk_for_account(creator_account.pubkey)?;
```
jgmontoya: Why move this to the top if it's not needed until much later? If some of the validations don't pass then we can avoid creating the instance altogether

src/whitenoise/groups.rs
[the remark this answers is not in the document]
erskingardner: I re-verified directly against the exact crate versions in our lockfile:

- nostr-sdk 0.44.1
- nostr-relay-pool 0.44.0
- nostr 0.44.2

### What I found:

- In nostr, JSON parsing itself does not verify signatures:
  - nostr-0.44.2/src/event/mod.rs:289 says from_json doesn’t verify signatures.
  - nostr-0.44.2/src/message/relay.rs:409 says RelayMessage::from_json doesn’t verify event signatures.
- In nostr-relay-pool, inbound relay events are signature-verified before being accepted/forwarded:
  - nostr-relay-pool-0.44.0/src/relay/inner.rs:1106 parses relay messages.
  - nostr-relay-pool-0.44.0/src/relay/inner.rs:1230 calls event.verify()? for new events (DatabaseEventStatus::NotExistent).
- nostr-sdk fetch_events_from goes through relay pool:
  - nostr-sdk-0.44.1/src/client/mod.rs:770 -> delegates to pool fetch_events_from.
- verify_subscriptions is separate (filter/subscription matching), not signature verification:
  - nostr-sdk-0.44.1/src/client/options.rs:129
  - defaults to false via Default-derived options.

PR #568, opened 2026-03-06
fix(messages): retry cascade delivery failure DB ops on SQLite lock errors

### LINKED ISSUE #554: Cascade delivery failure operations silently swallow errors
## Context

Identified during review of PR #519 (*feat(messages): replace fire-and-forget publishing with tracked delivery status*) by @jgmontoya.

**Original discussion:** https://github.com/marmot-protocol/whitenoise-rs/pull/519

## Problem

`cascade_delivery_failure` in `src/whitenoise/messages.rs` handles reverting optimistic side-effects when a message publish fails:

- **Kind 7 (reaction)**: Removes the reaction from the parent message's reaction summary
- **Kind 5 (deletion)**: Clears `deletion_event_id` on targets so they reappear

When these cascade operations fail (e.g., database update errors), they log warnings but don't propagate errors or retry. This means:

- A failed reaction removal could leave the reaction permanently visible in the parent's reaction summary even though delivery failed
- A failed deletion unmark could leave messages incorrectly hidden after the deletion event failed to publish

## Current Behavior

```rust
// Logs error but continues — side-effect is not reverted
if let Err(e) = AggregatedMessage::remove_reaction_from_message(...).await {
    tracing::warn!("Failed to remove reaction from parent: {e}");
}
```

## Expected Behavior

Failed cascade operations should either:
- Retry the database operation (with backoff)
- Propagate the error so the caller can handle it
- Queue a repair/reconciliation task

The exact approach needs design consideration — this issue tracks the gap so it isn't forgotten.

## Impact

Low probability (requires a DB error during cascade), but the result is inconsistent UI state that the user cannot easily fix.

mubarakcoded:
## Summary

Closes #554

- Add inline retry with linear backoff (3 attempts, 100ms × attempt) to `cascade_delivery_failure` for kind 7 (reaction) and kind 5 (deletion) DB operations
- Extract `DatabaseError::is_sqlite_lock_error()` method and refactor `delete_all_data` to use it
- Escalate exhausted retry logs from `warn!` to `error!` for visibility

## Context

During PR #519 review, @jgmontoya identified that `cascade_delivery_failure` silently swallows DB errors. When the cascade DB operations fail (e.g., SQLite BUSY/LOCKED), the method logged a warning and returned — leaving the UI inconsistent (phantom reaction or hidden message).

While SQLite's WAL mode + 5-second `busy_timeout` makes lock errors unlikely for these single-UPDATE statements, the retry provides defense in depth and the `error!` escalation ensures any persistent failures are surfaced.

## Test plan

- [x] `just check-clippy` — clean
- [x] `just fmt` — clean
- [x] `just test` — 1216 passed, 0 failed
- [x] `just int-test` — 18/18 scenarios, 181/181 test cases
- [x] 5 new unit tests for `is_sqlite_lock_error()` (busy, locked, row-not-found, other-code, serialization)
- [x] Existing cascade tests pass unchanged (retry transparent on first-attempt success)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cascade reliability for reactions and deletions by adding automatic retries on transient database lock errors, preventing panics and reducing failed cascades.
  * Better detection of transient DB lock conditions and clearer error-level logging when retries are exhausted; cascade flow now continues gracefully on persistent failures.

* **Tests**
  * Added tests covering retry flows, cascade outcomes, transient vs permanent DB errors, and logging behavior.

### DIFF
--- a/src/whitenoise/database/mod.rs
+++ b/src/whitenoise/database/mod.rs
@@ -248,24 +248,24 @@ impl Database {
 
 /// Retry an async database operation on transient SQLite lock errors.
 ///
-/// Uses linear backoff (100ms × attempt) for up to 3 attempts.
+/// Uses linear backoff (100 ms × attempt) for up to 3 attempts.
 /// Returns on first success or first non-lock error.
-pub async fn retry_on_lock<F, Fut, T>(mut op: F) -> std::result::Result<T, DatabaseError>
+pub(crate) async fn retry_on_lock<F, Fut, T>(mut op: F) -> std::result::Result<T, DatabaseError>
 where
     F: FnMut() -> Fut,
     Fut: std::future::Future<Output = std::result::Result<T, DatabaseError>>,
 {
-    const MAX_RETRIES: u32 = 3;
+    const MAX_ATTEMPTS: u32 = 3;
     let mut attempt: u32 = 0;
 
     loop {
         attempt += 1;
         match op().await {
             Ok(val) => return Ok(val),
-            Err(e) if e.is_sqlite_lock_error() && attempt < MAX_RETRIES => {
+            Err(e) if e.is_sqlite_lock_error() && attempt < MAX_ATTEMPTS => {
                 tracing::warn!(
                     target: "whitenoise::database",
-                    "SQLite lock on attempt {attempt}/{MAX_RETRIES}, \
+                    "SQLite lock on attempt {attempt}/{MAX_ATTEMPTS}, \
                      retrying...",
                 );
                 tokio::time::sleep(Duration::from_millis(100 * u64::from(attempt))).await;


src/whitenoise/messages.rs
```diff
@@ -399,36 +399,24 @@ impl Whitenoise {
...
-                    {
-                        tracing::warn!(
-                            target: "whitenoise::messages::delivery",
-                            "Failed to update reactions after cascade: {e}",
-                        );
-                        return;
-                    }
+                    .await?;
 
                     stream_manager.emit(
```
erskingardner: having this inside the retry closure will mean that we emit on each retry right? Don't we only want to emit once we're sure that we've actually removed the reaction?

src/whitenoise/database/mod.rs
```diff
@@ -244,6 +227,44 @@ impl Database {
...
+            }
+        }
+    }
+
+    Err(last_error.unwrap_or_else(|| {
+        DatabaseError::Sqlx(sqlx::Error::Protocol(
+            "Unexpected retry failure".to_string(),
+        ))
+    }))
+}
```
jgmontoya: Is this actually reachable? it seems like on the last iteration of the loop if there's an error it would get to the return Err(e) of line 256, no?

src/whitenoise/messages.rs
```diff
@@ -2485,4 +2493,118 @@ mod tests {
...
+            )
+            .await
+            .unwrap();
+
+        let reaction_event_id = reaction_result.message.id.to_string();
+        let tags = Tags::from_list(vec![Tag::parse(vec!["e", &target_id]).unwrap()]);
+        let stream_manager = MessageStreamManager::new();
+
+        // Close pool AFTER data is inserted — find_by_id will fail,
+        // exercising the error path
```
jgmontoya: This was raised by my agent:

> The doc-comment at the top of this test says it exercises "the second error branch" (i.e.,
> when `update_reactions` fails after a successful `find_by_id`). But both the inline comment
> and the actual behaviour disagree: closing the pool means `find_by_id` will fail on the
> first DB call, which is the same path tested by
> `test_cascade_reaction_failure_handles_db_error`. The `update_reactions` failure path (the
> stated goal) is not exercised.
> 
> To actually test the `update_reactions` failure branch you would need the reaction to be
> present in the cache (so `find_by_id` succeeds via an in-memory or mocked path), but the DB
> unavailable only for the write. With the current pool-close approach the two tests exercise
> the same code path. Either:
> 
> - Fix the comment to accurately describe what is being tested (find_by_id failure path with
>   real data pre-inserted), or
> - Remove the duplicate test, or
> - Find a way to inject a failure only on the `update_reactions` call.

src/whitenoise/database/mod.rs
```diff
@@ -248,6 +231,35 @@ impl Database {
...
+/// Retry an async database operation on transient SQLite lock errors.
+///
+/// Uses linear backoff (100ms × attempt) for up to 3 attempts.
+/// Returns on first success or first non-lock error.
+pub async fn retry_on_lock<F, Fut, T>(mut op: F) -> std::result::Result<T, DatabaseError>
+where
+    F: FnMut() -> Fut,
+    Fut: std::future::Future<Output = std::result::Result<T, DatabaseError>>,
+{
+    const MAX_RETRIES: u32 = 3;
```
jgmontoya: This should be called MAX_ATTEMPTS as retries makes it seem like there's going to be one more

src/whitenoise/database/mod.rs
```diff
@@ -248,6 +231,35 @@ impl Database {
     }
 }
 
+/// Retry an async database operation on transient SQLite lock errors.
+///
+/// Uses linear backoff (100ms × attempt) for up to 3 attempts.
+/// Returns on first success or first non-lock error.
+pub async fn retry_on_lock<F, Fut, T>(mut op: F) -> std::result::Result<T, DatabaseError>
```
jgmontoya: this should be pub(crate) imo

PR #581, opened 2026-03-07
feat: NIP-C7 reply support (q-tag + content stripping)

dannym-arx:
Implements NIP-C7 reply support in the message aggregator. \`q\` tags now take precedence over legacy \`e\` tags when detecting replies, and the leading \`nostr:nevent1...\` reference is stripped from reply content before it reaches the UI.

needed for https://github.com/marmot-protocol/whitenoise/issues/445

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable detection of reply targets by prioritizing explicit reply markers for accurate threading.
  * Cleaner message display by removing leading event-reference artifacts from replies.
  * Better handling of malformed or missing tags to avoid incorrect reply attribution and display issues.

* **Improvements**
  * Deleted messages are now detected and reflected in message display/state.
  * Replies now consistently link to their target and render cleaned content.

### DIFF
--- a/src/whitenoise/message_aggregator/processor.rs
+++ b/src/whitenoise/message_aggregator/processor.rs
@@ -197,20 +197,37 @@ fn extract_reply_info(tags: &Tags) -> Option<String> {
     None
 }
 
-/// Strip the leading `nostr:nevent1...\n` reference from reply content.
+/// Strip the first `nostr:nevent1...` reference from reply content.
 ///
-/// Per NIP-C7, reply content starts with a `nostr:nevent1...` URI followed by a newline.
-/// This is a protocol-level detail that should not be shown to the user.
+/// Per NIP-C7, reply content includes a `nostr:nevent1...` URI citing the replied-to event.
+/// The NIP does not require this reference to appear at the start of the content, so we scan
+/// for the first occurrence and remove it. Any surrounding whitespace (newlines/spaces) that
+/// was used to separate the reference from the rest of the message is also trimmed.
+///
+/// This is a protocol-level citation that should not be shown to the user — the UI renders
+/// the replied-to message inline above the reply body.
 fn strip_reply_event_reference(content: &str) -> String {
-    if !content.starts_with("nostr:nevent1") {
+    // Find the start of the first nostr:nevent1 token
+    let Some(start) = content.find("nostr:nevent1") else {
         return content.to_string();
-    }
+    };
 
-    content
-        .find('\n')
-        .map(|pos| &content[pos + 1..])
-        .unwrap_or("")
-        .to_string()
+    // Find the end of the token: whitespace or end of string terminates the URI
+    let end = content[start..]
+        .find(|c: char| c.is_ascii_whitespace())
+        .map(|rel| start + rel)
+        .unwrap_or(content.len());
+
+    // Build the stripped string: content before the token + content after the token
+    let before = content[..start].trim_end();
+    let after = content[end..].trim_start();
+
+    match (before.is_empty(), after.is_empty()) {
+        (true, true) => String::new(),
+        (true, false) => after.to_string(),
+        (false, true) => before.to_string(),
+        (false, false) => format!("{before}\n{after}"),
+    }
 }
 
 /// Try to process deletion message (kind 5)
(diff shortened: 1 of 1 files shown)


src/whitenoise/message_aggregator/processor.rs
```diff
@@ -137,17 +137,24 @@ pub(crate) async fn process_regular_message(
...
+    // Check if this is a reply (NIP-C7 q-tag, or legacy e-tag)
     let reply_to_id = extract_reply_info(&message.tags);
     let is_reply = reply_to_id.is_some();
 
+    // NIP-C7: strip the leading nostr:nevent1... reference from reply content
+    let (content, content_tokens) = if is_reply {
+        strip_reply_event_reference(&message.content, content_tokens)
+    } else {
+        (message.content.clone(), content_tokens)
+    };
```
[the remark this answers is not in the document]
dannym-arx: `strip_reply_event_reference` strips the leading `Nostr(...)` and `LineBreak` tokens, keeping `content` and `content_tokens` in sync.

In production, `NostrManager`'s parser emits a `Nostr` token for `nostr:nevent1...` URIs, so the filter matches and the token stream stays aligned.

`MockParser` is a different story -- it emits one `Text("nostr:nevent1abc\nHello")` token instead of `[Nostr(...), LineBreak, Text("Hello")]`, so the filter never matches. That's technically true, but no existing test runs `MockParser` through the `process_regular_message` path with reply content. 

In production, this path never gets hit because the real parser tokenizes correctly.

src/whitenoise/message_aggregator/processor.rs
```diff
@@ -347,19 +425,78 @@ mod tests {
...
 
+    #[test]
+    fn test_extract_reply_info_malformed_q_tag() {
+        let mut tags = Tags::new();
+        if let Ok(tag) = Tag::parse(vec!["q"]) {
+            tags.push(tag);
+        }
+
+        let reply_to_id = extract_reply_info(&tags);
+        assert!(reply_to_id.is_none());
```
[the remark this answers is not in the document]
dannym-arx: `Tag::parse` only fails on an empty vec (`Error::EmptyTag`). A single-element vec like `["e"]` always succeeds -- it creates a tag with kind `e` but no content. From there, `tag.content()` returns `None` because it looks up index 1 in the buffer, which doesn't exist. The test correctly exercises the path where `extract_reply_info` encounters an `e`-tag (or `q`-tag) with no content value and returns `None`.

src/whitenoise/message_aggregator/processor.rs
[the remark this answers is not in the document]
dannym-arx: I guess that makes sense, fixing it

src/whitenoise/message_aggregator/processor.rs
```diff
@@ -183,6 +197,22 @@ fn extract_reply_info(tags: &Tags) -> Option<String> {
     None
 }
 
+/// Strip the leading `nostr:nevent1...\n` reference from reply content.
+///
+/// Per NIP-C7, reply content starts with a `nostr:nevent1...` URI followed by a newline.
+/// This is a protocol-level detail that should not be shown to the user.
+fn strip_reply_event_reference(content: &str) -> String {
+    if !content.starts_with("nostr:nevent1") {
```
erskingardner: The NIP actually doesn't require that the `nostr:nevent...` needs to be at the beginning of the message. So I think that we need to scan for the first instance of the `nostr:nevent...` that matches the `q` tag and just remove it. We'll always render the reply at the top of the message.

src/whitenoise/message_aggregator/processor.rs
```diff
@@ -183,6 +197,22 @@ fn extract_reply_info(tags: &Tags) -> Option<String> {
...
+        return content.to_string();
+    }
+
+    content
+        .find('\n')
+        .map(|pos| &content[pos + 1..])
+        .unwrap_or("")
+        .to_string()
+}
+
```
jgmontoya, line 232: I'm not sure we can assume the `nostr:nevent1` to be exactly at the start

jgmontoya: woops comment ended up on some random lines for some reason, and also duplicated as Jeff beat me to it

src/whitenoise/message_aggregator/processor.rs
dannym-arx: :+1: makes sense to me, will fix tomorrow morning!

PR #582, opened 2026-03-07
Cut over long-lived subscriptions to relay planes

erskingardner:
## Summary
- move long-lived discovery, group, and account inbox subscription ownership onto relay-control planes
- route relay-plane events with typed source context instead of parsing subscription id prefixes
- remove the unused legacy NostrManager subscription orchestration and keep discovery batching at 500 authors per subscription

## Testing
- just precommit-quick

## Summary by CodeRabbit

* **New Features**
  * Discovery, group and inbox planes added to centralize relay subscriptions; per-account activation/deactivation and batched discovery sync.

* **Refactor**
  * Event routing now includes explicit event source metadata for clearer routing and retry behavior; subscription activation unified through the relay control plane.

* **Stability**
  * Improved relay session management, telemetry, relay policies and subscription lifecycle; startup now eagerly initializes discovery.

* **Behavior**
  * Logout and subscription refresh flows simplified to use centralized relay control paths.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -27,7 +27,7 @@ services:
         condition: service_completed_successfully
 
   blossom:
-    image: ghcr.io/hzrd149/blossom-server:master
+    image: ghcr.io/hzrd149/blossom-server@sha256:efef9fa3ef47934aff586dd2161f221bd9549f0ace4c3432f65e1afb5d3f4d0a
     volumes:
       - ./dev/data/blossom-ts-server:/app/data
       - ./dev/blossom-server-ts-config.yml:/app/config.yml


src/relay_control/mod.rs
```diff
@@ -68,12 +86,83 @@ impl RelayControlPlane {
...
+        watched_users: &[PublicKey],
+        follow_list_accounts: &[(PublicKey, Option<nostr_sdk::Timestamp>)],
+        public_since: Option<nostr_sdk::Timestamp>,
+    ) -> NostrResult<()> {
+        self.discovery
+            .sync(watched_users, follow_list_accounts, public_since)
+            .await
+    }
+
+    pub(crate) async fn activate_account_subscriptions(
```
jgmontoya, line 123: The inbox plane is activated (subscriptions created) before being inserted into the map. If activation succeeds but insertion fails or the task is cancelled between these operations, you have orphaned subscriptions receiving events with no plane to route them. Consider either:

1. Insert the plane first, then activate (requires plane to support an "inactive" state)
2. Use a transaction-like pattern where activation failure triggers cleanup
3. Document why this ordering is safe despite the apparent race

src/relay_control/groups.rs
```diff
@@ -21,9 +32,106 @@ impl Default for GroupPlaneConfig {
...
+        config.reconnect_policy = RelaySessionReconnectPolicy::FreshnessBiased;
+
+        Self {
+            session: RelaySession::new(config, event_sender),
+            session_salt,
+            accounts: RwLock::new(HashMap::new()),
+        }
+    }
+
+    pub(crate) async fn update_account(
```
jgmontoya, line 68: You unsubscribe before checking if the account is even in the map. On first call for a new account, this unnecessarily attempts to unsubscribe from a non-existent subscription. More concerning: if `update_account` is called concurrently for the same pubkey, both could unsubscribe, then both could subscribe, leading to duplicate active subscriptions with only one tracked in state. Consider checking `accounts.contains_key(&pubkey)` before unsubscribing, or using the lock throughout the update.

src/relay_control/sessions/session.rs
```diff
@@ -1,19 +1,546 @@
...
+
+    pub(crate) async fn set_signer(&self, signer: impl NostrSigner + 'static) {
+        self.client.set_signer(signer).await;
+    }
+
+    pub(crate) async fn unset_signer(&self) {
+        self.client.unset_signer().await;
+    }
+
+    pub(crate) async fn ensure_relays_connected(&self, relay_urls: &[RelayUrl]) -> Result<()> {
```
jgmontoya, line 104: If 2 out of 5 relays fail to connect, this silently succeeds but only returns the *last* error encountered (the other failure is lost). Callers have no way to know which relays failed. For production observability, consider:

1. Emit telemetry for each failed relay (you already do this in `subscribe_with_id_to`)
2. Log a warning listing all failed relays when partial success occurs
3. Return a structured result indicating which relays succeeded/failed

Without this, debugging "why isn't my subscription working?" will be difficult when the real issue is that 2 of 3 relays silently failed.

src/relay_control/sessions/session.rs
```diff
@@ -1,19 +1,546 @@
...
+            .cloned()
+            .collect();
+
+        for subscription_id in subscription_ids {
+            self.unsubscribe(&subscription_id).await;
+        }
+        self.client.unsubscribe_all().await;
+    }
+
+    fn spawn_notification_handler(&self, event_sender: Sender<ProcessableEvent>) {
```
jgmontoya, line 438: The spawned task is never tracked (no `JoinHandle` retained). On clean shutdown, there's no way to wait for this handler to finish processing queued notifications. If the process exits while notifications are in flight, events may be silently dropped. Consider:

1. Storing the `JoinHandle` in `RelaySessionState`
2. Adding a `shutdown()` method that cancels the client and waits for the handler to exit
3. At minimum, document that this is intentional and events in flight during shutdown are acceptable losses

src/relay_control/mod.rs
jgmontoya, line 123: This method activates both group and inbox planes. If group activation succeeds but inbox activation fails, what's the expected state? The method returns an error, but the group plane is already active and subscribed. Callers might reasonably assume that on error, no activation occurred. Add doc comments explaining:

- That activation is not atomic across planes
- On error, group subscriptions may be active even though inbox failed
- Whether callers should call `deactivate_account_subscriptions` to clean up partial state

src/relay_control/mod.rs
```diff
@@ -68,12 +86,83 @@ impl RelayControlPlane {
...
+        self.group_plane
+            .update_account(account_pubkey, group_relays, group_ids, since)
+            .await?;
+
+        let plane = account_inbox::AccountInboxPlane::new(
+            account_inbox::AccountInboxPlaneConfig::new(account_pubkey, inbox_relays.to_vec()),
+            self.event_sender.clone(),
+            self.session_salt,
+        );
+        let _ = user_relays;
```
jgmontoya: This parameter is passed but explicitly ignored. This suggests incomplete migration or future work. Either:
1. Remove it from the signature if it's truly unused
2. Add a TODO comment explaining when/why it will be used
3. Add a doc comment explaining this is for API compatibility during migration

Without context, reviewers (and future maintainers) will wonder if this is a bug.

src/relay_control/mod.rs
jgmontoya, line 1: The codebase uses both `pubkey_hash()` methods in `AccountInboxPlane` and `GroupPlane` that compute identical SHA256(salt || pubkey) hashes, but they're not shared code. Consider extracting to a shared utility function with clear documentation of its purpose (subscription ID generation). This would:

1. Ensure consistency if the hashing algorithm needs to change
2. Make it clear this is for subscription IDs, not security
3. Reduce duplication

Example:
```rust
// In relay_control/mod.rs or a utils module
pub(crate) fn subscription_id_hash(session_salt: &[u8; 16], pubkey: &PublicKey) -> String {
    let mut hasher = Sha256::new();
    hasher.update(session_salt);
    hasher.update(pubkey.to_bytes());
    format!("{:x}", hasher.finalize())[..12].to_string()
}
```

src/relay_control/discovery.rs
jgmontoya, line 1: The `DiscoveryPlane` tracks subscription IDs internally but doesn't expose any metrics about how many users are being watched. For debugging and monitoring, consider adding:

```rust
pub(crate) async fn stats(&self) -> DiscoveryPlaneStats {
    let state = self.state.read().await;
    DiscoveryPlaneStats {
        public_subscriptions: state.public_subscription_ids.len(),
        follow_list_subscriptions: state.follow_list_subscription_ids.len(),
    }
}
```

This would help answer questions like "how many discovery subscriptions are active?" without needing to dig through logs.

src/relay_control/groups.rs
```diff
@@ -35,4 +143,13 @@ mod tests {
...
     }
+
+    #[tokio::test]
+    async fn test_group_plane_hash_is_stable_for_account() {
+        let (sender, _) = mpsc::channel(8);
+        let plane = GroupPlane::new(sender, [7; 16]);
+        let pubkey = Keys::generate().public_key();
+
+        assert_eq!(plane.pubkey_hash(&pubkey), plane.pubkey_hash(&pubkey));
+    }
```
jgmontoya, line 269: The test suite validates basic group plane operations but doesn't test concurrent calls to `update_account` for the same pubkey. Given the unsubscribe-then-resubscribe pattern, a race condition test would be valuable:

```rust
#[tokio::test]
async fn test_concurrent_update_same_account_is_safe() {
    let (sender, _) = mpsc::channel(8);
    let plane = GroupPlane::new(sender, [42; 16]);
    let pubkey = Keys::generate().public_key();
    let relay = RelayUrl::parse("wss://relay.example.com").unwrap();
    let group_id = "test-group".to_string();

    // Fire off multiple updates concurrently
    let handles: Vec<_> = (0..10)
        .map(|_| {
            let plane = plane.clone(); // GroupPlane needs to be Clone or Arc
            let pubkey = pubkey;
            let relay = relay.clone();
            let group_id = group_id.clone();
            tokio::spawn(async move {
                plane.update_account(pubkey, &[relay], &[group_id], None).await
            })
        })
        .collect();

    for handle in handles {
        handle.await.unwrap().unwrap();
    }

    // Verify clean state: should have exactly one subscription
    assert!(plane.has_account(&pubkey).await);
}
```

src/types.rs
```diff
@@ -51,24 +54,43 @@ impl Default for RetryInfo {
 }
 
 /// Events that can be processed by the Whitenoise event processing system
+#[derive(Debug, Clone)]
+pub enum EventSource {
+    /// Legacy compatibility path that still identifies streams by subscription ID.
+    LegacySubscriptionId(Option<String>),
+    /// Relay-plane path with typed local routing context.
+    RelaySubscription(SubscriptionContext),
+}
```
jgmontoya, line 75: The enum itself is documented but the variant comments don't explain *when* each is used or how to choose. Expand the documentation:

```rust
/// Identifies where an event came from in the processing pipeline.
///
/// This enum supports gradual migration from legacy subscription management
/// to the relay-plane architecture. Events flow through one of two paths:
///
/// - **Legacy path**: Old NostrManager subscriptions identify streams by
///   string subscription IDs. The processor checks the ID prefix to determine
///   if it's a global or account-scoped event.
///
/// - **Relay-plane path**: New relay-control subscriptions carry full typed
///   routing context (plane, stream, account) directly from the relay session.
///
/// Once migration is complete, `LegacySubscriptionId` can be removed.
#[derive(Debug, Clone)]
pub enum EventSource {
    /// Legacy subscription identified only by a string ID.
    /// Used by old NostrManager subscriptions during migration.
    LegacySubscriptionId(Option<String>),

    /// Modern relay-plane subscription with full routing context.
    /// Used by new RelaySession subscriptions.
    RelaySubscription(SubscriptionContext),
}
```

src/relay_control/discovery.rs
```diff
@@ -48,6 +59,141 @@ impl DiscoveryPlaneConfig {
...
+
+            state.public_subscription_ids.clear();
+            state.follow_list_subscription_ids.clear();
+            stale_subscription_ids
+        };
+
+        for subscription_id in stale_subscription_ids {
+            self.session.unsubscribe(&subscription_id).await;
+        }
+
```
jgmontoya: Every call to `sync()` unsubscribes from *all* existing subscriptions then re-subscribes, even if the watched user set hasn't changed. For a user with 500 follows, this burns relay bandwidth and CPU on every sync. Consider:

1. Diffing the old and new user sets to only add/remove changed subscriptions
2. Implementing an incremental `add_users` / `remove_users` API
3. At minimum, document that this is intentional (simplicity over efficiency)

This is acceptable for now but will become a bottleneck if sync is called frequently.

src/integration_tests/test_cases/user_search/seeds.rs
dannym-arx, line 1: why was this removed?

src/nostr_manager/subscriptions.rs
dannym-arx, line 1: why not just delete the whole file, do we need to keep this file just for that utility function?

justfile
dannym-arx, line 1: you probably didn't want to commit this

dannym-arx (commented):
See inline comments already added.

src/integration_tests/test_cases/user_search/seeds.rs
erskingardner: it was moved.

src/nostr_manager/subscriptions.rs
erskingardner: we'll do clean up in a follow on PR. we're not quite done moving things.

justfile
erskingardner: why? this is a new e2e test.

src/relay_control/mod.rs
```diff
@@ -68,14 +90,163 @@ impl RelayControlPlane {
...
+            .write()
+            .await
+            .insert(account_pubkey, plane.clone());
+        if let Err(error) = plane.activate(inbox_relays, since, signer).await {
+            self.account_inbox_planes
+                .write()
+                .await
+                .remove(&account_pubkey);
+            plane.deactivate().await;
+            return Err(error);
```
dannym-arx: The plane goes into the map *before* `activate()` has run, so a concurrent caller (say, a relay-list event waking `refresh_account_subscriptions`) can see the entry and treat the account as healthy when it isn't.

The error path is also asymmetric. If inbox activation fails, `group_plane.update_account` (which already ran) isn't rolled back. The doc comment acknowledges this, but `setup_subscriptions` (the caller) doesn't call `deactivate_account_subscriptions` on error, so the group plane entry stays dirty.

src/relay_control/account_inbox.rs
```diff
@@ -31,11 +39,100 @@ impl AccountInboxPlaneConfig {
             plane: RelayPlane::AccountInbox,
             auth_policy: self.auth_policy,
             reconnect_policy: self.reconnect_policy,
+            relay_policy: RelaySessionConfig::new(RelayPlane::AccountInbox).relay_policy,
             connect_timeout: RelaySessionConfig::new(RelayPlane::AccountInbox).connect_timeout,
         }
     }
```
dannym-arx, line 42: `RelaySessionConfig::new(RelayPlane::AccountInbox)` is called twice to pull `relay_policy` and `connect_timeout` out of defaults. The cleaner approach is to call `new()` once, then override the two fields that differ. As written, if the defaults in `new()` ever change, this method silently diverges.

src/relay_control/discovery.rs
```diff
@@ -48,6 +62,239 @@ impl DiscoveryPlaneConfig {
...
+            state: RwLock::new(DiscoveryPlaneState::default()),
+        }
+    }
+
+    pub(crate) async fn start(&self) -> Result<()> {
+        self.session
+            .ensure_relays_connected(&self.config.relays)
+            .await
+    }
+
```
dannym-arx, line 103: `ensure_relays_connected` triggers connect + `wait_for_connection` every time sync runs, even when the relays are already live. That's an unnecessary round-trip on every discovery update. Even the empty-relay early-exit path goes through `start()` before it bails. Connection setup belongs in `start()` only; `sync()` should skip it when already connected.

src/relay_control/sessions/session.rs
```diff
@@ -1,19 +1,715 @@
...
+                        .await;
+                } else {
+                    tracing::error!(
+                        target: "whitenoise::relay_control::session",
+                        "Missing subscription context for relay {} subscription {}",
+                        relay_url,
+                        subscription_id
+                    );
+                }
+            }
```
dannym-arx: `spawn_notification_handler` handles `RelayPoolNotification::Event` directly in its outer match arm (lines 465–488) and only calls `process_notification` for `RelayPoolNotification::Message`. The `RelayNotification::Event` arm in `process_notification` can't be reached. It looks active, which is the problem. Either remove it or add a comment explaining why both paths exist.

src/relay_control/groups.rs
```diff
@@ -21,9 +34,225 @@ impl Default for GroupPlaneConfig {
...
+    pub(crate) async fn update_account(
+        &self,
+        pubkey: PublicKey,
+        relays: &[RelayUrl],
+        group_ids: &[String],
+        since: Option<Timestamp>,
+    ) -> Result<()> {
+        let _update_guard = self.update_lock.lock().await;
+        let subscription_id =
+            SubscriptionId::new(format!("{}_mls_messages", self.pubkey_hash(&pubkey)));
```
dannym-arx, line 77: @coderabbitai 

One subscription per account means a network observer can see the full set of group IDs being tracked. `relay-control-plane-rearchitecture.md` flags this under "Revisit Before Completion" but it's just a doc note. This should have a real tracking issue before the PR merges, otherwise it stays invisible, please make it one

src/relay_control/groups.rs
```diff
@@ -21,9 +34,225 @@ impl Default for GroupPlaneConfig {
...
+                .cmp(&right.account_pubkey)
+                .then(left.group_id.cmp(&right.group_id))
+        });
+        let known_relays = distinct_group_relays
+            .into_iter()
+            .map(|relay_url| {
+                RelayUrl::parse(&relay_url)
+                    .unwrap_or_else(|error| panic!("invalid group relay in snapshot: {error}"))
+            })
+            .collect::<Vec<_>>();
```
dannym-arx: These URLs were `RelayUrl` values when they went into `distinct_group_relays`. They got converted to `String` earlier in the snapshot path, and now they're being parsed back. The round-trip should always succeed, but a panic on a debug diagnostic endpoint is bad. Keep them as `RelayUrl` throughout and the parse never needs to happen.

src/relay_control/account_inbox.rs
```diff
@@ -31,11 +39,100 @@ impl AccountInboxPlaneConfig {
...
+            session_salt,
+        }
+    }
+
+    pub(crate) async fn activate(
+        &self,
+        inbox_relays: &[RelayUrl],
+        since: Option<Timestamp>,
+        signer: std::sync::Arc<dyn NostrSigner>,
+    ) -> Result<()> {
```
dannym-arx, line 70: this is minor, not sure if you wanna fix it, but: 

`activate()` takes `Arc<dyn NostrSigner>` and passes it to `set_signer()`, which takes `impl NostrSigner + 'static`. It works because `Arc<T: Trait>: Trait`, but it's not obvious. Someone reading `activate()` without knowing that coercion will wonder why the types don't line up. Document it or make both signatures consistent.

PR #585, opened 2026-03-08
feat(messages): add cursor-based pagination to fetch_aggregated_messages_for_group

dannym-arx:
marmot

Relates to marmot-protocol/whitenoise#122.

Adds `before: Option<Timestamp>` and `limit: Option<u32>` parameters to `fetch_aggregated_messages_for_group` so the Flutter client can implement infinite scroll without pulling all messages into memory at once. When no arguments are supplied the API returns the 50 most recent messages; passing the `created_at` of the oldest message already held pages backwards through history one chunk at a time.

## Summary by CodeRabbit

* **New Features**
  * Cursored pagination for message retrieval: before timestamp, companion message cursor, and per-request limit (default 50, cap 200).
  * CLI support for pagination flags to request paged results.
* **Improvements**
  * Direct single-message lookup for faster access.
  * Paginated endpoint returns stable oldest-first pages for clients.
* **Bug Fixes**
  * Validation rejects incomplete cursors to prevent inconsistent paging.
* **Tests**
  * Expanded coverage for pagination, limits, edge cases, and tied-timestamp ordering.

### DIFF
--- a/src/whitenoise/database/aggregated_messages.rs
+++ b/src/whitenoise/database/aggregated_messages.rs
@@ -271,7 +271,7 @@ impl AggregatedMessage {
         // so we multiply by 1000.  When the cursor is absent we use i64::MAX as a sentinel
         // so that `created_at < MAX` is unconditionally true and the single query branch
         // still uses the covering index.
-        let (before_ms, before_id_str): (i64, &str) = match (before, before_message_id) {
+        let (before_ms, before_id_str): (i64, String) = match (before, before_message_id) {
             // No cursor: use i64::MAX so `created_at < MAX` is unconditionally true for every
             // real timestamp, returning the newest page without a cursor filter.
             //
@@ -285,8 +285,19 @@ impl AggregatedMessage {
             // timestamp is a realistic value, so the second conjunct is unreachable and its
             // falseness is harmless.  Both bind slots must still be filled to satisfy sqlx's
             // parameter count, hence the empty string placeholder.
-            (None, None) => (i64::MAX, ""),
-            (Some(ts), Some(id)) => ((ts.as_secs() as i64).saturating_mul(1_000), id),
+            (None, None) => (i64::MAX, String::new()),
+            (Some(ts), Some(id)) => {
+                // Canonicalize to lowercase hex so the lexicographic tie-break comparison
+                // against stored message_ids (always lowercase) is stable regardless of the
+                // case the caller used.  An invalid or wrong-length ID is rejected here
+                // rather than silently producing an incorrect page boundary.
+                let canonical = EventId::from_hex(id).map(|eid| eid.to_hex()).map_err(|_| {
+                    DatabaseError::InvalidCursor {
+                        reason: "before_message_id is not a valid 64-character hex event ID",
+                    }
+                })?;
+                ((ts.as_secs() as i64).saturating_mul(1_000), canonical)
+            }
             (Some(_), None) => {
                 return Err(DatabaseError::InvalidCursor {
                     reason: "before_message_id is required when before is provided",
@@ -2795,6 +2806,43 @@ mod tests {
         assert!(
             matches!(err, DatabaseError::InvalidCursor { .. }),
             "expected InvalidCursor, got: {err}"
+        );
+    }
+
+    #[tokio::test]
+    async fn test_paginated_invalid_before_message_id_is_rejected() {
+        let (whitenoise, _data_temp, _logs_temp) = create_mock_whitenoise().await;
+        let group_id = GroupId::from_slice(&[208; 32]);
+        let ts = Timestamp::from(1_700_000_000u64);
+
+        // Non-hex string → InvalidCursor
+        let err = AggregatedMessage::find_messages_by_group_paginated(
+            &group_id,
+            &whitenoise.database,
+            Some(ts),
+            Some("not-valid-hex-at-all"),
+            None,
+        )
+        .await
+        .unwrap_err();
+        assert!(
+            matches!(err, DatabaseError::InvalidCursor { .. }),
+            "expected InvalidCursor for non-hex id, got: {err}"
+        );
+
+        // Valid hex but wrong length (32 chars instead of 64) → InvalidCursor
+        let err = AggregatedMessage::find_messages_by_group_paginated(
+            &group_id,
+            &whitenoise.database,
+            Some(ts),
+            Some("00000000000000000000000000000001"),
+            None,
+        )
+        .await
+        .unwrap_err();
+        assert!(
+            matches!(err, DatabaseError::InvalidCursor { .. }),
+            "expected InvalidCursor for short hex id, got: {err}"
         );
     }
 }


src/whitenoise/messages.rs
```diff
@@ -601,21 +601,28 @@ impl Whitenoise {
...
         &self,
         pubkey: &PublicKey,
         group_id: &GroupId,
+        before: Option<Timestamp>,
+        limit: Option<u32>,
     ) -> Result<Vec<ChatMessage>> {
         Account::find_by_pubkey(pubkey, &self.database).await?; // Verify account exists (security check)
 
-        AggregatedMessage::find_messages_by_group(group_id, &self.database)
+        AggregatedMessage::find_messages_by_group_paginated(group_id, before, limit, &self.database)
```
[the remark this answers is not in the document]
dannym-arx: I don't think this matters, the only users of whitenoise-rs are the flutter app and the cli, we're fine lol

src/whitenoise/database/aggregated_messages.rs
```diff
@@ -226,6 +228,92 @@ impl AggregatedMessage {
...
+        // Both fields must be present together or both absent.  A half-specified cursor
+        // (one Some, one None) is rejected immediately rather than falling back to lossy
+        // single-field logic that silently skips messages at a tied timestamp.
+        //
+        // The DB stores created_at as Unix milliseconds (i64); Timestamp is Unix seconds,
+        // so we multiply by 1000.  When the cursor is absent we use i64::MAX as a sentinel
+        // so that `created_at < MAX` is unconditionally true and the single query branch
+        // still uses the covering index.
+        let (before_ms, before_id_str): (i64, &str) = match (before, before_message_id) {
+            (None, None) => (i64::MAX, ""),
```
erskingardner: Don't know if this actually needs fixing but my agent flagged: 


When both cursor fields are absent, `before_ms = i64::MAX` and `before_id_str = ""` (the empty string). The SQL filter becomes:

```sql
AND (am.created_at < 9223372036854775807 OR (am.created_at = 9223372036854775807 AND am.message_id < ''))
```

The second conjunct (`am.message_id < ''`) is always false — no hex string is lexicographically less than the empty string in SQLite. The first conjunct (`created_at < i64::MAX`) handles normal timestamps correctly, so the visible behavior is fine *today*.

The real concern is the implicit meaning: the empty string `""` is being used as a "no lower bound on message_id" sentinel for a path that is supposed to be a pure "get the newest page" request. This works only because the first OR branch subsumes the second. It's a non-obvious invariant that could easily break if the SQL is refactored (e.g., the timestamp column changes semantics or the OR order changes). A cleaner approach is to use `LIMIT ?` with `created_at DESC` only when no cursor is provided (two separate query strings), or at minimum document this fragility explicitly with an inline comment that explains *why* the empty string doesn't matter here.

src/whitenoise/database/aggregated_messages.rs
```diff
@@ -226,6 +228,92 @@ impl AggregatedMessage {
...
+            (Some(ts), Some(id)) => ((ts.as_secs() as i64).saturating_mul(1_000), id),
```
erskingardner: Another agent flag: 

`ts.as_secs()` returns `u64`. Casting `u64` to `i64` with `as` silently wraps for values > `i64::MAX`. A timestamp beyond year 2262 would produce a large negative `before_ms`, causing the query to return no results instead of erroring. This isn't a real-world concern for dates stored by this app, but it's the kind of silent data-dependent bug that's very hard to diagnose in production. Use `i64::try_from(ts.as_secs()).map_err(|_| DatabaseError::InvalidCursor { reason: "timestamp out of i64 range" })?` or at least add a `saturating_as` cast with a comment explaining the range assumption.

src/whitenoise/messages.rs
```diff
@@ -601,24 +601,66 @@ impl Whitenoise {
...
+    /// * `pubkey`  - The public key of the user requesting messages
     /// * `group_id` - The group to fetch messages for
+    /// * `before`            - Cursor timestamp. If `Some`, return only messages with
+    ///   `created_at` strictly before this value **or** at the same second but with a
+    ///   lexicographically smaller message ID. Pass the `created_at` of the oldest message
+    ///   currently loaded to fetch the preceding page (infinite scroll upward). Omit for the
+    ///   initial load.
+    /// * `before_message_id` - Companion cursor ID. Pass the `id` of the same oldest message
+    ///   used for `before` so that ties at the same second are resolved deterministically.
+    ///   May be `None` when `before` is `None` or when timestamps are known to be unique.
```
erskingardner: The last sentence says `before_message_id` may be `None` when `before` is `Some` as long as timestamps are unique. But `find_messages_by_group_paginated` unconditionally returns `Err(DatabaseError::InvalidCursor)` when `before = Some` and `before_message_id = None`. The doc and the implementation contradict each other. The implementation's strict pairing is the right call — drop the "when timestamps are known to be unique" escape hatch from the doc comment, since callers can never guarantee timestamp uniqueness in advance.

src/whitenoise/messages.rs
```diff
@@ -601,24 +601,66 @@ impl Whitenoise {
...
+        AggregatedMessage::find_messages_by_group_paginated(
+            group_id,
+            &self.database,
+            before,
+            before_message_id,
+            limit,
+        )
+        .await
+        .map_err(|e| {
+            WhitenoiseError::from(anyhow::anyhow!("Failed to read cached messages: {}", e))
```
erskingardner: We should probably distinguish between different error types:

A malformed cursor from a caller produces a `WhitenoiseError` whose `Display` is `"Failed to read cached messages: Invalid cursor: before_message_id is required when before is provided"`. The `InvalidCursor` variant is lost — the Flutter client has no way to distinguish a cursor validation error from a DB I/O error. Consider propagating `DatabaseError` through `WhitenoiseError` via a dedicated variant (or `#[from]` if `WhitenoiseError` already has a `Database(DatabaseError)` arm), so the client can respond specifically to cursor errors without string-matching.

src/whitenoise/messages.rs
```diff
@@ -601,24 +601,66 @@ impl Whitenoise {
...
+
+    /// Fetch a single aggregated message by its event ID.
+    ///
+    /// Returns `None` if the message does not exist in the cache or belongs to a different group.
+    ///
+    /// # Arguments
+    /// * `pubkey`   - The public key of the requesting user (security check)
+    /// * `group_id` - The group the message belongs to
+    /// * `message_id` - Hex-encoded event ID of the message
+    pub async fn fetch_message_by_id(
```
erskingardner, line 658: `fetch_message_by_id` is used in production code (`unreact_to_message`) and is the primary fix for the pagination-scope regression. It has no unit tests. At minimum: one test verifying it returns `Some(message)` when the message exists, one returning `None` for an unknown ID, and one verifying the account-existence security check is enforced.

src/whitenoise/database/aggregated_messages.rs
dannym-arx: I'll... let people in 2262 know :laughing: 

for any time traveler reading this, I'm sorry for causing the war of 2260 with this pr

erskingardner (approved):
I think your failing tests here are ones that I have fixed in my big relay PR so I'm fine if you leave those failing here.

PR #591, opened 2026-03-08
Feat/archive chat

jgmontoya:
## Summary by CodeRabbit

* **New Features**
  * Archive and unarchive chats via CLI commands
  * View a separate archived chat list and subscribe to real-time archived-chat updates
  * Archival state is persisted so archived chats are surfaced separately from active chats

* **Bug Fixes**
  * Preserves delivery status for echoed outgoing messages

* **Documentation**
  * Changelog reorganized and new v0.2.1 entries added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - CLI message reactions (react/unreact) ([jgmontoya])
 - CLI message delete, retry, and reply commands ([jgmontoya])
 - CLI media commands for uploading files and images ([jgmontoya])
+- Chat archiving: archive/unarchive chats, list archived, subscribe to archived chat list updates ([jgmontoya])
 - Relay control plane with health monitoring scaffold ([erskingardner])
 - Relay observability persistence for connection metrics and failure tracking ([erskingardner])
 
@@ -19,7 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- CLI interface with account management, group chat, messaging, follows, profile, relays, settings, user search, notifications, and streaming subscriptions ([jgmontoya])
+- CLI with account management, group chat, messaging, follows, profile, relays, settings, user search, notifications, and streaming subscriptions ([jgmontoya])
 
 ### Fixed
 
--- a/src/whitenoise/mod.rs
+++ b/src/whitenoise/mod.rs
@@ -145,6 +145,7 @@ pub struct Whitenoise {
     message_aggregator: message_aggregator::MessageAggregator,
     message_stream_manager: Arc<message_streaming::MessageStreamManager>,
     chat_list_stream_manager: chat_list_streaming::ChatListStreamManager,
+    archived_chat_list_stream_manager: chat_list_streaming::ChatListStreamManager,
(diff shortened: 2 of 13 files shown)


CHANGELOG.md
```diff
@@ -9,34 +9,131 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
-- Integration test scenario (`login-flow`) covering happy path, no relays, publish defaults, custom relay, and cancel flows ([erskingardner])
-- KeyPackage lifecycle management with delayed cleanup: track published key packages from creation through Welcome consumption, then automatically clean up local MLS key material via a dedicated `ConsumedKeyPackageCleanup` scheduled task after a 30s quiet period ([mubarakcoded])
+- CLI message reactions (react/unreact) ([jgmontoya])
+- CLI message delete, retry, and reply commands ([jgmontoya])
+- CLI media commands for uploading files and images ([jgmontoya])
+- Chat archiving: archive/unarchive chats, list archived, subscribe to archived chat list updates ([jgmontoya])
+- Relay control plane with health monitoring scaffold ([erskingardner])
+- Relay observability persistence for connection metrics and failure tracking ([erskingardner])
+
+## [v0.2.1] - 2026-03-05
```
erskingardner, line 19: The `[Unreleased]` section now lists items from this PR (archive/unarchive) above `[v0.2.1]`. That's correct. However, `[v0.2.0]` and `[v0.1.0-alpha.5]` are listed as `2026-03-05` and `2025-05-02` respectively — the patch `v0.2.1` post-dates `v0.2.0` on the same calendar day, which is unusual and may indicate the dates were set to the same commit date rather than the actual release date. Worth verifying before tagging.

src/whitenoise/mod.rs
```diff
@@ -991,6 +1001,56 @@ impl Whitenoise {
...
+            match updates.try_recv() {
+                Ok(update) => {
+                    if update.item.archived_at.is_some() {
+                        items_map.insert(update.item.mls_group_id.clone(), update.item);
+                    } else {
+                        items_map.remove(&update.item.mls_group_id);
+                    }
+                }
+                Err(broadcast::error::TryRecvError::Empty) => break,
+                Err(broadcast::error::TryRecvError::Lagged(_)) => continue,
```
erskingardner: Flagged by my agent: 

both `subscribe_to_chat_list` and `subscribe_to_archived_chat_list`

When the broadcast buffer has been overrun the `Lagged` branch calls `continue`, which advances to the next loop iteration and tries `recv` again — but the channel has already skipped some messages. The correct behavior is to fall back to re-fetching the full snapshot, otherwise the initial items list can be permanently inconsistent with reality. The existing `subscribe_to_chat_list` has the same bug (predates this PR), but the archived variant copies it verbatim.

Options:
1. On `Lagged`, break out and re-fetch the full list (safest).
2. On `Lagged`, break out with a warning and accept that a future live update will eventually correct the item (acceptable for a chat list, and consistent with how the live stream handles it).

Either is better than silently continuing.

CHANGELOG.md
jgmontoya: I think this is correct, they were tagged the same day.

src/whitenoise/mod.rs
jgmontoya: I added some logging but other than that I don't think we need to care about this. For this to become a problem it needs to overflow the buffer in a very short time span

PR #592, opened 2026-03-08
Migrate one-off relay work to relay control

erskingardner:
## Summary
- move one-off fetches and one-off publishes onto relay control ephemeral sessions
- remove the legacy NostrManager query/publish compatibility modules and migrate remaining callers
- drop `wss://purplepag.es` from the curated discovery relay set and update related tests/docs

## Testing
- just precommit-quick

## Summary by CodeRabbit

* **New Features**
  * Ephemeral relay publishing layer introduced for more resilient, observable message delivery.
  * Account helpers added: list accounts, count, and find by public key.

* **Bug Fixes**
  * Delivery status updates now retry on transient DB locks.
  * Login tolerates transient relay-list publish failures when safe.

* **Refactor**
  * Publishing and query flows migrated to the new relay-control/ephemeral path.

* **Chores**
  * Removed an underperforming default relay from discovery.

### DIFF
--- a/src/relay_control/ephemeral.rs
+++ b/src/relay_control/ephemeral.rs
@@ -237,7 +237,10 @@ impl EphemeralPlane {
         account_pubkey: &PublicKey,
         relays: &[RelayUrl],
     ) -> Result<Output<EventId>> {
-        // Retry schedule is immediate first attempt, then 2 s and 4 s backoff by default.
+        // One session for the full retry group — connect once, retry the publish,
+        // then disconnect. Creating a new client per attempt would reconnect on
+        // every backoff cycle unnecessarily.
+        let session = self.spawn_session("publish", Some(*account_pubkey));
         let mut last_error: Option<NostrManagerError> = None;
 
         for attempt in 0..self.config.max_publish_attempts {
@@ -254,9 +257,7 @@ impl EphemeralPlane {
                 tokio::time::sleep(delay).await;
             }
 
-            let session = self.spawn_session("publish", Some(*account_pubkey));
             let result = session.publish_event_to(relays, &event).await;
-            session.shutdown().await;
 
             match result {
                 Ok(output) if !output.success.is_empty() => {
@@ -272,6 +273,7 @@ impl EphemeralPlane {
                         );
                     }
 
+                    session.shutdown().await;
                     return Ok(output);
                 }
                 Ok(output) => {
@@ -298,6 +300,7 @@ impl EphemeralPlane {
             }
         }
 
+        session.shutdown().await;
         Err(last_error.unwrap_or(NostrManagerError::NoRelayConnections))
     }
 
--- a/src/whitenoise/accounts/setup.rs
+++ b/src/whitenoise/accounts/setup.rs
@@ -33,7 +33,6 @@ impl Whitenoise {
         account: &Account,
         user: &User,
         is_new_account: bool,
-        _nip65_relays: &[Relay],
         inbox_relays: &[Relay],
         key_package_relays: &[Relay],
     ) -> Result<()> {
@@ -67,9 +66,7 @@ impl Whitenoise {
         &self,
         account: &Account,
         user: &User,
-        _nip65_relays: &[Relay],
         inbox_relays: &[Relay],
-        _key_package_relays: &[Relay],
     ) -> Result<()> {
         let (cancel_tx, _) = tokio::sync::watch::channel(false);
         self.background_task_cancellation
--- a/src/whitenoise/relays.rs
+++ b/src/whitenoise/relays.rs
@@ -158,7 +158,17 @@ impl Whitenoise {
                 .ok()
                 .flatten()
                 .map(|s| {
-                    if s.success_count > 0 {
+                    // A relay is currently connected when it has a recorded
+                    // success and that success is more recent than any failure.
+                    // Using timestamps rather than the cumulative success_count
+                    // prevents a relay that once connected but has since
+                    // disconnected from appearing as Connected indefinitely.
+                    let connected = match (s.last_connect_success_at, s.last_failure_at) {
+                        (Some(success), Some(failure)) => success > failure,
+                        (Some(_), None) => true,
+                        _ => false,
+                    };
+                    if connected {
                         RelayStatus::Connected
                     } else {
                         RelayStatus::Disconnected
--- a/src/whitenoise/accounts/login.rs
+++ b/src/whitenoise/accounts/login.rs
@@ -29,7 +29,7 @@ impl Whitenoise {
             .await?;
         tracing::debug!(target: "whitenoise::accounts", "Relays setup");
 
-        self.activate_account(&account, &user, true, &relays, &relays, &relays)
+        self.activate_account(&account, &user, true, &relays, &relays)
             .await?;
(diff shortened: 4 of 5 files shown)


src/whitenoise/database/aggregated_messages.rs
```diff
@@ -574,6 +576,47 @@ impl AggregatedMessage {
...
+
+        if let Some(error) = last_error {
+            match Self::update_delivery_status(message_id, group_id, status, database).await {
+                Ok(message) => Ok(message),
+                Err(final_error) if Self::is_database_lock_error(&final_error) => Err(error),
+                Err(final_error) => Err(final_error),
+            }
+        } else {
+            Self::update_delivery_status(message_id, group_id, status, database).await
+        }
```
jgmontoya: This `else` branch doesn't  seem reachable. If there was no error then it would've returned earlier, no?

src/relay_control/ephemeral.rs
```diff
@@ -17,6 +48,754 @@ impl Default for EphemeralPlaneConfig {
...
+                Err(_) => Ok(()),
+            };
+
+            session.shutdown().await;
+            tracking_result?;
+
+            match result {
+                Ok(output) if !output.success.is_empty() => return Ok(output),
+                Ok(output) => {
+                    last_error = Some(NostrManagerError::NoRelayConnections);
```
jgmontoya: This error makes it seem like the connection failed, but in reality it would be that the relays didn't accept the event right?

src/nostr_manager/mod.rs
```diff
@@ -141,7 +83,7 @@ impl NostrManager {
     pub(crate) async fn new(
         event_sender: Sender<crate::types::ProcessableEvent>,
         event_tracker: std::sync::Arc<dyn EventTracker>,
-        timeout: Duration,
+        _timeout: Duration,
```
jgmontoya: Why not just remove the parameter altogether?

src/relay_control/ephemeral.rs
```diff
@@ -17,6 +48,754 @@ impl Default for EphemeralPlaneConfig {
...
+    }
+
+    fn next_operation_id(&self) -> u64 {
+        self.operation_counter.fetch_add(1, Ordering::Relaxed) + 1
+    }
+
+    fn should_persist_telemetry(_telemetry: &RelayTelemetry) -> bool {
+        true
+    }
+
```
jgmontoya: What's the idea here? that we later turn it off on "production builds" or something?

erskingardner: yeah - this doesn't make much sense. we're going to keep persisting telemetry into the future. this should be removed.

src/nostr_manager/utils.rs
```diff
@@ -56,35 +56,35 @@ impl NostrManager {
...
+    tag.kind() == TagKind::SingleLetter(SingleLetterTag::lowercase(Alphabet::R))
+}
 
-    /// Checks if a tag is a "relay" tag.
-    pub(crate) fn is_relay_tag(tag: &Tag) -> bool {
-        tag.kind() == TagKind::Relay
-    }
+/// Checks if a tag is a "relay" tag.
+pub(crate) fn is_relay_tag(tag: &Tag) -> bool {
+    tag.kind() == TagKind::Relay
```
jgmontoya, line 87: why are these being taken 'out of' the NostrManager struct?

src/relay_control/sessions/session.rs
```diff
@@ -206,6 +221,83 @@ impl RelaySession {
...
+                        RelayTelemetry::new(
+                            RelayTelemetryKind::PublishFailure,
+                            self.config.plane,
+                            relay_url.clone(),
+                        )
+                        .with_message(error.to_string()),
+                    ));
+                }
+
+                Err(error.into())
```
jgmontoya, line 296: I think we're overdoing it with the telemetry but I guess we can strip some of that later

src/relay_control/ephemeral.rs
```diff
@@ -17,6 +47,747 @@ impl Default for EphemeralPlaneConfig {
...
+                    max_attempts = self.config.max_publish_attempts,
+                    ?delay,
+                    "Retrying ephemeral publish after failure"
+                );
+                tokio::time::sleep(delay).await;
+            }
+
+            let session = self.spawn_session("publish", Some(*account_pubkey));
+            let result = session.publish_event_to(relays, &event).await;
+            session.shutdown().await;
```
jgmontoya: On each retry we're:
- creating a new session
- publishing
- disconnecting

Is this ok or should we instead create a new session, then retry just the publishing and then disconnect?

src/whitenoise/relays.rs
```diff
@@ -145,20 +145,27 @@ impl Whitenoise {
...
+                )
+                .await
+                .ok()
+                .flatten()
+                .map(|s| {
+                    if s.success_count > 0 {
+                        RelayStatus::Connected
+                    } else {
+                        RelayStatus::Disconnected
+                    }
```
jgmontoya: This got flagged by my agent:

> ### Correctness: relay_status `success_count` increments on `Connected` but does not reset on `Disconnected`
> 
> **`src/whitenoise/database/relay_status.rs`, `upsert_from_telemetry`**
> 
> The new upsert SQL accumulates `success_count` additively and only ever increments it. A relay that connects and disconnects in a loop will show an ever-growing `success_count`. The code in `relays.rs` that derives relay status uses `s.success_count > 0` as the "connected" proxy:
> 
> ```rust
> if s.success_count > 0 {
>     RelayStatus::Connected
> } else {
>     RelayStatus::Disconnected
> }
> ```
> 
> This means once a relay has ever connected, it is shown as `Connected` forever, even after logout or after the relay becomes unreachable. The old code queried the live client pool status; the new code queries a historical counter. This is a functional regression in the relay status display. A `Disconnected` event should set a `is_currently_connected = false` flag (or clear `last_connect_success_at` relative to `last_failure_at`) that `relays.rs` can use to distinguish "ever connected" from "currently connected". The telemetry design doc mentions this as a known gap, but it is worth calling out explicitly since `relays.rs` code silently gives wrong answers today.

src/whitenoise/accounts/setup.rs
```diff
@@ -34,7 +33,7 @@ impl Whitenoise {
         account: &Account,
         user: &User,
         is_new_account: bool,
-        nip65_relays: &[Relay],
+        _nip65_relays: &[Relay],
```
jgmontoya: Why not just remove that? Preparing for full outbox in the future?

src/whitenoise/accounts/setup.rs
```diff
@@ -76,23 +67,14 @@ impl Whitenoise {
         &self,
         account: &Account,
         user: &User,
-        nip65_relays: &[Relay],
+        _nip65_relays: &[Relay],
         inbox_relays: &[Relay],
-        key_package_relays: &[Relay],
+        _key_package_relays: &[Relay],
```
jgmontoya: same here

src/relay_control/sessions/session.rs
erskingardner: Agreed — leaving as-is for now, can trim later if it proves noisy in practice.

PR #593, opened 2026-03-08
fix(groups): remove admin status when member is removed

### LINKED ISSUE #514: Removing someone from a group doesn't remove them from the list of admins
Create a group, add someone, make them an admin, then remove them. Then add them back and they show back up in the group as an admin again. 

Not entirely sure it's a "bug" per se but it's not what I expected.

mubarakcoded:
## Summary
- When removing a member from a group via `remove_members_from_group`, any admin status held by that member is now also removed from the group data
- Previously, removed members retained their admin entry in the MLS group data extension, causing them to reappear as admins if re-added to the group

Closes #514

## Test plan
- [x] New unit test `test_remove_member_also_removes_admin_status` verifies the fix end-to-end
- [x] `just precommit-quick` passes (fmt, docs, clippy, tests)

## Summary by CodeRabbit

* **Bug Fixes**
  * Administrators removed from a group now also lose admin status in the group's admin list.
* **Improvements**
  * Admin-list synchronization is performed as a separate update step; if it fails it is logged but does not block member removal.
* **Tests**
  * Added a test ensuring removed admins lose admin rights while the creator remains an admin.

### DIFF
--- a/src/whitenoise/groups.rs
+++ b/src/whitenoise/groups.rs
@@ -576,6 +576,8 @@ impl Whitenoise {
     /// 1. Creates an MLS remove members proposal
     /// 2. Publishes the evolution event to relays (with retry)
     /// 3. Only after relay acceptance, merges the pending commit locally
+    /// 4. If any removed members were admins, updates the group
+    ///    data to remove them from the admin list (logged on failure)
     ///
     /// Per MIP-03, the evolution event is published to relays *before* merging
     /// the pending commit locally. This ensures we only advance local state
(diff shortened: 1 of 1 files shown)


src/whitenoise/groups.rs
```diff
@@ -550,7 +552,46 @@ impl Whitenoise {
...
+        }
+
+        Ok(())
+    }
+
+    /// Removes any of the given `removed_members` from the group's admin list.
+    ///
+    /// This is a separate MLS evolution commit from the membership removal
+    /// itself, so callers should treat failures as non-fatal.
+    async fn strip_removed_members_from_admins(
```
erskingardner, line 631: I'm not sure this is the right way to solve this. 

1. It creates two back to back commits in rapid succession which is more likely to result in out of order commit processing for clients. 
2. It's not atomic - failures can result in a state where the member is removed but is still an admin. 

I think that we actually need to do this as a single atomic commit containing both changes, which will require changes upstream in MDK.

mubarakcoded: @erskingardner Good call — two back-to-back commits is indeed problematic for ordering.

I've pushed an upstream fix in MDK that handles this atomically: https://github.com/marmot-protocol/mdk/pull/225

It uses OpenMLS's `CommitBuilder` to bundle the removal proposals + a `GroupContextExtensions` update (stripping removed admins) into a single MLS commit. Once that merges, I'll update this PR to use the new MDK rev and remove the best-effort workaround.

erskingardner: @mubarakcoded does this PR need to closed then? Or is it still valid?

mubarakcoded: @erskingardner Waiting for the MDK fix (marmot-protocol/mdk#225) to be merged and confirmed, then I'll close this PR.

mubarakcoded:
Closing — the upstream fix landed in MDK (marmot-protocol/mdk#225). A rev bump will pick it up automatically, no whitenoise-rs changes needed.

PR #594, opened 2026-03-09
Fix TOCTOU race and orphaned discovery subscriptions in since-timestamp computation

### LINKED ISSUE #590: Investigate repeated 'No minimum last_synced_at found' discovery warning
We are seeing repeated logs like:

```
WARN whitenoise::setup_global_users_subscriptions: No minimum last_synced_at found; defaulting to since=None
```

This should be investigated separately from the relay-control migration work.

What the current code does:

- `last_synced_at` is still advanced for account-scoped events in `src/whitenoise/event_processor/account_event_processor.rs`.
- Successful `Kind::ContactList` and `Kind::MlsGroupMessage` processing call `Account::update_last_synced_max(...)`.
- Successful `Kind::GiftWrap` processing also calls `Account::update_last_synced_max(...)`, using the extracted rumor timestamp.
- The underlying DB write lives in `src/whitenoise/database/accounts.rs` in `Account::update_last_synced_max(...)`.

What does *not* update it:

- Global discovery events do not advance `last_synced_at`.
- The new relay-control discovery plane only reads it to compute `since`; it does not write it.

Why this warning looks suspicious:

- The warning is emitted in `src/whitenoise/mod.rs` inside `compute_global_since_timestamp()`.
- That function first checks `if accounts.iter().any(|a| a.last_synced_at.is_none())` and, in that case, logs an INFO line about unsynced accounts and returns `None` early.
- The WARN only happens later if no minimum timestamp can be computed at all.

Given the current logic, the WARN does not look like the normal "some accounts are unsynced" case. It appears more likely to mean one of:

1. `sync_discovery_subscriptions()` is being called through a path where account state is unexpectedly empty or inconsistent.
2. Account loading / persistence is not reflecting the expected `last_synced_at` state.
3. The warning is effectively unreachable in the normal path and should be replaced with a more precise log or assertion once root cause is known.

Suggested investigation:

- Confirm the exact runtime conditions under which `compute_global_since_timestamp()` reaches the WARN branch.
- Log/account-count instrumentation around `Account::all(...)` in that path.
- Verify whether any login / account deletion / startup sequencing can briefly leave discovery refresh running with no accounts.
- Decide whether the WARN branch should be downgraded/removed if it is not a meaningful production state.

This came up while reviewing relay-control migration work, but it does not appear to be caused by the new telemetry or the shared-plane routing changes.

mubarakcoded:
## Summary

- `compute_global_since_timestamp` loaded `Account::all()` independently from its caller, creating a TOCTOU window where a concurrent logout could empty the accounts table between queries — hitting a spurious WARN log repeatedly
- Changed the function from `async fn(&Whitenoise) -> Result<Option<Timestamp>>` to `fn(&[Account]) -> Option<Timestamp>`, accepting the caller's already-loaded accounts
- Moved the empty-account teardown from `setup_global_users_subscriptions` into `sync_discovery_subscriptions` so that all callers (`refresh_global_subscription_for_user`, `refresh_all_global_subscriptions`) also get proper cleanup — previously, after logout, watched users could keep discovery subscriptions alive with no accounts to consume them
- Added unit tests covering all branches of `compute_global_since_timestamp`

## Test plan

- [x] `just precommit-quick` passes (fmt, docs, clippy, unit tests)

Closes #590

## Summary by CodeRabbit

* **Refactor**
  * Improved subscription setup and timestamp handling to avoid race conditions and reduce redundant lookups, yielding more consistent subscription behavior.

* **Bug Fixes**
  * After logout, discovery subscriptions are refreshed; failures are non-fatal and logged as warnings to avoid interrupting logout.

* **Tests**
  * Added unit tests for timestamp calculation to strengthen reliability.

### DIFF
--- a/src/whitenoise/mod.rs
+++ b/src/whitenoise/mod.rs
@@ -579,14 +579,20 @@ impl Whitenoise {
         whitenoise_ref.sync_discovery_subscriptions().await
     }
 
+    /// Buffer (in seconds) subtracted from `last_synced_at` timestamps when
+    /// computing subscription `since` values, to account for clock drift and
+    /// relay propagation delay.
+    const SUBSCRIPTION_BUFFER_SECS: u64 = 10;
+
     // Compute a shared since timestamp for global user subscriptions.
     // - Accepts the already-loaded account list to avoid TOCTOU races
     // - If any account is unsynced (last_synced_at = None), return None
-    // - Otherwise, use min(last_synced_at) minus a 10s buffer, floored at 0
+    // - Otherwise, use min(last_synced_at) minus a buffer, floored at 0
+    #[perf_instrument("whitenoise")]
     fn compute_global_since_timestamp(accounts: &[Account]) -> Option<nostr_sdk::Timestamp> {
         if accounts.is_empty() {
             tracing::debug!(
-                target: "whitenoise::setup_global_users_subscriptions",
+                target: "whitenoise::compute_global_since_timestamp",
                 "No accounts; defaulting to since=None"
             );
             return None;
@@ -598,34 +604,24 @@ impl Whitenoise {
                 .filter(|a| a.last_synced_at.is_none())
                 .count();
             tracing::info!(
-                target: "whitenoise::setup_global_users_subscriptions",
+                target: "whitenoise::compute_global_since_timestamp",
                 "Global subscriptions using since=None due to {} unsynced accounts",
                 unsynced
             );
             return None;
         }
 
-        const BUFFER_SECS: u64 = 10;
         let since = accounts
             .iter()
-            .filter_map(|a| a.since_timestamp(BUFFER_SECS))
+            .filter_map(|a| a.since_timestamp(Self::SUBSCRIPTION_BUFFER_SECS))
             .min_by_key(|t| t.as_secs());
 
-        match since {
-            Some(ts) => {
-                tracing::info!(
-                    target: "whitenoise::setup_global_users_subscriptions",
-                    "Global subscriptions using since={} ({}s buffer)",
-                    ts.as_secs(), BUFFER_SECS
-                );
-            }
-            None => {
-                tracing::debug!(
-                    target: "whitenoise::setup_global_users_subscriptions",
-                    "All accounts synced but since_timestamp returned None for all; \
-                     defaulting to since=None"
-                );
-            }
+        if let Some(ts) = since {
+            tracing::info!(
+                target: "whitenoise::compute_global_since_timestamp",
+                "Global subscriptions using since={} ({}s buffer)",
+                ts.as_secs(), Self::SUBSCRIPTION_BUFFER_SECS
+            );
         }
         since
     }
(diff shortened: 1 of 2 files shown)


src/whitenoise/mod.rs
```diff
@@ -556,19 +544,23 @@ impl Whitenoise {
...
+                    ts.as_secs(), BUFFER_SECS
+                );
+            }
+            None => {
+                tracing::debug!(
+                    target: "whitenoise::setup_global_users_subscriptions",
+                    "All accounts synced but since_timestamp returned None for all; \
+                     defaulting to since=None"
+                );
+            }
```
jgmontoya: Is this branch reachable?

src/whitenoise/mod.rs
```diff
@@ -509,34 +509,22 @@ impl Whitenoise {
...
     // - If any account is unsynced (last_synced_at = None), return None
     // - Otherwise, use min(last_synced_at) minus a 10s buffer, floored at 0
-    async fn compute_global_since_timestamp(
-        whitenoise_ref: &Whitenoise,
-    ) -> Result<Option<nostr_sdk::Timestamp>> {
-        let accounts = Account::all(&whitenoise_ref.database).await?;
+    fn compute_global_since_timestamp(accounts: &[Account]) -> Option<nostr_sdk::Timestamp> {
+        if accounts.is_empty() {
+            tracing::debug!(
+                target: "whitenoise::setup_global_users_subscriptions",
```
jgmontoya: is this the right target?

src/whitenoise/mod.rs
```diff
@@ -1212,13 +1204,23 @@ impl Whitenoise {
...
         let watched_users = User::all_pubkeys(&self.database).await?;
-        let follow_list_accounts = Account::all(&self.database)
-            .await?
-            .into_iter()
+        let follow_list_accounts = accounts
+            .iter()
             .map(|account| (account.pubkey, account.since_timestamp(10)))
             .collect::<Vec<_>>();
-        let public_since = Self::compute_global_since_timestamp(self).await?;
+        let public_since = Self::compute_global_since_timestamp(&accounts);
```
jgmontoya, line 1313: should the '10' here use the `BUFFER_SECS` constant?

src/whitenoise/mod.rs
```diff
@@ -2631,6 +2636,85 @@ mod tests {
...
+                    user_id: 2,
+                    account_type: AccountType::Local,
+                    last_synced_at: None,
+                    created_at: now,
+                    updated_at: now,
+                },
+            ];
+            // One unsynced account forces since=None
+            assert!(Whitenoise::compute_global_since_timestamp(&accounts).is_none());
+        }
```
jgmontoya, line 2777: Maybe create a helper to reduce the 'noise' on creating the accounts, something like:

```rust
fn make_account(user_id: i64, last_synced_at: Option<DateTime<Utc>>) -> Account {
    Account {
        id: None,
        pubkey: Keys::generate().public_key(),
        user_id,
        account_type: AccountType::Local,
        last_synced_at,
        created_at: Utc::now(),
        updated_at: Utc::now(),
    }
}
```

Then the tests become easier to read:

```rust
#[test]
fn test_compute_global_since_mixed_synced_unsynced_returns_none() {
    let now = Utc::now();
    let accounts = vec![
        make_account(1, Some(now - TimeDelta::seconds(100))),
        make_account(2, None),
    ];
    assert!(Whitenoise::compute_global_since_timestamp(&accounts).is_none());
}
```

PR #597, opened 2026-03-09
Fix relay-control group subscriptions and relay sync

erskingardner:
## Summary
- pin CLI-created groups to `nos.lol`, `relay.primal.net`, and `relay.damus.io` while keeping core group creation and updates relay-driven
- keep group-plane subscriptions scoped to each distinct relay set, remove the temporary randomized sharding, and refresh group subscriptions in the background after relay-affecting MLS changes
- reconcile stored account relay lists instead of only appending, and downgrade untracked ephemeral fetch subscription events from error to debug

## Testing
- `just fmt`
- `just precommit-quick` (`fmt`, `docs`, and `clippy` passed; `tests` stalled in existing `whitenoise::accounts::login` nextest cases and was stopped)`

## Summary by CodeRabbit

* **Refactor**
  * Subscription model reworked to manage multiple per-group subscriptions with isolated relay sets.
  * Centralized relay synchronization replaces scattered per-relay updates to avoid stale entries.

* **New Features**
  * Background refresh of account group subscriptions after group changes and message commits.
  * CLI uses a default set of relays for group creation; login/setup flows now sync discovered relays centrally.

* **Bug Fixes**
  * Improved logging to distinguish tracked vs untracked subscriptions and reduce noisy errors.

* **Tests**
  * Updated/added tests validating per-group relay behavior and syncing.

### DIFF
--- a/src/cli/dispatch.rs
+++ b/src/cli/dispatch.rs
@@ -7,6 +7,7 @@ use tokio::io::AsyncWriteExt;
 use crate::Whitenoise;
 use crate::whitenoise::accounts_groups::AccountGroup;
 use crate::whitenoise::app_settings::{Language, ThemeMode};
+use crate::whitenoise::relays::Relay;
 use crate::whitenoise::users::UserSyncMode;
 
 use super::protocol::{Request, Response};
@@ -752,17 +753,11 @@ async fn resolve_display_name(wn: &Whitenoise, pubkey: &PublicKey) -> Option<Str
         .cloned()
 }
 
-fn cli_group_relay_urls() -> Result<Vec<RelayUrl>, Response> {
-    [
-        "wss://nos.lol",
-        "wss://relay.primal.net",
-        "wss://relay.damus.io",
-    ]
-    .into_iter()
-    .map(|url| {
-        RelayUrl::parse(url).map_err(|e| Response::err(format!("invalid group relay URL: {e}")))
-    })
-    .collect()
+fn cli_group_relay_urls() -> Vec<RelayUrl> {
+    Relay::defaults()
+        .into_iter()
+        .map(|relay| relay.url)
+        .collect()
 }
 
 async fn create_group(
@@ -785,7 +780,7 @@ async fn create_group(
         None, // image_hash
         None, // image_key
         None, // image_nonce
-        cli_group_relay_urls()?,
+        cli_group_relay_urls(),
         vec![account.pubkey], // admins — creator only
     );
 
--- a/src/relay_control/groups.rs
+++ b/src/relay_control/groups.rs
@@ -113,6 +113,7 @@ impl GroupPlane {
         // (The unsubscribe above already tore down the previous subscription,
         // so leaving stale relay info in the map would make the health check
         // falsely report healthy with no live subscription.)
+        let mut installed_subscription_indices = Vec::new();
         for subscription in &subscriptions {
             if subscription.relays.is_empty() || subscription.group_ids.is_empty() {
                 continue;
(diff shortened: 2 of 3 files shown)


src/cli/dispatch.rs
```diff
@@ -752,6 +752,19 @@ async fn resolve_display_name(wn: &Whitenoise, pubkey: &PublicKey) -> Option<Str
         .cloned()
 }
 
+fn cli_group_relay_urls() -> Result<Vec<RelayUrl>, Response> {
+    [
+        "wss://nos.lol",
+        "wss://relay.primal.net",
+        "wss://relay.damus.io",
+    ]
```
jgmontoya: We should use the `default` method on `Relay`

src/relay_control/groups.rs
```diff
@@ -68,70 +81,82 @@ impl GroupPlane {
...
+                    filter,
+                    SubscriptionStream::GroupMessages,
+                    Some(pubkey),
+                    &subscription.group_ids,
+                )
+                .await
+            {
+                self.accounts.write().await.remove(&pubkey);
+                return Err(e);
+            }
```
jgmontoya, line 157: My agent had this to say about this:

> The loop subscribes relay-sets one at a time. If the first succeeds and the second fails, the account is removed from the map (`remove(&pubkey)`), but the first subscription is still active in the relay session. The `unsubscribe` block at the top of the function already ran and tore down the _old_ subscriptions; it will not clean up the partially-created new ones.
> 
> The comment at lines 111–115 shows this is intentional: removing the account entry triggers the health check recovery path (`has_active_subscription` → `false` → `ensure_account_subscriptions` → `refresh_account_subscriptions`). However, the recovery has a gap: `refresh_account_subscriptions` calls `deactivate_account_subscriptions`, which calls `group_plane.remove_account(&pubkey)`, but the pubkey was already removed from the map during the failure. So `remove_account` finds nothing and the orphaned subscriptions persist.
> 
> The orphaned subscriptions self-heal only if the new subscription indices (from `BTreeMap::enumerate`) happen to match the old ones, because the Nostr relay replaces subscriptions with the same ID. If the group list changed between failure and recovery (e.g., a group was left), the indices shift and the old ones leak permanently.
> 
> In practice this is low-severity: it only fires on the relay-connection error path, the leak is bounded per subscription, and relays eventually clean up idle subscriptions. But since the comment claims recovery handles this and it actually doesn't for the general case, either the cleanup should be explicit (collect successful IDs, unsubscribe on failure) or the comment should be updated to note the gap.

src/whitenoise/accounts/setup.rs
```diff
@@ -604,25 +625,122 @@ impl Whitenoise {
...
+        tokio::spawn(async move {
+            if let Err(error) = refresh_task.await {
+                tracing::error!(
+                    target: "whitenoise::accounts::background_refresh_account_group_subscriptions",
+                    account_pubkey = %account_pubkey,
+                    "Background group subscription refresh task panicked: {}",
+                    error
+                );
+            }
+        });
```
jgmontoya, line 743: From my agent:

> The outer watcher task exists solely to log panics from the inner task. Without it, the dropped `JoinHandle` would silently swallow panics — so the pattern does serve a purpose. However, 5 of 6 other background spawners in the codebase (`background_publish_account_metadata`, `background_publish_account_relay_list`, `background_publish_account_follow_list`, `background_sync_group_image_cache_if_needed`, `background_fetch_user_data`) use a single spawn without panic logging, making this an outlier. If panic logging is valuable here, it should arguably be applied consistently across all background spawners; if it isn't worth adding everywhere else, it isn't worth the extra task here either. Non-blocking — this is a consistency concern, not a crash risk.

erskingardner: Keeping this as-is for this PR. The watcher here was added to avoid silently losing panics on a detached task in a path we were already touching, so I still think it has value. I agree the project-wide background-task panic handling is inconsistent, but that is broader than this fix set and should be handled as a separate cleanup rather than partially standardized here.

PR #600, opened 2026-03-10
Refactor relay control around warm ephemeral sessions

erskingardner:
## Summary
- add an `EphemeralExecutor` with caller-owned scopes, warm relay reuse, and TTL eviction for ad-hoc relays
- move discovery bootstrap toward batched catch-up and watch the full local user set instead of issuing per-user ephemeral sync churn
- keep relay-list-driven refresh work off the hot event-processing path and make subscription-processing integration tests fail fast with localhost relays

## Testing
- `just precommit`

## Summary by CodeRabbit

* **New Features**
  * Scoped ephemeral relay management with warm/unwarm, per-account scopes, TTL-based ad-hoc eviction, and ephemeral state included in snapshots.
  * Batched discovery-based contact-list catch-up and background subscription refresh.

* **Bug Fixes**
  * More robust relay connectivity handling and serialized publish operations to reduce publish conflicts and partial failures.

* **Performance**
  * Telemetry/observability refocused to reduce high-volume legacy paths and optimize persisted telemetry.

* **Tests**
  * Integration tests hardened with retry-based verification; new tests for session reuse, TTL eviction, and scoped behavior.

### DIFF
--- a/src/relay_control/sessions/session.rs
+++ b/src/relay_control/sessions/session.rs
@@ -29,6 +29,20 @@ struct RelaySessionState {
     subscription_relays: RwLock<HashMap<SubscriptionId, Vec<RelayUrl>>>,
 }
 
+#[derive(Debug)]
+struct PreparedRelayUrls {
+    usable_relay_urls: Vec<RelayUrl>,
+    failed_relays: Vec<(RelayUrl, String)>,
+}
+
+#[derive(Debug)]
+struct RelaySetupOutcome {
+    usable_relay_urls: Vec<RelayUrl>,
+    failed_relays: Vec<(RelayUrl, String)>,
+    added_new_relay: bool,
+    last_error: Option<NostrManagerError>,
+}
+
 /// Reusable single-client relay session used by relay planes.
 #[derive(Debug, Clone)]
 pub(crate) struct RelaySession {
@@ -105,8 +119,16 @@ impl RelaySession {
     }
 
     pub(crate) async fn ensure_relays_connected(&self, relay_urls: &[RelayUrl]) -> Result<()> {
+        self.prepare_relay_urls(relay_urls).await?;
+        Ok(())
+    }
+
+    async fn prepare_relay_urls(&self, relay_urls: &[RelayUrl]) -> Result<PreparedRelayUrls> {
         if relay_urls.is_empty() {
-            return Ok(());
+            return Ok(PreparedRelayUrls {
+                usable_relay_urls: Vec::new(),
+                failed_relays: Vec::new(),
+            });
         }
 
         let relay_futures = relay_urls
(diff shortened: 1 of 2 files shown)


src/relay_control/ephemeral_executor.rs
```diff
@@ -0,0 +1,336 @@
...
+        &self,
+        account_pubkey: Option<PublicKey>,
+        relays: &[RelayUrl],
+        event: &Event,
+    ) -> Result<Output<EventId>> {
+        let entry = self.session_entry(account_pubkey).await;
+        entry
+            .prepare_ad_hoc_relays(relays, self.config.ad_hoc_relay_ttl)
+            .await?;
+        entry.session.publish_event_to(relays, event).await
```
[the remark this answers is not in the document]
erskingardner: We don't want any auth to ever happen on this relay plane. The inbox plane is the only place where auth makes sense for fetching events.

src/whitenoise/accounts/login.rs
```diff
@@ -1127,6 +1149,19 @@ impl Whitenoise {
...
+                .unwarm_ephemeral_relays(&ephemeral_warm_relays)
+                .await
+        {
+            tracing::warn!(
+                target: "whitenoise::accounts",
+                account_pubkey = %pubkey,
+                "Failed to unwarm anonymous ephemeral relays during logout: {error}"
+            );
+        }
+
```
jgmontoya, line 1164: From my agent:

> `unwarm_ephemeral_relays` calls `EphemeralExecutor::unwarm_relays` which calls `unpin_relays(None, ...)` — it only touches the anonymous session's pinned set. The account-scoped session is dropped separately by `remove_account_scope`, which shuts it down without going through `unpin_relays`. That is correct for the account-scoped session itself, but the anonymous session still has pin-counts from when `warm_ephemeral_account_relays` called both `warm_relays` (anonymous) and `warm_relays_for_account` (account-scoped). So after logout the anonymous session correctly decrements its pin-counts via `unwarm_ephemeral_relays`. This appears intentional.
> 
> However, there is a race: `deactivate_account_subscriptions` is called between the relay URL collection and the `unwarm_ephemeral_relays` call. If relay lists mutate between those two points (e.g. a relay-list subscription fires), the pin-counts on the anonymous session will be decremented for the *old* relay set, not the current one, potentially leaving stale pins alive. Consider collecting and unwarming while still holding the relay data reference — or, simpler, call `remove_account_scope` first and then `unwarm_ephemeral_relays`, so everything about this account is torn down atomically before other events can fire.

src/relay_control/sessions/session.rs
```diff
@@ -143,11 +146,79 @@ impl RelaySession {
             );
         }
 
-        if added_new_relay || !self.has_any_relay_connected(relay_urls).await {
-            self.client.connect().await;
-            self.client
-                .wait_for_connection(self.config.connect_timeout)
-                .await;
+        let mut relays_to_wait = Vec::new();
+        for relay_url in relay_urls {
```
jgmontoya: should we skip the failed relays urls here, no?

erskingardner: Fixed in f0a4474. The session now carries a filtered usable relay set forward after setup, so relays that fail `ensure_relay_in_client` are skipped in the wait/check path and are not passed back into `fetch_events_from` or `send_event_to`. I also added a unit test around the relay-setup partitioning logic.

src/whitenoise/accounts/login.rs
erskingardner: `remove_account_scope` is already happening before the anonymous unwarm here: `logout()` calls `deactivate_account_subscriptions()` first, and that method tears down the account-scoped ephemeral session before `unwarm_ephemeral_relays()` runs. The stale-pin issue you called out is still real, but it is the broader warm-set reconciliation gap in `EphemeralExecutor`, not this logout ordering specifically. I updated `relay-control-plane-rearchitecture.md` to call that remaining work out explicitly.

PR #601, opened 2026-03-10
feat: message content search with Unicode-aware forward-order matching

dannym-arx:
marmot

Adds `search_messages_in_group` — a forward-order substring search that works across all scripts (Latin, CJK, Arabic, Devanagari) without language detection, using Unicode combining-mark-aware tokenization to keep grapheme clusters intact. The marmot digs through the entire message archive and brings back exactly what you were looking for.

## Summary by CodeRabbit

* **New Features**
  * Added group message search via CLI and server API with Unicode-aware, case-insensitive, tokenized substring queries and an optional result limit (defaults applied).
* **Chores / Migrations**
  * DB schema updated to store normalized message content to support reliable multilingual search.
* **Tests**
  * Extensive tests covering multilingual scripts, combining marks, substring/forward-order matching, case-insensitivity, empty queries, and limit behavior.

### DIFF
--- a/src/cli/dispatch.rs
+++ b/src/cli/dispatch.rs
@@ -1599,7 +1599,8 @@ async fn search_messages(
         .iter()
         .filter_map(|r| {
             let mut msg = format_chat_message(&r.message, &display_names)?;
-            msg["highlight_spans"] = serde_json::to_value(&r.highlight_spans).unwrap_or_default();
+            msg["highlight_spans"] = serde_json::to_value(&r.highlight_spans)
+                .unwrap_or(serde_json::Value::Array(vec![]));
             Some(msg)
         })
         .collect();
--- a/src/whitenoise/database/content_search.rs
+++ b/src/whitenoise/database/content_search.rs
@@ -138,8 +138,8 @@ pub fn query_tokens(query: &str) -> Vec<String> {
 /// Returns a `Vec` of `[start, end]` pairs (char indices, half-open: `content[start..end]`
 /// is the matched substring) in the order the tokens were found.
 ///
-/// If any token is not found (content does not match the query), returns an
-/// empty `Vec` rather than a partial result.
+/// If any token is not found, or if the normalization mapping cannot be built
+/// reliably (e.g. `to_lowercase` expands a codepoint), returns an empty `Vec`.
 pub fn find_highlight_spans(content: &str, query: &str) -> Vec<[usize; 2]> {
     let tokens = query_tokens(query);
     if tokens.is_empty() {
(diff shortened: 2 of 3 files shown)


src/whitenoise/messages.rs
```diff
@@ -670,6 +670,25 @@ impl Whitenoise {
...
+        query: &str,
+        limit: Option<u32>,
+    ) -> Result<Vec<ChatMessage>> {
+        Account::find_by_pubkey(pubkey, &self.database).await?;
+
+        let limit_val = limit.unwrap_or(50);
+        AggregatedMessage::search_messages_in_group(group_id, query, limit_val, &self.database)
+            .await
+            .map_err(|e| WhitenoiseError::from(anyhow::anyhow!("Failed to search messages: {}", e)))
+    }
```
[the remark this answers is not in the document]
dannym-arx: this is a hard choice, because we could register a SQLite function for it... but it would be impractical to keep up...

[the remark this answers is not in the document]
dannym-arx: that's a good solution, yes, that's what I'm gonna do

dannym-arx:
@erskingardner rebased!

src/whitenoise/database/content_search.rs
```diff
@@ -0,0 +1,349 @@
...
+/// Matches each query token in forward order against the normalized form of
+/// `content`, then maps the match positions back to **char indices** into the
+/// original `content` string.
+///
+/// Returns a `Vec` of `[start, end]` pairs (char indices, half-open: `content[start..end]`
+/// is the matched substring) in the order the tokens were found.
+///
+/// If any token is not found (content does not match the query), returns an
+/// empty `Vec` rather than a partial result.
+pub fn find_highlight_spans(content: &str, query: &str) -> Vec<[usize; 2]> {
```
erskingardner, line 143: From the agent: 

`find_highlight_spans` assumes the original `content` and `normalize_for_search(content)` have a
1:1 character-position relationship, then maps byte offsets in the normalized string back to char
indices in the original string using that assumption. That is not safe. `to_lowercase()` can
expand one scalar into multiple scalars (the canonical example is Turkish `İ` -> `i` + combining
dot), so the normalized string can have more character positions than the original. When that
happens, the current `.position(...)` lookup returns the wrong indices and the fallback
`unwrap_or(0)` / `unwrap_or(content_chars.len())` silently produces incorrect highlight ranges.

This matters because the PR explicitly advertises multilingual Unicode-aware search and returns
these spans as part of the public result contract.

**Fix:** build an explicit mapping from normalized byte offsets back to original char indices, or
fail closed (empty spans) when normalization changes positional cardinality in a way the current
algorithm cannot represent correctly.

src/cli/dispatch.rs
```diff
@@ -1567,6 +1577,36 @@ async fn list_messages(
...
+
+    let messages: Vec<crate::whitenoise::message_aggregator::ChatMessage> =
+        results.iter().map(|r| r.message.clone()).collect();
+    let display_names = resolve_chat_display_names(wn, &messages).await;
+
+    let clean: Vec<serde_json::Value> = results
+        .iter()
+        .filter_map(|r| {
+            let mut msg = format_chat_message(&r.message, &display_names)?;
+            msg["highlight_spans"] = serde_json::to_value(&r.highlight_spans).unwrap_or_default();
```
erskingardner: > `msg["highlight_spans"] = serde_json::to_value(&r.highlight_spans).unwrap_or_default();`

`Vec<[usize; 2]>` should serialize reliably, so this is unlikely to fire. But if it ever does,
`unwrap_or_default()` yields JSON `null`, not `[]`, which changes the response shape silently.
That makes downstream failures harder to diagnose. Propagating the error or using an explicit
empty-array fallback would be more consistent.

src/whitenoise/messages.rs
```diff
@@ -670,6 +672,27 @@ impl Whitenoise {
...
+        group_id: &GroupId,
+        query: &str,
+        limit: Option<u32>,
+    ) -> Result<Vec<SearchResult>> {
+        Account::find_by_pubkey(pubkey, &self.database).await?;
+
+        let limit_val = limit.unwrap_or(50);
+        AggregatedMessage::search_messages_in_group(group_id, query, limit_val, &self.database)
+            .await
+            .map_err(|e| WhitenoiseError::from(anyhow::anyhow!("Failed to search messages: {}", e)))
```
erskingardner: The new search path wraps the database error into a generic `anyhow` error before converting it to
`WhitenoiseError`, which throws away the structured error variant. This file already has adjacent
legacy patterns like this, but new code is a good place not to perpetuate it. If `DatabaseError`
already converts into `WhitenoiseError`, using `?` preserves more information.

src/whitenoise/database/content_search.rs
dannym-arx: Good catch, how did I miss that? Thanks

src/cli/dispatch.rs
dannym-arx: I mean that's not gonna happen, but sure, I'll fix it

PR #615, opened 2026-03-10
Add NIP-46 remote signing support

dannym-arx:
marmot

## Summary

NIP-55 (Amber) gives Android users external signer support, but iOS and desktop have nothing equivalent. This adds NIP-46 (Nostr Connect) as a platform-independent alternative: the remote signer talks over Nostr relays instead of Android intents, so it works everywhere.

`NostrConnect` already implements `NostrSigner`, so it drops straight into the existing signing pipeline. Nothing changes about how events get signed, how giftwraps get decrypted, or how the relay control plane works.

## What changed

- **`Cargo.toml`**: added `nostr-connect v0.44` dependency.
- **`src/whitenoise/secrets_store.rs`**: three new methods to store, load, and delete NIP-46 credentials (app secret key + bunker URI) in the platform keychain. The app keypair needs to survive restarts so the remote signer still recognizes us.
- **`src/whitenoise/accounts/login.rs`**: eight new public methods covering both connection flows (bunker-initiated and client-initiated) plus single-step and multi-step login variants. Thin wrappers over the existing external signer machinery.
- **`src/whitenoise/mod.rs`**: auto-reconnect at startup. Walks external accounts, checks for saved NIP-46 credentials, rebuilds the signer if found. NIP-55 accounts get skipped quietly.
- **`src/whitenoise/error.rs`**: two new error variants, `Nip46InvalidUri` and `Nip46Connection`.
- **`src/lib.rs`**: re-exports `NostrConnect` and `NostrConnectURI` so consumers don't need a direct `nostr-connect` dependency.

## New public API

| Method | What it does |
|--------|-------------|
| `create_nip46_signer` | Parses a bunker URI, generates app keys, builds the signer, saves credentials |
| `create_nip46_client_uri` | Generates a `nostrconnect://` URI for QR display (client-initiated flow) |
| `complete_nip46_client_connect` | Finishes the client-initiated flow after the signer responds |
| `login_nip46` | One-shot bunker login |
| `login_nip46_start` | Multi-step login, step 1 (relay discovery) |
| `login_nip46_publish_default_relays` | Multi-step login, step 2a |
| `login_nip46_with_custom_relay` | Multi-step login, step 2b |

## What didn't change

- `AccountType` stays `Local` / `External`. NIP-46 accounts reuse `External`, same as NIP-55.
- No database migration needed.
- `get_signer_for_account()`, `register_external_signer()`, and the entire NIP-55 flow are untouched.
- All existing tests pass.

## Summary by CodeRabbit

* **New Features**
  * Remote-signer authentication via NIP-46 (Nostr Connect) with one-shot and multi-step login flows, background per-account reconnection at startup, and public exposure of Nostr Connect types
  * Secure storage, retrieval, and idempotent cleanup of NIP-46 session credentials

* **Bug Fixes**
  * Improved error handling and messages for invalid remote-signer URIs, connection failures, and pubkey mismatches

* **Tests**
  * Unit tests for NIP-46 flows, credential store operations, and related error cases

### DIFF
--- a/src/whitenoise/accounts/login.rs
+++ b/src/whitenoise/accounts/login.rs
@@ -6696,7 +6696,12 @@ mod tests {
             .unwrap();
 
         // Verify credentials exist
-        assert!(whitenoise.secrets_store.get_nip46_credentials(&pubkey).is_ok());
+        assert!(
+            whitenoise
+                .secrets_store
+                .get_nip46_credentials(&pubkey)
+                .is_ok()
+        );
 
         // Logout should clean up credentials
         whitenoise.logout(&account.pubkey).await.unwrap();
@@ -7007,7 +7012,10 @@ mod tests {
             msg.contains(&pk1.to_hex()),
             "Error should contain expected pubkey"
         );
-        assert!(msg.contains(&pk2.to_hex()), "Error should contain got pubkey");
+        assert!(
+            msg.contains(&pk2.to_hex()),
+            "Error should contain got pubkey"
+        );
         assert!(msg.contains("mismatch"), "Error should mention mismatch");
     }
 }
--- a/src/integration_tests/test_cases/login_flow/nip46_credential_lifecycle.rs
+++ b/src/integration_tests/test_cases/login_flow/nip46_credential_lifecycle.rs
@@ -4,18 +4,19 @@ use nostr_sdk::prelude::*;
 use crate::WhitenoiseError;
 use crate::integration_tests::core::*;
 
-/// Verifies the NIP-46 credential lifecycle through an external signer account:
+/// Verifies the NIP-46 external signer account lifecycle through public APIs:
 ///
-/// 1. Login with external signer and store NIP-46 credentials in the keychain.
-/// 2. Verify credentials are retrievable from the keychain.
-/// 3. Logout and verify credentials are cleaned up.
-/// 4. Verify the reconnect-failed set is managed correctly.
+/// 1. Login with external signer creates an external account.
+/// 2. External account has a signer registered and is retrievable.
+/// 3. Reconnect-failed list is initially empty for healthy accounts.
+/// 4. Re-registering a signer keeps the account healthy.
+/// 5. Logout removes the account.
 pub struct Nip46CredentialLifecycleTestCase;
 
 #[async_trait]
 impl TestCase for Nip46CredentialLifecycleTestCase {
     async fn run(&self, context: &mut ScenarioContext) -> Result<(), WhitenoiseError> {
-        tracing::info!("Testing NIP-46 credential lifecycle (store → retrieve → logout cleanup)");
+        tracing::info!("Testing NIP-46 external signer account lifecycle via public APIs");
 
         let keys = Keys::generate();
         let pubkey = keys.public_key();
@@ -30,64 +31,40 @@ impl TestCase for Nip46CredentialLifecycleTestCase {
             "Expected account to use external signer"
         );
 
-        // Step 2: Store NIP-46 credentials (simulating what login_nip46 would do)
-        let app_keys = Keys::generate();
-        let bunker_uri = "bunker://0000000000000000000000000000000000000000000000000000000000000000?relay=wss://relay.nsec.app";
-        context
-            .whitenoise
-            .secrets_store
-            .store_nip46_credentials(&pubkey, app_keys.secret_key(), bunker_uri)?;
+        // Step 2: Verify the account is findable and has signer registered
+        let found = context.whitenoise.find_account_by_pubkey(&pubkey).await?;
+        assert_eq!(found.pubkey, pubkey);
+        assert!(found.uses_external_signer());
 
-        // Verify credentials exist in keychain
-        let (retrieved_secret, retrieved_uri) = context
-            .whitenoise
-            .secrets_store
-            .get_nip46_credentials(&pubkey)?;
-        assert_eq!(retrieved_uri, bunker_uri);
-        assert_eq!(
-            retrieved_secret.to_secret_hex(),
-            app_keys.secret_key().to_secret_hex()
-        );
-
-        // Step 3: Simulate reconnect failure (e.g. bunker offline at startup)
-        context
-            .whitenoise
-            .nip46_reconnect_failed
-            .insert(pubkey);
+        // Step 3: Reconnect-failed list should not contain this healthy account
+        let failed = context.whitenoise.nip46_reconnect_failed_accounts();
         assert!(
-            !context
-                .whitenoise
-                .nip46_reconnect_failed_accounts()
-                .is_empty(),
-            "Should have a reconnect-failed account"
+            !failed.contains(&pubkey),
+            "Healthy account should not be in reconnect-failed list"
         );
 
-        // Step 4: Re-register signer clears reconnect failure
+        // Step 4: Re-registering the signer should succeed and keep account healthy
         context
             .whitenoise
             .register_external_signer(pubkey, keys.clone())
             .await?;
+        let still_failed = context.whitenoise.nip46_reconnect_failed_accounts();
         assert!(
-            !context
-                .whitenoise
-                .nip46_reconnect_failed_accounts()
-                .contains(&pubkey),
-            "Re-registration should clear the reconnect-failed flag"
+            !still_failed.contains(&pubkey),
+            "Re-registered account should not be in reconnect-failed list"
         );
 
-        // Step 5: Logout should clean up NIP-46 credentials
+        // Step 5: Logout should remove the account
         context.whitenoise.logout(&pubkey).await?;
-
-        let cred_result = context
-            .whitenoise
-            .secrets_store
-            .get_nip46_credentials(&pubkey);
+        let find_result = context.whitenoise.find_account_by_pubkey(&pubkey).await;
         assert!(
-            cred_result.is_err(),
-            "NIP-46 credentials should be cleaned up after logout"
+            find_result.is_err(),
+            "Account should not be findable after logout"
         );
 
-        tracing::info!("✓ NIP-46 credential lifecycle verified: store → retrieve → logout cleanup");
+        tracing::info!(
+            "✓ NIP-46 external signer lifecycle verified: login → signer → re-register → logout"
+        );
         Ok(())
     }
 }
--- a/src/whitenoise/app_settings.rs
+++ b/src/whitenoise/app_settings.rs
@@ -5,17 +5,12 @@ use serde::{Deserialize, Serialize};
 
 use crate::{Whitenoise, whitenoise::Result};
 
-#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Hash)]
+#[derive(Debug, Clone, Default, PartialEq, Eq, Serialize, Deserialize, Hash)]
 pub enum ThemeMode {
     Light,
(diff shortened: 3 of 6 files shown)


src/whitenoise/accounts/login.rs
```diff
@@ -756,6 +770,259 @@ impl Whitenoise {
...
+
+        let signer = NostrConnect::new(uri, app_keys, timeout, None)
+            .map_err(WhitenoiseError::Nip46Connection)?;
+
+        // Seed the cached pubkey so the signer knows which identity it
+        // represents, then verify via a get_public_key() round-trip that the
+        // remote bunker still controls this key. This catches stale or
+        // corrupted keychain entries before they can register a wrong identity.
+        signer
+            .non_secure_set_user_public_key(account.pubkey)
```
erskingardner: The comment says "verify via a get_public_key() round-trip that the remote bunker still controls this key." But `non_secure_set_user_public_key` seeds the *local* pubkey cache, and `NostrConnect::get_public_key()` returns that cached value without doing a network round-trip when the cache is populated. The verification check will always pass because you seeded exactly the value you then compare against. The stated purpose — catching stale or corrupted keychain entries — is not achieved.

If the intent is to avoid a full NIP-46 handshake at startup (reasonable for latency), remove the misleading check and the misleading comment. If the intent is genuine verification, do not call `non_secure_set_user_public_key` and let `get_public_key()` do the full round-trip (accepting the latency cost, gated by the 30-second timeout).

src/whitenoise/accounts/login.rs
```diff
@@ -756,6 +770,259 @@ impl Whitenoise {
...
+        let result = self.login_external_signer_start(pubkey, signer).await?;
+
+        // Persist credentials after the start step succeeds (regardless of
+        // whether relay lists were found). If the keychain write fails, cancel
+        // the pending login to avoid an orphaned pending state. If login
+        // completes successfully later, login_cancel would have cleaned these up
+        // anyway.
+        if let Err(store_err) =
+            self.secrets_store
+                .store_nip46_credentials(&pubkey, app_keys.secret_key(), bunker_uri)
```
erskingardner: The multi-step login stores NIP-46 credentials after `login_nip46_start`, even when the result is `LoginStatus::NeedsRelayLists` (i.e., login is NOT complete yet). `login_cancel` does clean these up, but only if it is called by the caller after a timeout or user abort — there is no guarantee of that. If the user abandons the flow without an explicit cancel, you have an orphaned keychain entry referencing a bunker URI that may no longer be valid.

The one-shot `login_nip46` and `complete_nip46_client_connect` correctly store credentials only after full login success. The multi-step path should follow the same pattern: store credentials only in `login_nip46_publish_default_relays` / `login_nip46_with_custom_relay` after `LoginStatus::Complete` is confirmed.

src/whitenoise/mod.rs
```diff
@@ -543,6 +543,11 @@ impl Whitenoise {
 
         init_timing::record("background_tasks");
 
+        // Bring back NIP-46 signers for any external accounts that have saved
+        // credentials. Needs to happen before subscriptions go up — otherwise
+        // relays that require NIP-42 AUTH won't have a signer to talk to.
+        Self::reconnect_nip46_signers(whitenoise_ref).await;
```
erskingardner: The reconnection futures run concurrently via `join_all`, but the entire `.await` of `reconnect_nip46_signers` still blocks `initialize_whitenoise` until every NIP-46 handshake completes or times out (up to `NIP46_RECONNECT_TIMEOUT_SECS = 30` per account, potentially serialized if the NIP-46 relay is unreachable). For a user with multiple NIP-46 accounts on a slow connection, startup can stall for minutes before any subscriptions come up.

The comment says it must happen before subscriptions because "relays that require NIP-42 AUTH won't have a signer." That is a real concern, but it applies only to the specific account whose signer is missing, not to all accounts. Consider spawning each reconnect into a background task (`tokio::spawn`) and letting subscriptions for accounts that already have signers proceed immediately. Add a short grace window (e.g., 2–3 seconds) before bringing subscriptions up for NIP-46 accounts whose reconnect is still in flight, or accept that those accounts may transiently fail NIP-42 AUTH and rely on reconnect logic.

src/whitenoise/error.rs
```diff
@@ -213,6 +224,12 @@ impl From<Box<dyn std::error::Error + Send + Sync>> for WhitenoiseError {
     }
 }
 
+impl From<nostr_connect::prelude::SignerError> for WhitenoiseError {
```
erskingardner: `SignerError` is a general signing error from `nostr-sdk`. Wrapping it unconditionally as `Nip46Connection(Response(...))` makes every signing failure from any signer look like a NIP-46 connection-response error. This `From` impl will apply globally any time a `SignerError` is propagated up through `?` in a context that expects `WhitenoiseError`. If a local-key signer fails (e.g., due to a key format error), it would appear as a NIP-46 connection issue.

Use a distinct variant like `WhitenoiseError::SignerError(String)` for generic signer failures, or scope the conversion to NIP-46 call sites only by converting inline rather than via a blanket `From` impl.

src/whitenoise/accounts/login.rs
```diff
@@ -756,6 +770,259 @@ impl Whitenoise {
...
+    /// yourself.
+    ///
+    /// Returns `(user_pubkey, signer, app_keys)`. The pubkey comes from the
+    /// remote signer during the initial NIP-46 handshake. The app_keys are
+    /// needed by the caller to persist credentials after a successful login.
+    pub async fn create_nip46_signer(
+        &self,
+        bunker_uri: &str,
+        timeout_secs: u64,
+    ) -> Result<(PublicKey, NostrConnect, Keys)> {
```
erskingardner, line 906: The method returns `(PublicKey, NostrConnect, Keys)` and documents that the caller must persist credentials themselves. This creates a footgun in the public API: any external consumer who calls `create_nip46_signer`, gets a signer, and calls `login_with_external_signer` directly will end up with a NIP-46 account that can't reconnect after restart (no keychain entry). The higher-level `login_nip46` method exists precisely to avoid this, but `create_nip46_signer` is `pub` and not named to signal its partial nature (e.g., `create_nip46_signer_uncredentialed`).

Either make it `pub(crate)` / `pub(super)` and document `login_nip46` as the public entry point, or rename it to make the incomplete-lifecycle nature obvious.

src/whitenoise/accounts/login.rs
```diff
@@ -756,6 +770,259 @@ impl Whitenoise {
...
+        let account = self.login_with_external_signer(pubkey, signer).await?;
+
+        // Persist credentials only after login succeeds. If the keychain write
+        // fails, roll back the login so the caller doesn't get an activated
+        // account with no stored credentials (which would break reconnects).
+        if let Err(store_err) =
+            self.secrets_store
+                .store_nip46_credentials(&pubkey, app_keys.secret_key(), bunker_uri)
+        {
+            if let Err(rollback_err) = self.logout(&pubkey).await {
```
erskingardner: Both `login_nip46` and `complete_nip46_client_connect` contain an identical post-login credential-store-with-rollback block:

```rust
if let Err(store_err) = self.secrets_store.store_nip46_credentials(...) {
    if let Err(rollback_err) = self.logout(&pubkey).await {
        tracing::warn!(...);
    }
    return Err(store_err.into());
}
```

Extract this into a small `async fn store_nip46_credentials_or_rollback(pubkey, app_keys, bunker_uri) -> Result<()>` helper to reduce duplication and make the invariant explicit.

src/whitenoise/mod.rs
```diff
@@ -670,6 +675,78 @@ impl Whitenoise {
...
+    /// signer from saved credentials (app keys + bunker URI).
+    ///
+    /// Reconnections run concurrently so a slow or stalled remote signer
+    /// cannot delay unrelated accounts or overall startup. Accounts that
+    /// don't have NIP-46 credentials — typically NIP-55/Amber accounts —
+    /// are quietly skipped. If something goes wrong for a specific account
+    /// we log a warning and keep going; one broken reconnect shouldn't take
+    /// down the whole startup.
+    async fn reconnect_nip46_signers(whitenoise_ref: &Whitenoise) {
+        const NIP46_RECONNECT_TIMEOUT_SECS: u64 = 30;
```
erskingardner, line 718: This constant is local to `reconnect_nip46_signers`. Most other timeout-bearing NIP-46 methods take `timeout_secs: u64` as a caller-controlled parameter. The asymmetry is understandable (startup callers have no natural place to pass it), but 30 seconds is conspicuously long for a startup step that already blocks subscriptions. Document why 30 seconds was chosen, or reduce it to something like 10 seconds with a comment.

src/whitenoise/error.rs
```diff
@@ -198,6 +200,15 @@ pub enum WhitenoiseError {
     #[error("Unsupported media format: {0}")]
     UnsupportedMediaFormat(String),
 
+    #[error("NIP-46 invalid bunker URI: {0}")]
+    Nip46InvalidUri(#[from] nip46::Error),
```
erskingardner: Using `#[from]` makes `nip46::Error → WhitenoiseError` a blanket conversion active everywhere in the crate. Given the project's careful scoping of conversions, it may be preferable to call `.map_err(WhitenoiseError::Nip46InvalidUri)` at the two call sites and drop the `#[from]` annotation, keeping the error boundary explicit. Same applies to `Nip46Connection(#[from] NostrConnectError)`.

src/whitenoise/accounts/login.rs
```diff
@@ -5922,4 +6201,78 @@ mod tests {
...
+
+    #[test]
+    fn nip46_create_client_uri_has_correct_scheme() {
+        let whitenoise = tokio::runtime::Runtime::new()
+            .unwrap()
+            .block_on(async { test_get_whitenoise().await });
+
+        let relay_url = RelayUrl::parse("wss://relay.example.com").unwrap();
+        let (app_keys, uri) =
+            whitenoise.create_nip46_client_uri(vec![relay_url.clone()], "Test App");
```
erskingardner, line 6611: `create_nip46_client_uri` is a pure synchronous function (`fn`, not `async fn`) that takes `self` only for API consistency (it doesn't use any fields). The test spins up a full Whitenoise instance just to call it. This function could instead be tested with a standalone call to `NostrConnectURI::client(...)` directly, or `create_nip46_client_uri` could be made a free function / associated function. If it stays as `&self`, the test should be `#[tokio::test]` for consistency with the rest of the test suite rather than `#[test]` with `block_on`.

src/whitenoise/mod.rs
```diff
@@ -670,6 +675,78 @@ impl Whitenoise {
...
+                    tracing::debug!(
+                        target: "whitenoise::initialize_whitenoise",
+                        "No NIP-46 credentials for external account {} (likely NIP-55)",
+                        pubkey.to_hex()
+                    );
+                }
+                Err(e) => {
+                    tracing::warn!(
+                        target: "whitenoise::initialize_whitenoise",
+                        "Failed to reconnect NIP-46 signer for account {}: {}",
```
erskingardner, line 774: When a NIP-46 signer fails to reconnect (e.g., the user's signer app is offline), the account appears in the account list but any signing attempt will fail with `AccountNotAuthorized`. There's no stream event or status flag to let the Flutter layer surface "this account needs reconnection." Consider either streaming a notification or adding a method the UI can poll to detect unsigned NIP-46 accounts.

PR #616, opened 2026-03-11
perf_span! zero-cost performance tracking across the crate

dannym-arx:
marmot

Adds a zero-cost performance tracking system to the crate. One macro call drops a timing marker anywhere in the codebase. It has no external dependencies, works without feature flags in library code, and leaves existing benchmarks untouched.

## How it works

`perf_span!("marker::name")` returns a `PerfGuard` that records `Instant::now()` on creation and emits a `tracing::info!` event (target `"whitenoise::perf"`, fields `name` + `duration_ns`) when dropped. The guard is `Send`, so it holds safely across `.await` points in Tokio tasks.

In production the `whitenoise::perf` filter is off by default. The only runtime cost is a single level-filter comparison per call site.

In benchmark builds, `PerfTracingLayer` is registered at subscriber init. It captures every `whitenoise::perf` event and accumulates `PerfSample` records. After each benchmark loop, the layer is drained and per-marker statistics (mean, median, p95, p99, min, max, call count) are appended to `BenchmarkResult::perf_breakdown` and printed in the summary table.

## What was instrumented

60+ markers across the hot paths:

- **messages**: `send_message_to_group` and all sub-steps (MLS encrypt, event build, token parse, cache write, reaction/deletion apply)
- **groups**: `create_group`, `add_members`, `remove_members`, `update_group_data`, `leave_group`, `visible_groups`, `publish_and_merge_commit`
- **event processor**: per-event routing, `handle_mls_message` (including `mls_process_message`), `handle_giftwrap` (decrypt + process_welcome), `handle_metadata`, `handle_relay_list`, `handle_contact_list`, cache operations
- **message aggregator**: `aggregate_messages_for_group`, `process_messages`, `process_single_message`, `process_regular_message`
- **accounts**: `create_identity`, `login`, `activate_account`, `setup_key_package`, `setup_subscriptions`
- **users**: `find_or_create`, `sync_metadata`, `sync_user_blocking`
- **key packages**: `create_and_publish`, `encode`, `publish_to_relays`
- **media**: `upload_chat_media`, `download_chat_media`, `store_and_record`, `store_parsed_references`

## Adding a new marker

```rust
let _span = perf_span!("my_module::my_operation");
// ... everything in scope is timed until _span drops
```

The marmot colony now has eyes everywhere.

## Summary by CodeRabbit

* **New Features**
  * App-wide performance tracing and a perf-instrument macro; benchmark reports now include detailed perf breakdowns.

* **Improvements**
  * More accurate benchmark timing and trace propagation; warmup skew reduced.
  * In-memory caching to speed event processing and reduce redundant lookups.
  * Minor error-reporting refinements.

* **Chores**
  * Extensive non-functional instrumentation added across the codebase for richer telemetry.

### DIFF
--- a/src/whitenoise/database/accounts.rs
+++ b/src/whitenoise/database/accounts.rs
@@ -232,8 +232,7 @@ impl Account {
             "SELECT u.id, u.pubkey, u.metadata, u.created_at, u.updated_at
              FROM account_follows af
              JOIN users u ON af.user_id = u.id
-             WHERE af.account_id = ?
-             ORDER BY u.created_at ASC, u.id ASC",
+             WHERE af.account_id = ?",
         )
         .bind(self.id)
         .fetch_all(&database.pool)
--- a/src/whitenoise/database/users.rs
+++ b/src/whitenoise/database/users.rs
@@ -203,6 +203,7 @@ impl User {
     /// # Errors
     ///
     /// Returns a [`WhitenoiseError::UserNotFound`] if no user with the given public key exists.
+    #[perf_instrument("db::users")]
     pub(crate) async fn find_by_pubkey(
         pubkey: &PublicKey,
         database: &Database,
@@ -230,6 +231,7 @@ impl User {
     ///
     /// Returns a `Vec<User>` containing all users found.
     /// Users that don't exist in the database are simply not included in the result.
+    #[perf_instrument("db::users")]
     pub(crate) async fn find_by_pubkeys(
         pubkeys: &[PublicKey],
         database: &Database,
@@ -272,6 +274,7 @@ impl User {
     /// # Errors
     ///
     /// Returns a [`WhitenoiseError`] if the database query fails.
+    #[perf_instrument("db::users")]
     pub(crate) async fn relays(
         &self,
         relay_type: RelayType,
@@ -311,6 +314,7 @@ impl User {
     /// Use this to record "we checked this user's metadata" even when no new
     /// metadata was found, so that `needs_metadata_refresh()` respects the TTL
     /// for empty-profile users.
+    #[perf_instrument("db::users")]
     pub(crate) async fn touch_updated_at(
         &mut self,
         database: &Database,
@@ -341,6 +345,7 @@ impl User {
     /// # Errors
     ///
     /// Returns a [`WhitenoiseError`] if the database operation fails.
+    #[perf_instrument("db::users")]
     pub(crate) async fn save(&self, database: &Database) -> Result<User, WhitenoiseError> {
         let mut tx = database.pool.begin().await.map_err(DatabaseError::Sqlx)?;
 
--- a/src/whitenoise/users.rs
+++ b/src/whitenoise/users.rs
@@ -103,39 +103,28 @@ impl User {
             .relay_control
             .fetch_metadata_from(&relays_urls, self.pubkey)
             .await?;
-        let mut saved = false;
+
         if let Some(event) = metadata_event {
-            let metadata = Metadata::from_json(&event.content)?;
-            let should_update = self
-                .should_update_metadata(&event, false, &whitenoise.database)
-                .await?;
+            // Overwrite local metadata with whatever the relay returned.
+            // We don't guard on timestamps or "already processed" here because
+            // this is a deliberate, targeted fetch — not reactive event processing.
+            // The caller (sync_user_blocking / background_fetch_user_data) already
+            // gates on TTL, so we won't over-fetch.
+            self.metadata = Metadata::from_json(&event.content)?;
+            self.save(&whitenoise.database).await?;
 
-            if should_update {
-                self.metadata = metadata;
-
-                // Save the updated user metadata (also bumps updated_at)
-                self.save(&whitenoise.database).await?;
-                saved = true;
-
-                whitenoise
-                    .event_tracker
-                    .track_processed_global_event(&event)
-                    .await?;
-
-                tracing::debug!(
-                    target: "whitenoise::users::sync_metadata",
-                    "Updated metadata for user {} with event timestamp {} via background sync",
-                    self.pubkey,
-                    event.created_at
-                );
-            }
-        }
-        // Always record that we checked, even if no metadata was found or
-        // the event didn't need updating.  This lets needs_metadata_refresh()
-        // respect the TTL for empty-profile users.
-        if !saved {
+            tracing::debug!(
+                target: "whitenoise::users::sync_metadata",
+                "Updated metadata for user {} with event timestamp {} via background sync",
+                self.pubkey,
+                event.created_at
+            );
+        } else {
+            // No metadata found — record that we checked so TTL is respected
+            // for empty-profile users, preventing repeated fruitless fetches.
             self.touch_updated_at(&whitenoise.database).await?;
         }
+
         Ok(())
     }
 


src/integration_tests/benchmarks/registry.rs
```diff
@@ -139,6 +139,39 @@ impl BenchmarkRegistry {
             tracing::info!("    P99:       {:?}", result.p99);
             tracing::info!("");
             tracing::info!("  Throughput:  {:.2} ops/sec", result.throughput);
+
+            if let Some(breakdown) = &result.perf_breakdown
+                && !breakdown.is_empty()
+            {
+                tracing::info!("");
+                tracing::info!("  Perf Breakdown (sorted by call count):");
```
erskingardner: The output says the table is sorted by call count, but `PerfTracingLayer::drain()` actually sorts by `total_duration`:

> `breakdowns.sort_by(|a, b| b.total_duration.cmp(&a.total_duration));`

The current sort order is reasonable, but the label is false and will mislead anyone reading benchmark output.

src/integration_tests/benchmarks/core/benchmark_scenario.rs
```diff
@@ -68,6 +69,17 @@ pub trait BenchmarkScenario {
...
+            // Clear perf samples accumulated during warmup so they don't skew benchmark stats
+            if let Some(layer) = PERF_LAYER.get() {
+                layer.clear();
+            }
+        }
+
+        // Clear perf samples before the timed phase (catches setup spans
+        // when warmup_iterations == 0, since the warmup branch handles its own clear)
+        if let Some(layer) = PERF_LAYER.get() {
+            layer.clear();
```
erskingardner, line 83: This unconditional clear happens immediately after the warmup branch, which already calls `layer.clear()` when `warmup_iterations > 0`. In the warmup case one of those calls is always redundant. It does not break behavior, but it makes the benchmark lifecycle harder to reason about and suggests the phase boundaries were not fully tightened.

crates/whitenoise-macros/src/lib.rs
```diff
@@ -0,0 +1,103 @@
...
+                format!("{}::{}", prefix.value(), fn_name)
+            }
+        }
+    };
+
+    let body = &func.block;
+
+    let new_body = quote! {
+        {
+            let _perf_span = crate::perf_span!(#span_name);
```
erskingardner, line 128: The new `whitenoise-macros` crate is added as a normal workspace crate with no `publish = false`, but it hardcodes expansion to `crate::perf_span!`, which only resolves inside `whitenoise` itself. Any external consumer of `whitenoise-macros` will get a confusing compile failure from generated code. If this macro crate is intentionally internal, mark it non-publishable; otherwise the expansion strategy needs to be made reusable.

crates/whitenoise-macros/src/lib.rs
```diff
@@ -0,0 +1,103 @@
...
+    //   #[perf_instrument("prefix")]
+    //   #[perf_instrument("prefix", name = "explicit::span::name")]
+    let span_name = {
+        let attr2: proc_macro2::TokenStream = attr.into();
+        let mut tokens = attr2.into_iter();
+
+        // First token group: the prefix literal
+        let prefix_tt = tokens
+            .next()
+            .expect("perf_instrument: expected a prefix string argument");
```
erskingardner: The parser uses `expect!`/`assert!` throughout. That turns user mistakes into `proc macro panicked` compiler errors instead of proper, span-anchored diagnostics. Since this macro is now applied pervasively, parse failures should use `syn::Error::new(...).to_compile_error()` so mistakes point to the actual attribute site.

src/cli/dispatch.rs
```diff
@@ -489,6 +491,7 @@ pub async fn dispatch(req: Request) -> Response {
 }
 
 /// Write a single response line to the writer. Returns false if the client disconnected.
+#[perf_instrument("dispatch")]
```
erskingardner: `write_response()` and `write_stream_end()` are now instrumented under the same `dispatch` prefix as actual request handling. Those functions mainly measure pipe/socket backpressure and reader speed, not CLI dispatch computation. That will pollute the breakdown with I/O timing that is dominated by the consumer rather than by Whitenoise logic, and it also double-counts time already covered by the top-level `dispatch()` span.

dannym-arx: I can remove those

src/integration_tests/benchmarks/scenarios/user_search.rs
```diff
@@ -91,6 +91,12 @@ impl BenchmarkScenario for UserSearchBenchmark {
                 WhitenoiseError::Other(anyhow::anyhow!("Failed to follow target: {}", e))
             })?;
 
+        // Clear perf samples accumulated during setup/follow so only
+        // actual search timings appear in the breakdown.
+        if let Some(layer) = crate::integration_tests::benchmarks::PERF_LAYER.get() {
+            layer.clear();
+        }
+
```
jgmontoya, line 99: From my agent:

> **`src/integration_tests/benchmarks/scenarios/user_search.rs`:~99 and `src/integration_tests/benchmarks/core/benchmark_scenario.rs`:~74**
> 
> The generic `BenchmarkScenario::run()` already calls `PERF_LAYER.get().map(|l| l.clear())` just before the timed phase. `UserSearchBenchmark::run()` overrides the entire `run()` method and inserts its own `layer.clear()` call during setup — but then at the end it calls `layer.drain()` *instead of* delegating to the generic runner. There is no double-clear bug here, but the two sites are out of sync in principle: if `BenchmarkScenario::run()` ever changes its pre-benchmark `clear()` placement, `UserSearchBenchmark::run()` won't benefit. Consider if `UserSearchBenchmark` should delegate to the default `run()` implementation rather than reimplementing it.

src/lib.rs
```diff
@@ -83,6 +95,76 @@ static TRACING_GUARDS: OnceLock<Mutex<Option<(WorkerGuard, WorkerGuard)>>> = Onc
...
+                .add_directive("whitenoise::perf=off".parse().unwrap())
+                .add_directive("sqlx::query=off".parse().unwrap()),
+            Err(_) => EnvFilter::new(
+                "info,refinery_core=warn,refinery=warn,whitenoise::perf=off,sqlx::query=off",
+            ),
+        };
+
+        // Add a dedicated INFO-level filter so the perf layer receives events
+        // from both our manual perf_span! markers and sqlx's query logger.
+        let perf_filter = tracing_subscriber::filter::Targets::new()
```
jgmontoya: From my agent:

> 
> ```rust
> let output_filter = match EnvFilter::try_from_default_env() { ... };
> let file_filter   = match EnvFilter::try_from_default_env() { ... };
> ```
> 
> Both filters are constructed from identical logic and identical string literals. `EnvFilter` is not `Clone`, so duplication here is intentional in one sense — but this means the environment variable is parsed twice, and if it ever differs between calls (it won't, but) the two layers could behave differently. More importantly, it is simply confusing to read. If the intent is "both output layers share the same filter", a comment saying so and/or extracting a helper would clarify the design. Alternatively, if stdout and file genuinely *should* be filterable independently, name the variable `stdout_filter` instead of `output_filter`.

src/lib.rs
```diff
@@ -83,6 +95,76 @@ static TRACING_GUARDS: OnceLock<Mutex<Option<(WorkerGuard, WorkerGuard)>>> = Onc
...
+
+        Registry::default()
+            .with(stdout_layer.with_filter(output_filter))
+            .with(file_layer.with_filter(file_filter))
+            .with(perf_layer.with_filter(perf_filter))
+            .init();
+    });
+}
+
+fn init_tracing_base(logs_dir: &std::path::Path) {
```
jgmontoya: From my agent:

> `init_tracing_base` is a private function called only by the no-op `init_tracing` wrapper. `init_tracing_with_perf_layer` duplicates its entire body. This is ~50 lines of copy-paste. If the two ever diverge (e.g. someone adds a new layer to the base path), the benchmark path silently misses it.
> 
> Consider extracting the shared setup into a helper that builds the two `fmt::Layer`s and returns them, then have both callers finish the `Registry::default().with(...).init()` call with their own additions. Or make `init_tracing_with_perf_layer` the canonical path and have the non-benchmark path call it with a no-op layer.

src/lib.rs
jgmontoya, line 1: From my agent:

> **`crates/whitenoise-macros/src/lib.rs`:51**
> 
> ```rust
> let _perf_span = crate::perf_span!(#span_name);
> ```
> 
> The macro is in an external proc-macro crate but expands to code that refers to `crate::perf_span!`, which means it can *only* be used inside the `whitenoise` crate. The `Cargo.toml` marks `publish = false`, but the macro's `pub(crate)` re-export in `lib.rs` and the comment "crate-private because…" already acknowledge this. This is an acceptable constraint, but it should be documented in the proc-macro crate's `Cargo.toml` or `lib.rs` top-level comment so future contributors don't wonder why it fails elsewhere.

src/integration_tests/benchmarks/perf_layer.rs
```diff
@@ -0,0 +1,222 @@
...
+}
+
+impl PerfBreakdown {
+    fn from_samples(marker: String, mut samples: Vec<Duration>) -> Self {
+        let call_count = samples.len() as u64;
+        let total_duration: Duration = samples.iter().sum();
+        let mean = stats::calculate_mean(&samples);
+        let median = stats::calculate_median(&mut samples);
+        let p95 = stats::calculate_percentile(&mut samples.clone(), 0.95);
+        let p99 = stats::calculate_percentile(&mut samples.clone(), 0.99);
```
jgmontoya: Are both of these clones necessary?

crates/whitenoise-macros/src/lib.rs
```diff
@@ -0,0 +1,142 @@
...
+                "perf_instrument: `name` value must be a string literal, e.g. name = \"my_module::MyType::my_fn\"",
+            )
+        })?;
+        explicit_name = Some(value.value());
+    }
+
+    Ok(match explicit_name {
+        Some(name) => name,
+        None => {
+            let fn_name = func.sig.ident.to_string();
```
jgmontoya: From my agent:

> The manual token-tree iteration (peek at `comma.to_string() != ","`, `ident.to_string() != "name"`, etc.) is fragile: it doesn't handle trailing commas, whitespace variation, or grouped token trees correctly. The standard idiomatic approach for proc-macro attribute parsing is `syn::parse::Parser` with a `Punctuated` list, or `syn::meta::parser` (syn 2.x). This would eliminate the string comparisons against `","` and `"="` and produce much better compiler error messages with proper `Span` info. Given this is a developer-facing tool, error quality matters.

src/integration_tests/benchmarks/scenarios/user_search.rs
jgmontoya, line 1: From my agent:

> `UserSearchBenchmark::run()` calls `PERF_LAYER.get().map(|layer| layer.drain())` directly and passes the result to `BenchmarkResult::from_timings`. The generic `BenchmarkScenario::run()` in `benchmark_scenario.rs` does the same at line ~105. Since `UserSearchBenchmark` fully reimplements `run()`, both drain paths exist — but only one fires per run. This is correct, just invisible: a future reader maintaining `benchmark_scenario.rs` may not realise that overriding `run()` also requires duplicating the drain logic. A doc comment on the `run()` trait method explaining that overriders must handle `PERF_LAYER.drain()` themselves would help.

src/lib.rs
dannym-arx: this is not a miss, it's the expected behavior, I don't think it makes any sense to ever make whitenoise-macros publicly available, it's definitely meant to be used only internally

src/whitenoise/users.rs
```diff
@@ -125,6 +130,12 @@ impl User {
                 );
             }
         }
+        // Always record that we checked, even if no metadata was found or
+        // the event didn't need updating.  This lets needs_metadata_refresh()
+        // respect the TTL for empty-profile users.
+        if !saved {
+            self.touch_updated_at(&whitenoise.database).await?;
+        }
```
jgmontoya: Interesting.

This will go to the database to update the `updated_at` field.
But then I feel like we can simplify things and just remove the whole "should_update" check and just directly update the metadata with whatever we just fetched. What do you think?

src/whitenoise/database/accounts.rs
```diff
@@ -217,12 +226,14 @@ impl Account {
...
     /// Returns a [`WhitenoiseError::AccountNotFound`] if the database query fails.
+    #[perf_instrument("db::accounts")]
     pub(crate) async fn follows(&self, database: &Database) -> Result<Vec<User>, WhitenoiseError> {
         let user_rows = sqlx::query_as::<_, UserRow>(
             "SELECT u.id, u.pubkey, u.metadata, u.created_at, u.updated_at
              FROM account_follows af
              JOIN users u ON af.user_id = u.id
-             WHERE af.account_id = ?",
+             WHERE af.account_id = ?
+             ORDER BY u.created_at ASC, u.id ASC",
```
jgmontoya: why are we now sorting here? I don't see the benefit/need of this and it will make every call more expensive

src/whitenoise/database/users.rs
```diff
@@ -304,6 +306,28 @@ impl User {
         Ok(relays)
     }
 
+    /// Bumps `updated_at` to now without changing any other fields.
+    ///
+    /// Use this to record "we checked this user's metadata" even when no new
+    /// metadata was found, so that `needs_metadata_refresh()` respects the TTL
+    /// for empty-profile users.
+    pub(crate) async fn touch_updated_at(
```
jgmontoya, line 318: why no `#[perf_instrument("db::users")]` on this one? I think I don't quite understand which ones should have it and which ones shouldn't

src/whitenoise/database/accounts.rs
dannym-arx: good catch... not sure why... maybe it was in an old commit, and it got back added when I rebased, not sure? anyway, will definitely change that

src/whitenoise/users.rs
dannym-arx: Yeah that makes sense to me!

src/whitenoise/database/users.rs
dannym-arx: whatever we want to track the performance of should get a `#[perf_instrument]`, I can add it to this one too, if you prefer

jgmontoya: I have no preference on this

PR #619, opened 2026-03-11
fix(relay_control/sessions/session): reconnect pending relays on prepare_relay_urls

jgmontoya:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed relays getting stuck in Pending so they properly reset and reconnect.
  * New identities are immediately marked as synced after creation, preventing subscription poisoning.

* **Tests**
  * Added integration and regression tests verifying identity sync state and relay reconnect recovery.
  * Added end-to-end tests for multi-identity messaging stability, early-exit targets, timeouts, and broader message-topic matching.

### DIFF
--- a/src/whitenoise/accounts/login.rs
+++ b/src/whitenoise/accounts/login.rs
@@ -1,4 +1,4 @@
-use chrono::Utc;
+use chrono::{DateTime, Utc};
 use nostr_sdk::prelude::*;
 
 use super::{
@@ -36,7 +36,7 @@ impl Whitenoise {
         // global subscriptions to use `since=None` (unbounded re-fetch).
         let now_ms = Utc::now().timestamp_millis();
         Account::update_last_synced_max(&account.pubkey, now_ms, &self.database).await?;
-        account.last_synced_at = Some(Utc::now());
+        account.last_synced_at = DateTime::from_timestamp_millis(now_ms);
 
         let user = account.user(&self.database).await?;
 


src/whitenoise/accounts/login.rs
```diff
@@ -29,6 +30,14 @@ impl Whitenoise {
         let mut account = self.create_base_account_with_private_key(keys).await?;
         tracing::debug!(target: "whitenoise::accounts", "Keys stored in secret store and account saved to database");
 
+        // A brand new identity has no history to catch up on — mark as synced
+        // immediately. Without this, `last_synced_at = NULL` poisons
+        // `compute_global_since_timestamp()` for ALL accounts, forcing
+        // global subscriptions to use `since=None` (unbounded re-fetch).
+        let now_ms = Utc::now().timestamp_millis();
+        Account::update_last_synced_max(&account.pubkey, now_ms, &self.database).await?;
```
erskingardner, line 38: Two separate `Utc::now()` calls are made: one for the DB write (`now_ms`) and one to populate the in-memory `account.last_synced_at`. In practice the delta is sub-millisecond, but they are not guaranteed to be identical: the DB stores millisecond precision via `timestamp_millis()`, while the in-memory field carries nanosecond precision. If anything ever round-trips through the DB and compares the two values (e.g. `loaded_account.last_synced_at == account.last_synced_at`), it will fail.

Capture the time once and derive both values from it:

```rust
let now = Utc::now();
let now_ms = now.timestamp_millis();
Account::update_last_synced_max(&account.pubkey, now_ms, &self.database).await?;
// Truncate to milliseconds to match what the DB stores
account.last_synced_at = Some(now.with_nanosecond(0).unwrap_or(now));
```

Or, more simply — since `update_last_synced_max` writes to the DB and its SQL result is what matters, re-read the account from the DB after the write (if that's cheap). But capturing `now` once is the minimal fix.

src/integration_tests/test_cases/account_management/verify_new_identity_synced.rs
```diff
@@ -0,0 +1,58 @@
...
+        tracing::info!(
+            "Verifying {} accounts have last_synced_at set...",
+            self.account_names.len()
+        );
+
+        for name in &self.account_names {
+            let account = context.get_account(name)?;
+
+            assert!(
+                account.last_synced_at.is_some(),
```
erskingardner, line 37: `context.get_account()` returns the `Account` value stored in the in-memory context map at account-creation time. This is the same struct that `create_identity` returns, so it will always have `last_synced_at = Some(...)` if the production code sets it — but it does **not** prove the value was persisted to the database. The unit test in `login.rs` adds a DB-read assertion, but this integration test case does not.

Consider doing a fresh `whitenoise.find_account_by_pubkey()` fetch inside this test case to prove persistence end-to-end, or document explicitly in the test's docstring that it only validates in-memory state and points to the unit test for DB verification.

tests/cli_e2e.rs
```diff
@@ -300,3 +300,87 @@ async fn cross_daemon_group_messaging() {
...
+            "--account",
+            &alice_pk1,
+            "groups",
+            "create",
+            "Multi-Account Test",
+            &bob_pk,
+        ],
+    );
+    let gid = group_id_hex(&group);
+    wait_for_group(&bob.socket, &bob_pk).await;
```
erskingardner, line 338: The test uses two raw sleeps: 5 seconds to let the group sync propagate, and 3 seconds to let subscriptions re-settle after the second identity is added. These are the most common source of flakiness in the existing test suite. The existing helpers (`wait_for_group`, `wait_for_message`) already poll with timeouts — could the re-settle window also be covered by polling `wait_for_message` directly rather than sleeping and hoping? At minimum, consider replacing the 5-second sleep with `wait_for_group(&bob.socket, &bob_pk)` which the test already calls right after, making the sleep redundant.

Specifically, these two lines appear back-to-back:

```rust
wait_for_group(&bob.socket, &bob_pk).await;
tokio::time::sleep(Duration::from_secs(5)).await;
```

The sleep after `wait_for_group` has already verified Bob has a group — if the group is present, MLS state is settled enough to send. The 5-second sleep likely guards against a race in subscription activation, which is a valid concern, but that same concern applies to `wait_for_message` which already has a 30-second polling window. The sleep can almost certainly be dropped here.

src/integration_tests/test_cases/account_management/verify_new_identity_synced.rs
jgmontoya: DB persistence is already verified by the unit test in login.rs (test_new_identity_has_last_synced_at), which calls Account::find_by_pubkey after creation and asserts last_synced_at.is_some() on the DB-loaded account.

This test case intentionally validates the API return-value contract (what callers of create_identity receive), while the unit test covers the DB round-trip. Adding a DB read here would duplicate that coverage and blur the test's responsibility.

tests/cli_e2e.rs
jgmontoya: I think these sleeps are fine for an e2e test

`wait_for_group` just waits until there's a group in the db, but we would still want to give some time for subscriptions to be established, key packages rotated, etc..

PR #634, opened 2026-03-12
fix(welcome): reject Welcome on key package DB lookup failure

### LINKED ISSUE #622: Issue B: Welcome Message Processing Skips KeyPackage Ownership Verification and Mandatory Rotation
marmot

**Status: Partially Fixed** in commit `6a52586` (`fix(welcome): reject malformed or missing key-package e-tags`).

What was fixed:
- Welcome rumors with a missing `e` tag are now rejected before reaching MDK.
- Welcome rumors with an unparseable `e` tag are now rejected.
- `rotate_key_package` now takes a non-optional `EventId`, so the silent early return on `None` is gone.

What remains open:
- When the `e` tag is present but the database lookup returns `Err`, the handler still falls through to MLS processing with `"Failed to look up key package: {}, proceeding anyway"`. This is not safe to ignore — the handler should return an error or queue the Welcome for retry.

---

## Locations

- `src/whitenoise/event_processor/event_handlers/handle_giftwrap.rs#L91`
- `src/whitenoise/event_processor/event_handlers/handle_giftwrap.rs#L114`
- `src/whitenoise/event_processor/event_handlers/handle_giftwrap.rs#L390`

## Synopsis

When a Welcome's `EventId` was `None` or failed to parse, ownership verification was skipped entirely and the Welcome still went to MDK. A database error during the ownership check also let execution continue to MLS acceptance. On a successful Welcome, `rotate_key_package` returned early on `None`, so the key package was never marked consumed, never deleted from relays, and never replaced.

## Impact

**Medium.**

Accepting a Welcome without verifying KP ownership breaks the key lifecycle. Consumed KeyPackages stay on relays, and after enough repetitions the account's KP pool drains. The receiver has no confirmation that the Welcome was intended for a key package they own.

## Feasibility

**Medium.**

MDK still validates the Welcome cryptographically. But the `e` tag is a Nostr-layer field — it's not in the MLS ciphertext, and a noncompliant sender can omit it. The database fall-through path doesn't need an attacker; a transient storage fault is enough.

## Severity

**Medium.**

## Preconditions

- An attacker or noncompliant sender must be able to produce or relay a gift-wrapped Welcome event addressed to the victim.
- For the database fall-through path, no adversarial action is required — a fault alone is sufficient.

## Technical Details

`key_package_event_id` was parsed as `Option<EventId>`. The precheck was guarded by `if let Some(ref kp_event_id) = key_package_event_id`, so the ownership lookup was skipped entirely when the tag was absent.

When the tag was present but the database lookup returned `Err`, the handler logged a warning and fell through:

```rust
Err(e) => {
    // line 114 – DB error is non-fatal; falls through to MDK
    tracing::warn!("Failed to look up key package: {}, proceeding anyway", e);
}
```

After a successful Welcome, `rotate_key_package` took the same `Option<EventId>` and returned silently on `None`:

```rust
let Some(kp_event_id) = key_package_event_id else {
    return Ok(());
};
```

`PublishedKeyPackage::mark_consumed` was never called, relay-side deletion never happened, and `publish_key_package_for_account` was never invoked.

## Remediation

- Reject the Welcome rumor if it lacks a parseable `e` tag pointing to a known KeyPackage event ID, before invoking MDK. ✅ Done.
- Return an error or queue the Welcome for retry when the ownership database lookup fails. ⚠️ Still open.
- Return an error from `rotate_key_package` when the KeyPackage event ID is absent. ✅ Done.

mubarakcoded:
## Summary

Closes #622 (Issue B: Welcome Message Processing Skips KeyPackage Ownership Verification and Mandatory Rotation)

- Completes the final remaining remediation item from the Least Authority audit: when the key package ownership database lookup returns an error, reject the Welcome instead of falling through to MLS processing
- The returned error triggers the existing retry mechanism in `process_account_event`, so transient DB failures cause a retry rather than silently accepting an unverified Welcome

### All three audit remediation items are now addressed:
1. **Reject Welcome if missing/unparseable `e` tag** — done in `6a52586`
2. **Return error on DB lookup failure** — this PR
3. **Return error from `rotate_key_package` when KP event ID absent** — done in `6a52586`

## Test plan

- [x] `cargo check` passes
- [x] All 11 `handle_giftwrap` tests pass

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened error handling for message processing to properly reject failed database operations instead of silently continuing. When database lookups fail, appropriate error rejection and retry mechanisms are now triggered, improving system resilience.

* **Tests**
  * Added test case to verify that database failures properly trigger error rejection and upstream retry logic, ensuring robust error handling.

### DIFF
--- a/src/whitenoise/event_processor/event_handlers/handle_giftwrap.rs
+++ b/src/whitenoise/event_processor/event_handlers/handle_giftwrap.rs
@@ -1024,4 +1024,35 @@ mod tests {
             .unwrap();
         assert!(ag.is_some(), "AccountGroup must survive an early return");
     }
+
+    /// When the DB lookup for the key package fails (e.g. table missing),
+    /// `handle_giftwrap` must return an error so the upstream retry logic kicks in.
+    /// This covers the `Err(e) => return Err(e.into())` path in `process_welcome`.
+    #[tokio::test]
+    async fn test_handle_giftwrap_welcome_db_error_rejects() {
+        let (whitenoise, _data_temp, _logs_temp) = create_mock_whitenoise().await;
+
+        let creator_account = whitenoise.create_identity().await.unwrap();
+        let members = setup_multiple_test_accounts(&whitenoise, 1).await;
+        let member_account = members[0].0.clone();
+
+        // Build a real MLS Welcome giftwrap addressed to the member
+        let giftwrap_event =
+            build_welcome_giftwrap(&whitenoise, &creator_account, member_account.pubkey).await;
+
+        // Corrupt the database by dropping the published_key_packages table
+        sqlx::query("DROP TABLE published_key_packages")
+            .execute(&whitenoise.database.pool)
+            .await
+            .unwrap();
+
+        // Processing should fail because the DB lookup errors out
+        let result = whitenoise
+            .handle_giftwrap(&member_account, giftwrap_event)
+            .await;
+        assert!(
+            result.is_err(),
+            "Welcome with broken DB should return an error, got Ok"
+        );
+    }
 }


dannym-arx (approved):

erskingardner (approved):

src/whitenoise/event_processor/event_handlers/handle_giftwrap.rs
erskingardner, line 121: I know that this isn't part of this PR specifically, but it would be great if we started to think about how we would surface these sorts of errors up to users. I think probably we need a table for failures of this sort. But it's something we can explore maybe in an issue and then another PR later

PR #636, opened 2026-03-12
Improve performance of whitenoise initialization

jgmontoya:
## Summary by CodeRabbit

* **New Features**
  * Added startup benchmarking recipes and a seeded-startup benchmark script.
  * Benchmark tool gains init-only and login modes for targeted startup tests.
  * Initialization-phase timing diagnostics exposed in benchmark builds.

* **Performance**
  * Subscription setup and relay warming now run concurrently to speed startup.
  * Sessions enforce a minimum number of connected relays to improve reliability.

### DIFF
--- a/scripts/benchmark-startup-seeded.sh
+++ b/scripts/benchmark-startup-seeded.sh
@@ -38,7 +38,7 @@ echo "Generated keypair: $PUB"
 echo ""
 
 echo "=== Publishing contact list to discovery relays ==="
-cat "$FIXTURE" | nak event --sec "$SEC" $RELAYS 2>&1 || true
+cat "$FIXTURE" | nak event --sec "$SEC" $RELAYS 2>&1
 echo ""
 
 # ---------------------------------------------------------------------------


src/bin/benchmark_test.rs
```diff
@@ -32,6 +43,34 @@ async fn main() -> Result<(), WhitenoiseError> {
...
+            if !follows.is_empty() {
+                tracing::info!("Contact list synced: {} follows", follows.len());
+                break;
+            }
+            if Instant::now() > deadline {
+                tracing::warn!("Timed out waiting for contact list (30s)");
+                break;
+            }
+            tokio::time::sleep(Duration::from_millis(500)).await;
+        }
```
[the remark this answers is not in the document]
jgmontoya: follows() is a DB query against account_follows — a fresh account returns Ok(vec![]) before relay sync populates that table. Breaking on first Ok would exit the loop immediately on iteration 1, before any contacts arrive, producing an unseeded data dir. The non-empty check is the sync-completion signal, not an assumption about account state. The 30s timeout is the correct fallback for when relay sync genuinely fails.

src/relay_control/account_inbox.rs
```diff
@@ -37,6 +37,7 @@ impl AccountInboxPlaneConfig {
         config.telemetry_account_pubkey = Some(self.account_pubkey);
         config.auth_policy = self.auth_policy;
         config.reconnect_policy = self.reconnect_policy;
+        config.min_connected_relays = Some(2);
```
erskingardner, line 40: what will happen if a user has only one inbox relay? (like I do)

jgmontoya: it's handled correctly :)

jgmontoya: that is, if it fails to connect to that one it fails altogether and if it succeeds it continues normally

PR #637, opened 2026-03-13
fix: enforce HTTPS on Blossom URLs and add upload timeout

### LINKED ISSUE #627: Issue G: Missing HTTPS Enforcement on Blossom URLs
marmot

**Status: Not fixed.** `Url::parse` is called on `blossom_url` values from peer-controlled group data with no scheme check. HTTP URLs are accepted and used as-is for downloads.

---

## Locations

- `src/whitenoise/groups.rs#L815` (now `src/whitenoise/groups/media.rs` after refactor)
- `src/whitenoise/groups.rs#L926`
- `src/whitenoise/groups.rs#L1211`

## Synopsis

Blossom URLs stored in the `media_files` table come from peer-controlled group data. They're used for media downloads with no check that the scheme is `https`. A group admin can set an `http://` URL and the client will issue a cleartext request to whatever server they specify.

## Impact

**Low.**

The media content itself is safe — blobs are encrypted and hash-verified, so cleartext transmission doesn't expose the plaintext. But the HTTP request leaks the blob hash, the client's IP address, and the download timing to the server operator and any network observer. A malicious group admin can use this to track which members fetch a given resource and link group participation to IP addresses.

## Feasibility

**Medium.**

A group admin can set a group image or contribute media metadata with an HTTP URL. No network position is required — the victim's own client makes the request.

## Severity

**Low.**

## Preconditions

The attacker must be a group admin or must otherwise get a peer-specified Blossom URL into the group data.

## Technical Details

`download_and_cache_group_image` takes `blossom_url`, parses it with `Url::parse`, and passes it directly to `download_blob_from_blossom` — no scheme check. `download_and_decrypt_chat_media_blob` does the same with `media_file.blossom_url`. `Url::parse` accepts any scheme, so `http://` goes through without complaint.

## Remediation

After parsing, check that `url.scheme() == "https"` and return an error for anything else.

### LINKED ISSUE #631: Suggestion 2: Add Timeout to upload_profile_picture Function
marmot

**Status: Not fixed.** `upload_profile_picture` in `src/whitenoise/accounts/mod.rs` calls `client.upload_blob(...).await` with no timeout.

---

## Location

`src/whitenoise/accounts.rs#L404` (now `src/whitenoise/accounts/mod.rs#L499` after refactor)

## Synopsis

`upload_profile_picture` has no timeout on the Blossom upload. A slow or unresponsive server will stall the call indefinitely.

## Mitigation

Wrap the upload with `tokio::time::timeout`:

```rust
tokio::time::timeout(Duration::from_secs(30), client.upload_blob(...))
    .await
    .map_err(|_| WhitenoiseError::Other(anyhow::anyhow!("upload timed out")))?
```

mubarakcoded:
## Summary
- **#627**: Add `require_https()` gate on all Blossom download and upload paths to reject non-HTTPS URLs, preventing cleartext metadata leakage from peer-controlled HTTP URLs. Debug builds permit `http://localhost` for local testing.
- **#631**: Wrap `upload_profile_picture`'s `BlossomClient::upload_blob()` call with `tokio::time::timeout` (300s) so a stalled or unresponsive server cannot block the caller indefinitely.
- Add `BlossomInsecureUrl` error variant for clear error reporting.
- Move `use nostr::hashes::...` from inside function body to file-level import per STYLE.md.

Closes #627
Closes #631

## Test plan
- [x] Unit tests for `require_https`: HTTPS accepted, HTTP rejected, FTP rejected, localhost HTTP allowed in debug, non-localhost HTTP rejected in debug
- [x] Error display test for `BlossomInsecureUrl` variant
- [x] `cargo fmt` clean
- [x] `cargo clippy` clean (zero warnings)
- [x] Full test suite passes (1241 tests)

## Summary by CodeRabbit

* **New Features**
  * Image and profile-picture uploads now enforce a timeout to avoid long-running uploads.
  * Blossom server URLs are validated to require HTTPS (localhost allowed in debug builds).

* **Improved Error Handling**
  * Clear timeout error when uploads exceed the allowed duration.
  * New explicit error when a Blossom URL is not HTTPS.

* **Tests**
  * Added tests for HTTPS validation, upload timeout behavior, URL/key parsing, and giftwrap target validation.

### DIFF
--- a/src/whitenoise/accounts/mod.rs
+++ b/src/whitenoise/accounts/mod.rs
@@ -535,6 +535,8 @@ impl Account {
             })?
             .map_err(|err| WhitenoiseError::Other(anyhow::anyhow!(err)))?;
 
+        Whitenoise::require_https(&descriptor.url)?;
+
         Ok(descriptor.url.to_string())
     }
 
--- a/src/whitenoise/groups/media.rs
+++ b/src/whitenoise/groups/media.rs
@@ -652,7 +652,7 @@ impl Whitenoise {
 
     /// Rejects non-HTTPS Blossom URLs to prevent cleartext metadata leakage.
     /// Debug builds also allow `http://localhost` for local testing.
-    fn require_https(url: &Url) -> Result<()> {
+    pub(crate) fn require_https(url: &Url) -> Result<()> {
         match url.scheme() {
             "https" => Ok(()),
             "http" if cfg!(debug_assertions) && url.host_str() == Some("localhost") => Ok(()),
@@ -859,7 +859,7 @@ impl Whitenoise {
             Some(upload_keypair),
         );
 
-        tokio::time::timeout(Self::BLOSSOM_TIMEOUT, upload_future)
+        let descriptor = tokio::time::timeout(Self::BLOSSOM_TIMEOUT, upload_future)
             .await
             .map_err(|_| {
                 WhitenoiseError::Other(anyhow::anyhow!(
@@ -867,7 +867,11 @@ impl Whitenoise {
                     Self::BLOSSOM_TIMEOUT.as_secs()
                 ))
             })?
-            .map_err(|err| WhitenoiseError::Other(anyhow::anyhow!(err)))
+            .map_err(|err| WhitenoiseError::Other(anyhow::anyhow!(err)))?;
+
+        Self::require_https(&descriptor.url)?;
+
+        Ok(descriptor)
     }
 }
 


src/whitenoise/accounts/mod.rs
```diff
@@ -489,6 +490,9 @@ impl Account {
         Ok(())
     }
 
+    /// Timeout for Blossom upload operations.
+    const UPLOAD_TIMEOUT: Duration = Duration::from_secs(300);
```
jgmontoya: we have a `BLOSSOM_TIMEOUT` constant in `src/whitenoise/groups/media.rs`

src/whitenoise/groups/media.rs
```diff
@@ -636,13 +637,27 @@ impl Whitenoise {
...
+    }
+
     async fn download_blob_from_blossom(
         blossom_url: &Url,
         image_hash: &[u8; 32],
     ) -> Result<Vec<u8>> {
-        use nostr::hashes::{Hash, sha256::Hash as Sha256Hash};
-
-        let client = BlossomClient::new(blossom_url.clone());
+        let client = Self::blossom_client(blossom_url)?;
```
erskingardner, line 674: This only checks the starting `blossom_url`. `nostr_blossom::BlossomClient` builds a `reqwest` client with redirects enabled, so an HTTPS Blossom endpoint can still answer `GET /<sha>` with `30x Location: http://...` and the request will follow it. In that scenario `download_blob_from_blossom()` still fetches the blob over cleartext, so the new HTTPS guard is bypassed whenever a server or TLS-terminating proxy redirects blob requests.

I'm not sure we have to (or even have the ability to care about that?) but worth maybe looking at quickly.

src/whitenoise/groups/media.rs
```diff
@@ -803,7 +818,7 @@ impl Whitenoise {
         mime_type: &str,
         upload_keypair: &Keys,
     ) -> Result<nostr_blossom::bud02::BlobDescriptor> {
-        let client = BlossomClient::new(blossom_server_url.clone());
+        let client = Self::blossom_client(blossom_server_url)?;
```
erskingardner, line 854: This validates only `blossom_server_url`; the code still trusts `BlobDescriptor.url` and stores/returns it unchanged in `upload_group_image()`, `upload_chat_media()`, and `upload_profile_picture()`. On HTTPS deployments behind a reverse proxy that emits `http://...` descriptor URLs, uploads succeed here, but the next cache miss or restart feeds that stored URL into `download_blob_from_blossom()` and hits `BlossomInsecureUrl`, making freshly uploaded media unreadable.

mubarakcoded: Valid point. Added a post-upload require_https check on the descriptor URL in both upload_encrypted_blob_to_blossom and upload_profile_picture. If a reverse proxy causes the server to return an http:// descriptor URL, we now reject it immediately at upload time rather than silently storing it and failing later on download.

src/whitenoise/groups/media.rs
mubarakcoded: Good point — BlossomClient owns the reqwest::Client internally, and its redirect policy is private, so we can't control where it follows redirects. If an HTTPS origin responds with a 30x to an HTTP target, the request will follow it over cleartext. Not much we can do about that on our end without switching to our own HTTP client for downloads.

PR #641, opened 2026-03-13
feat(zapstore): add fetch_latest_zapstore_version

dannym-arx:
marmot

Adds \`fetch_latest_zapstore_version\`, a single async function that queries the Zapstore relay for the kind-32267 Software Application event and extracts the current release version from its \`a\` tag.

The \`a\` tag on a kind-32267 event points at the canonical release in the form \`30063:<pubkey>:<identifier>@<version>\`. Reading from that pointer avoids the race condition that comes from scanning kind-30063 release events by recency. One event, one source of truth.

The implementation spins up a temporary \`nostr-sdk\` client, fetches from \`relay.zapstore.dev\` with a 10-second timeout, disconnects, and returns \`Option<String>\`.

Ten unit tests cover the tag-parsing logic: valid version extraction, missing tag, wrong kind, wrong pubkey, wrong identifier, missing version suffix, and multiple-tag precedence. All run without network access.

## Summary by CodeRabbit

* **New Features**
  * Added a public API to fetch the latest Zapstore version from a relay, with network fetch, timeout handling, and graceful absence reporting so users can query current version info.
* **Tests**
  * Added unit tests covering version extraction, tag/event guard conditions, ordering, and edge cases to improve reliability.

### DIFF
--- a/src/whitenoise/zapstore.rs
+++ b/src/whitenoise/zapstore.rs
@@ -9,6 +9,13 @@ const ZAPSTORE_APP_PUBKEY: &str =
     "75d737c3472471029c44876b330d2284288a42779b591a2ed4daa1c6c07efaf7";
 const ZAPSTORE_APP_IDENTIFIER: &str = "org.parres.whitenoise";
 const FETCH_TIMEOUT_SECS: u64 = 10;
+const ZAPSTORE_A_TAG_PREFIX: &str = concat!(
+    "30063:",
+    "75d737c3472471029c44876b330d2284288a42779b591a2ed4daa1c6c07efaf7",
+    ":",
+    "org.parres.whitenoise",
+    "@",
+);
 
 /// Fetches the latest version string published on Zapstore for White Noise.
 ///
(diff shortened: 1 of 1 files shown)


src/whitenoise/zapstore.rs
```diff
@@ -0,0 +1,277 @@
...
+    let events = fetch_result.map_err(|e| {
+        tracing::warn!(
+            target: "whitenoise::zapstore",
+            error = %e,
+            "failed to fetch app event from relay"
+        );
+        WhitenoiseError::from(e)
+    })?;
+
+    let event = match events.first() {
```
jgmontoya, line 67: is it guaranteed to be the first? Or should we sort by created_at perhaps?

dannym-arx: I'll add a sort, tomorrow tho

dannym-arx: actually: 

`fetch_events` returns `nostr_database::Events`, which is a `BTreeCappedSet<Event>` documented as a "descending sorted collection of events".

erskingardner: yeah nostr-sdk sorts already.

PR #642, opened 2026-03-14
fix(security): cap media blob downloads at 100 MiB to prevent DoS (closes #624)

### LINKED ISSUE #624: Issue D: Unbounded Media Downloads Are a Potential Denial-of-Service (DoS) Attack Vector
marmot

**Status: Not fixed.** No `Content-Length` check, streaming byte counter, or server whitelist has been added. Downloads are still fully buffered in memory, with only the existing 300-second timeout providing any constraint.

---

## Locations

- `src/whitenoise/media_files.rs#L257`
- `src/whitenoise/groups.rs#L1112`

## Synopsis

The application downloads media blobs into memory from URLs that group members control, with no size cap and no server whitelist. A malicious group member can point a message at a server they own and make the victim download an arbitrarily large file.

## Impact

**Medium.**

The application can become slow, unresponsive, or crash from memory pressure.

## Feasibility

**Medium.**

The attack is straightforward: send a valid MLS message with an `imeta` URL pointing to attacker-controlled infrastructure. The victim still has to initiate the download, and the attacker must be a group member — those are the main constraints.

## Severity

**Medium.**

## Preconditions

- The attacker must be a group member.
- The attacker must control a server that serves oversized payloads.
- The victim must initiate the download.

## Technical Details

Untrusted media URLs enter through MLS message content. `handle_mls_message` calls `parse_imeta_tags_from_event` and persists the URL to the database as `blossom_url` with no whitelist check, so it can point anywhere. The download path reads from that URL and buffers the full response in memory with no byte cap. A 300-second timeout exists but offers only partial protection — a fast server on a fast connection can fill memory well within that window.

## Remediation

- Check the `Content-Length` header before downloading (keeping in mind it can be spoofed).
- Switch to a streaming read that counts bytes and aborts above a defined threshold.
- Maintain a server whitelist of approved Blossom servers and reject URLs outside it.

dannym-arx:
marmot

Group members control the URLs in `imeta` tags. Before this fix, `download_blob_from_blossom` called `nostr-blossom`'s `get_blob()`, which calls `response.bytes().await` with no size limit. A malicious member pointing at their own server could fill the victim's memory.

This replaces the unbounded `get_blob()` call with a two-layer defence:

1. **Content-Length fast path:** rejects the response immediately if the server advertises more than 100 MiB, before any body bytes are transferred.
2. **Streaming byte counter:** the authoritative guard. Reads the body chunk-by-chunk and aborts with `DownloadSizeLimitExceeded` the moment the running total would exceed 100 MiB, regardless of what `Content-Length` said (or omitted).

Both layers report the same `DownloadSizeLimitExceeded` error so callers have a single variant to handle.

A `Vec::with_capacity` hint is used when a valid `Content-Length` is present, avoiding repeated allocations for well-behaved servers. The `u64` to `usize` conversion uses `try_from` with a `usize::MAX` fallback to keep the logic correct on 32-bit targets.

The 300-second timeout introduced previously remains in place as the third line of defence.

Reported by Least Authority TFA GmbH in the whitenoise-rs Initial Audit Report (2026-03-11), Issue D.

## Summary by CodeRabbit

* **New Features**
  * Media downloads now stream for more efficient, incremental retrieval.
  * Downloads enforce HTTPS and a 100 MiB cap to protect against oversized transfers.
  * No public API changes.

* **Bug Fixes**
  * Requests that exceed the size limit now fail with a clear error instead of silent failure.

* **Tests**
  * Added tests covering content-length guards, streaming limits, and successful small-body downloads.

### DIFF
--- a/src/whitenoise/groups/media.rs
+++ b/src/whitenoise/groups/media.rs
@@ -2,13 +2,20 @@ use std::path::{Path, PathBuf};
 use std::sync::LazyLock;
 use std::time::Duration;
 
+use base64ct::{Base64, Encoding as _};
 use futures::StreamExt;
 use mdk_core::encrypted_media::types::MediaReference;
 use mdk_core::extension::group_image;
 use mdk_core::media_processing::MediaProcessingOptions;
 use mdk_core::prelude::{GroupId, group_types};
 use mdk_storage_traits::Secret;
+use nostr_blossom::bud01::{
+    BlossomAuthorization, BlossomAuthorizationScope, BlossomAuthorizationVerb,
+    BlossomBuilderExtension,
+};
 use nostr_blossom::client::BlossomClient;
+use nostr_sdk::prelude::hashes::Hash;
+use nostr_sdk::prelude::hashes::sha256::Hash as Sha256Hash;
 use nostr_sdk::prelude::*;
 use sha2::{Digest, Sha256};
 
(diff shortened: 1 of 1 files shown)


src/whitenoise/groups/media.rs
```diff
@@ -649,18 +645,77 @@ impl Whitenoise {
...
+        let download_future = async {
+            // Ensure the ring TLS provider is installed before building the client.
+            // reqwest 0.13 with rustls-no-provider requires an explicit provider.
+            // The `let _ =` ignores the error when it is already installed.
+            let _ = rustls::crypto::ring::default_provider().install_default();
 
-        let client = BlossomClient::new(blossom_url.clone());
-        let sha256 = Sha256Hash::from_slice(image_hash)
-            .map_err(|e| WhitenoiseError::Other(anyhow::anyhow!("Invalid SHA256 hash: {}", e)))?;
+            let client = reqwest::Client::builder()
```
erskingardner: `reqwest::Client` is designed to be created once and shared. Internally it holds a connection
pool, a DNS resolver, and TLS state. Rebuilding it on every download:

- defeats keep-alive / connection reuse (each call incurs a full TLS handshake);
- allocates and drops significant resources on every media download;
- calls `install_default()` before every request, which is only safe because of the `let _ =`
  suppression — this is more of a smell than a bug, but it signals the client creation logic
  belongs at a higher level.

The right fix is to create the `reqwest::Client` once — either as a `Whitenoise` field, a
`lazy_static` / `OnceLock`, or at minimum as a function-level static. Since `Whitenoise`
already owns long-lived resources, adding a `reqwest::Client` field there would be most
consistent with the existing architecture.

src/whitenoise/groups/media.rs
```diff
@@ -649,18 +645,77 @@ impl Whitenoise {
...
+            // Stream the body chunk-by-chunk, counting bytes as they arrive.
+            // This is the authoritative size guard: it fires even when the server
+            // omits or lies about Content-Length.
+            let mut body: Vec<u8> = Vec::with_capacity(hint_capacity);
+            let mut stream = response.bytes_stream();
+            while let Some(chunk) = stream.next().await {
+                let chunk = chunk.map_err(|e| {
+                    WhitenoiseError::BlossomDownload(format!("Error reading response body: {}", e))
+                })?;
+                if body.len() + chunk.len() > Self::MAX_BLOB_BYTES {
```
erskingardner: If `body.len()` is near `usize::MAX` (impossible in practice but true in theory, and
especially relevant on 32-bit targets where `MAX_BLOB_BYTES = 100 * 1024 * 1024 ≈ 100M` and
`usize::MAX ≈ 4G`) this addition wraps and the guard is bypassed. Use a saturating add or
restructure the check:

```rust
if chunk.len() > Self::MAX_BLOB_BYTES - body.len() {
```

This is always safe because `body.len()` never exceeds `MAX_BLOB_BYTES` at this point (the
loop aborts before allowing that), so the subtraction never underflows.

src/whitenoise/groups/media.rs
erskingardner, line 16: The PR removes the `BlossomClient` usage from `download_blob_from_blossom` but
`BlossomClient` is still used in `upload_encrypted_blob_to_blossom` (line 839), so this is
not a dead import. However, the old `download_blob_from_blossom` also used
`nostr::hashes::{Hash, sha256::Hash as Sha256Hash}` inside the function scope — the diff
shows these were removed correctly. Worth confirming no `cargo check` warnings surface.

src/whitenoise/groups/media.rs
```diff
@@ -649,18 +645,77 @@ impl Whitenoise {
...
+                        limit: Self::MAX_BLOB_BYTES,
+                    });
+                }
+                body.extend_from_slice(&chunk);
+            }
+
+            Ok(body)
+        };
 
         tokio::time::timeout(Self::BLOSSOM_TIMEOUT, download_future)
```
erskingardner: The original code needed a named future because `nostr-blossom`'s `get_blob()` returns an
opaque future that must be passed to `timeout`. Now that the body is directly `await`ed
inside an `async fn`, wrapping it in an extra `async { }` block just to hand it to `timeout`
is unnecessary indirection. The function itself is `async`, so you can write:

```rust
tokio::time::timeout(Self::BLOSSOM_TIMEOUT, async {
    // ... entire body inline ...
}).await...
```

or, more readably, extract the inner logic into a private helper and call `timeout` at the
callsite. The current nesting adds a level of indentation without adding clarity.

src/whitenoise/groups/media.rs
dannym-arx: I mean... if someone downloads 18 exabytes of data, or whatever `2^64 -1` is, they may have a different problem lol

And as for 32bit cpus, we've moved on from 32 bit CPUs a long time ago, no major hardware manufacturer in the last 10 years (or longer) has manufactured a 32 bit cpu... ESP32 being the only exception I can think of, and, if anyone is running whitenoise on an ESP32 I'd love to see it, haha

src/whitenoise/groups/media.rs
dannym-arx: Cargo check passes just fine, as does just precommit... not sure what your agent was thinking, this is clearly a hallucination

src/whitenoise/groups/media.rs
dannym-arx: also on 32 bit we can't even allocate that memory, the cpu is gonna OOM at :703... I'll add the check anyways, but it's not needed imo

PR #643, opened 2026-03-14
perf(whitenoise/groups): improve performance of create_group method

jgmontoya:
## Summary by CodeRabbit

* **New Features**
  * Added a "group-creation" benchmark scenario and test coverage for measuring group creation performance.

* **Refactor**
  * Reworked group-creation flow to resolve members/key packages concurrently, handle missing or invalid keys more robustly, finalize local group records, and publish welcome messages asynchronously for more reliable and faster group setup.
  * Simplified keyring/test configuration gating to better support integration and benchmark builds.

### DIFF
--- a/src/integration_tests/benchmarks/scenarios/group_creation.rs
+++ b/src/integration_tests/benchmarks/scenarios/group_creation.rs
@@ -1,0 +1,134 @@
+use std::time::Duration;
+
+use async_trait::async_trait;
+
+use crate::WhitenoiseError;
+use crate::integration_tests::benchmarks::test_cases::CreateGroupBenchmark;
+use crate::integration_tests::benchmarks::{BenchmarkConfig, BenchmarkScenario, BenchmarkTestCase};
+use crate::integration_tests::core::ScenarioContext;
+
+/// Default number of members per group during benchmark iterations.
+const DEFAULT_MEMBERS_PER_GROUP: usize = 2;
+
+/// Benchmark scenario for measuring `create_group` performance.
+///
+/// This scenario benchmarks the MLS group creation flow, which includes:
+/// - Fetching key packages for each member from relays
+/// - Creating the MLS group via MDK
+/// - Publishing welcome messages to each member's inbox relays
+/// - Creating group information and account-group records
+/// - Refreshing account subscriptions
+///
+/// **Setup phase** (not timed):
+/// - Creates a single creator account that persists across all iterations
+/// - Creates N×members_per_group fresh member accounts (one set per iteration)
+///   Each member account has key packages published to the test relays via
+///   `create_identity()`, which handles relay lists and key package publishing.
+///
+/// **Benchmark iteration** (timed):
+/// - Calls `create_group` with the creator and a fresh set of members
+///
+/// MLS key packages are single-use, so each iteration requires unique members.
+pub struct GroupCreationBenchmark {
+    members_per_group: usize,
+    test_case: Option<CreateGroupBenchmark>,
+}
+
+impl GroupCreationBenchmark {
+    pub fn new(members_per_group: usize) -> Self {
+        Self {
+            members_per_group,
+            test_case: None,
+        }
+    }
+}
+
+impl Default for GroupCreationBenchmark {
+    fn default() -> Self {
+        Self::new(DEFAULT_MEMBERS_PER_GROUP)
+    }
+}
+
+#[async_trait]
+impl BenchmarkScenario for GroupCreationBenchmark {
+    fn name(&self) -> &str {
+        "Group Creation Performance"
+    }
+
+    fn config(&self) -> BenchmarkConfig {
+        BenchmarkConfig {
+            // Group creation is network-heavy (key package fetches, welcome publishes);
+            // keep iterations modest for reasonable benchmark duration.
+            iterations: 10,
+            warmup_iterations: 1,
+            cooldown_between_iterations: Duration::from_millis(500),
+        }
+    }
+
+    async fn setup(&mut self, context: &mut ScenarioContext) -> Result<(), WhitenoiseError> {
+        let config = self.config();
+        let total_iterations = (config.iterations + config.warmup_iterations) as usize;
+        let total_members = total_iterations * self.members_per_group;
+
+        tracing::info!(
+            "Setting up group creation benchmark: {} iterations, {} members per group, {} total member accounts to create",
+            total_iterations,
+            self.members_per_group,
+            total_members,
+        );
+
+        // Create the creator account (reused across all iterations)
+        let creator = context.whitenoise.create_identity().await?;
+        tracing::info!("Created creator account: {}", creator.pubkey.to_hex());
+        context.add_account("creator", creator);
+
+        // Create member accounts — each gets key packages published via create_identity()
+        let mut member_sets: Vec<Vec<String>> = Vec::with_capacity(total_iterations);
+
+        for i in 0..total_iterations {
+            let mut members_for_iteration = Vec::with_capacity(self.members_per_group);
+
+            for j in 0..self.members_per_group {
+                let member_name = format!("member_{}_{}", i, j);
+                let member = context.whitenoise.create_identity().await?;
+                context.add_account(&member_name, member);
+                members_for_iteration.push(member_name);
+            }
+
+            member_sets.push(members_for_iteration);
+
+            if (i + 1) % 5 == 0 || i + 1 == total_iterations {
+                tracing::info!(
+                    "Prepared member accounts for {}/{} iterations",
+                    i + 1,
+                    total_iterations
+                );
+            }
+        }
+
+        // Brief pause to let relays index published key packages
+        tokio::time::sleep(Duration::from_secs(2)).await;
+
+        tracing::info!(
+            "Setup complete: 1 creator + {} member accounts ({} groups × {} members)",
+            total_members,
+            total_iterations,
+            self.members_per_group,
+        );
+
+        self.test_case = Some(CreateGroupBenchmark::new("creator", member_sets));
+
+        Ok(())
+    }
+
+    async fn single_iteration(
+        &self,
+        context: &mut ScenarioContext,
+    ) -> Result<Duration, WhitenoiseError> {
+        self.test_case
+            .as_ref()
+            .expect("test_case must be initialized in setup()")
+            .run_iteration(context)
+            .await
+    }
+}
--- a/src/whitenoise/groups.rs
+++ b/src/whitenoise/groups.rs
@@ -1,6 +1,8 @@
-use std::collections::{BTreeSet, HashMap, HashSet};
+use std::collections::{BTreeSet, HashMap};
+use std::sync::Arc;
 use std::time::Duration;
 
+use futures::future::{join_all, try_join_all};
 use mdk_core::prelude::*;
 use nostr_sdk::prelude::*;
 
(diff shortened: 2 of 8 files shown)


src/integration_tests/benchmarks/scenarios/group_creation.rs
```diff
@@ -0,0 +1,134 @@
...
+                tracing::info!(
+                    "Prepared member accounts for {}/{} iterations",
+                    i + 1,
+                    total_iterations
+                );
+            }
+        }
+
+        // Brief pause to let relays index published key packages
+        tokio::time::sleep(Duration::from_secs(2)).await;
```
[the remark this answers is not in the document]
jgmontoya: The 32ms standard deviation across iterations empirically demonstrates
  that any subscription overhead from create_identity() is negligible
  relative to the measured signal. The suggested
  publish_key_package_without_subscription infrastructure doesn't exist and
  would need to be purpose-built for a development-only benchmark — that's
  not a worthwhile trade-off for a non-existent measurement problem.

  The 2s sleep matches the pattern used by other benchmark scenarios (e.g.
  login_performance.rs). We can revisit if we ever see flaky benchmark
  results, but the current data doesn't support that concern.

src/whitenoise/groups.rs
```diff
@@ -265,6 +221,138 @@ impl Whitenoise {
...
+                if let Err(error) = result {
+                    tracing::warn!(
+                        target: "whitenoise::accounts::groups::create_group",
+                        "Background welcome publish failed: {}",
+                        error
+                    );
+                }
+            }
+        });
+    }
```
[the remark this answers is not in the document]
jgmontoya: This is a deliberate design decision. The group IS created, Ok(group) is accurate. Welcome delivery is best-effort network I/O: even synchronous publishing can silently fail (relay drops the event, network partition, etc.). The relay layer already retries with exponential backoff (3 attempts in publish_event_to_scope). If a welcome isn't received, the member simply doesn't see the invitation, there's no corrupted state.
Moving delivery to background is one of the purposes of this performance PR.

src/whitenoise/groups.rs
```diff
@@ -86,165 +88,120 @@ impl Whitenoise {
...
+                let kp_event_id = rumor.tags.event_ids().next().ok_or(WhitenoiseError::Other(
+                    anyhow::anyhow!("No event ID found in welcome rumor"),
+                ))?;
+                let member_pubkey = kp_pubkey_by_event_id.get(kp_event_id).copied().ok_or(
+                    WhitenoiseError::Other(anyhow::anyhow!(
+                        "No public key found in key package event"
+                    )),
+                )?;
+                Ok((rumor, member, member_pubkey))
+            })
```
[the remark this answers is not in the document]
jgmontoya: Verified by tracing through the mdk-core source.
  build_welcome_rumors_for_key_packages (groups.rs:1740) iterates the input
  key_package_events vec in order with a simple for event in
  key_package_events loop, pushing each rumor to welcome_rumors_vec 
  sequentially. The ordering is guaranteed by construction — it's not an
  assumption about MLS protocol behavior. The zip in prepare_welcomes safely
   mirrors this same input order.

src/whitenoise/groups.rs
```diff
@@ -265,6 +221,141 @@ impl Whitenoise {
...
+                Ok(wn) => wn,
+                Err(error) => {
+                    tracing::error!(
+                        target: "whitenoise::accounts::groups::create_group",
+                        "Failed to get Whitenoise instance for background welcome publishing: {}",
+                        error
+                    );
+                    return;
+                }
+            };
```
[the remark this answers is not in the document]
jgmontoya: I don't think we can use self here because of the tokio::spawn, as I believe it requires a 'static future so &self can't be moved into the closure

src/whitenoise/groups.rs
```diff
@@ -265,6 +221,141 @@ impl Whitenoise {
...
+                            .publish_welcome(
+                                &member_pubkey,
+                                rumor,
+                                &[Tag::expiration(one_month_future)],
+                                creator.pubkey,
+                                &Relay::urls(&relays_to_use),
+                                signer,
+                            )
+                            .await
+                            .map_err(WhitenoiseError::from)?;
```
dannym-arx, line 286: If `publish_welcome` fails for any member, the failure is only logged; there's no retry and the caller always gets `Ok(group)`. A transient relay failure at this point permanently locks that member out of the group, unless the member is removed and re-invited

jgmontoya: `publish_welcome` calls `publish_gift_wrap_to` -> `publish_event_to` -> `publish_event_to_scope` which is where the retries are (3 retries with exponential backoff).

It is true though that if those fail then removing the member and having him re-invited is the only way to get him back on the group, however I'd argue that it's still better than the previous design: welcomes were sent before local group finalization (`GroupInformation`, `AccountGroup`) and used `?` propagation. A single member's welcome failure would abort the entire method leaving an orphaned MLS group with no DB records (on the whitenoise db), while already-welcomed members held invites to an improperly registered group.

The new design finalizes all local state first, then publishes welcomes in the background. A welcome failure affects only that specific member, and the recovery path is clean.

dannym-arx: yeah I agree with you, but I feel like we should have some way to handle this (perhaps in another PR)

src/whitenoise/groups.rs
```diff
@@ -265,6 +221,141 @@ impl Whitenoise {
...
+
+    /// Delivers welcome messages to group members in a background task.
+    /// Uses `join_all` to attempt all members even if individual publishes fail.
+    fn background_publish_welcomes(
+        welcome_data: Vec<(UnsignedEvent, User, PublicKey)>,
+        signer: Arc<dyn NostrSigner>,
+        creator_account: Account,
+    ) {
+        tokio::spawn(async move {
+            let whitenoise = match Whitenoise::get_instance() {
```
erskingardner, line 246: `background_publish_welcomes` is a free `fn` (not a method), but inside the spawned task it calls `Whitenoise::get_instance()` to reach the global singleton. This has two problems:

1. **Tests are silently broken.** Any test using `create_mock_whitenoise()` never initializes the global singleton. `get_instance()` returns `Err`, the task silently returns, and no welcomes are ever published. The background publish path is therefore untested by the existing test suite.
2. **Semantically wrong.** The `create_group` caller already has `&self` — there's no reason to re-fetch from a global. This will fail in any embedding that doesn't use the singleton pattern.

Fix: pass an `Arc<Whitenoise>` (or just the sub-fields `relay_control.clone()`, `database.clone()`) into the background task instead of calling `get_instance()` inside it. Since `create_group` is a method, the simplest approach is to call `Whitenoise::get_instance()?` once in `create_group` (surfacing the error to the caller) and pass the result into `background_publish_welcomes`.

src/whitenoise/groups.rs
```diff
@@ -86,165 +88,120 @@ impl Whitenoise {
...
+        let kp_pubkey_by_event_id: HashMap<EventId, PublicKey> = key_package_events
+            .iter()
+            .map(|event| (event.id, event.pubkey))
+            .collect();
 
-        let _ = (group_ids, relays, signer);
-        self.refresh_account_subscriptions(creator_account).await?;
+        welcome_rumors
+            .into_iter()
+            .zip(members)
```
erskingardner: `member_pubkey` is correctly resolved by key-package event ID, but `member` (the `User` record used to look up delivery relays) is taken from a positional `zip`. If the MLS library emits `welcome_rumors` in a different order than the input `key_package_events`, a welcome will be published to the **wrong member's relay set**. The rumor itself is encrypted for the right key, but it's delivered to the wrong relay — that member never receives it.

Using two different pairing strategies (positional for `member`, ID-based for `member_pubkey`) in the same closure is the clearest indicator this is a latent ordering bug. Unify them:

erskingardner: This also makes the `welcome_rumors.len() != members.len()` length guard unnecessary (the map lookup will fail gracefully if a member is missing).

src/whitenoise/mod.rs
erskingardner, line 366: The inner `initialize_keyring_store()` now routes `feature = "benchmark-tests"` to the mock store, but the public convenience wrapper is only compiled for `test` and `integration-tests`. Any benchmark binary that calls `Whitenoise::initialize_mock_keyring_store()` directly will fail to compile with `benchmark-tests` enabled.

Fix:

```rust
#[cfg(any(test, feature = "integration-tests", feature = "benchmark-tests"))]
pub fn initialize_mock_keyring_store() {
    Self::initialize_keyring_store();
}
```

src/whitenoise/groups.rs
```diff
@@ -265,6 +221,141 @@ impl Whitenoise {
...
+        self.finalize_group_records(
+            &group,
+            &member_pubkeys,
+            group_type,
+            &group_name,
+            creator_account,
+        )
+        .await?;
+
+        Self::background_publish_welcomes(welcome_data, signer, creator_account.clone());
```
erskingardner, line 373: `create_group` returns `Ok(group)` regardless of whether any welcome was delivered. If `background_publish_welcomes` fails for all members (transient relay outage, etc.), those members are permanently excluded from the group with no retry mechanism and no error visible to the caller or the UI. The group has been committed locally and the creator's side is fully set up, but the invitees have no path to join.

Note: `add_members_to_group` does **not** have this problem — welcomes there are awaited synchronously and errors propagate. This is an asymmetry in the API that will confuse future maintainers.

This was flagged by `dannym-arx` in review and is still unresolved. Minimum acceptable options:

1. **Persist and retry**: Queue failed welcomes and retry on next startup.
2. **Surface partial failure**: Return `(Group, Vec<PublishFailure>)` so callers can warn users.
3. **If keeping fire-and-forget**: Document the failure mode prominently in the doc comment and add a test that covers the scenario, even if only to verify the warning log fires.

erskingardner (commented):
Added a few agent comments

src/whitenoise/mod.rs
jgmontoya: benchmark-tests implies integration-tests in Cargo.toml (benchmark-tests = ["integration-tests", "indicatif"]), so the #[cfg(any(test, feature = "integration-tests"))] gate is satisfied. The benchmark binary also doesn't call initialize_mock_keyring_store(), it goes through initialize_whitenoise() which handles keyring setup internally.

src/whitenoise/groups.rs
jgmontoya: I've added some documentation to it. I agree that the current situation has failure modes that are not ideal and we should handle better. But I think we can defer that to a future PR and leave this one as-is as I believe it's already way better than the current design.

src/whitenoise/groups.rs
jgmontoya: `tokio::spawn` requires 'static, so &self can't be moved in, this is why `get_instance()` is used inside all 7 background spawns across the codebase (e.g. `background_refresh_account_group_subscriptions`, `handle_giftwrap`, `handle_contact_list`). Calling get_instance()? in `create_group` itself would break the unit test suite, test_create_group and 10+ other tests use create_mock_whitenoise() which doesn't set the global. `Whitenoise` and `RelayControlPlane` are neither Clone nor Arc-wrapped, so passing sub-fields isn't feasible today.

If we want to address this I suggest we do it on a separate PR where we can evaluate the right approach across all 7 call sites.

PR #644, opened 2026-03-16
fix: filter non-actionable relay messages from telemetry pipeline

### LINKED ISSUE #606: OK/EOSE/Count relay messages incorrectly mapped to Notice telemetry
In <notifications.rs:62>-86, RelayMessage::Ok, EndOfStoredEvents, Count, NegMsg, and NegErr are all mapped to RelayNotification::Notice with hardcoded string messages like "ok" and "eose". These flow through process_notification which emits RelayTelemetryKind::Notice telemetry and forwards ProcessableEvent::RelayMessage.
This causes three problems:
	1.	Every OK response and EOSE generates a database write (Notice events are persisted as state-change events in <observability.rs:347>-358), creating significant telemetry noise on the happy path.
	2.	OK messages are high-volume (one per published event per relay), so this pollutes the relay_events table with non-actionable data.
	3.	The "ok" and "eose" strings never match any failure classifier, so they all get RelayFailureCategory::Unknown as fallback, which is misleading in observability data.
Suggested fix: Either filter these message types out of the notification pipeline entirely, or introduce distinct RelayTelemetryKind variants for them with appropriate persistence rules (e.g., don’t persist OK/EOSE to relay_events at all).
Files: src/relay_control/sessions/<notifications.rs:62>-86, src/relay_control/sessions/<session.rs:866>-958, src/relay_control/<observability.rs:347>-358

mubarakcoded:
## Summary

- OK, EOSE, Count, NegMsg, and NegErr relay messages are now filtered out (`None`) instead of being incorrectly mapped to `RelayNotification::Notice`
- Only actionable messages (Notice, Closed, Auth) produce notifications

## Problem

These five message types were mapped to `Notice` with hardcoded strings like `"ok"` and `"eose"`, causing:
1. Unnecessary database writes on every OK response and EOSE (high-volume happy-path messages)
2. Pollution of the `relay_events` table with non-actionable data
3. Misleading `RelayFailureCategory::Unknown` entries since the hardcoded strings never match any failure classifier

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo clippy -- -D warnings`
- [x] Integration tests (`just precommit`)

Closes #606

## Summary by CodeRabbit

* **Refactor**
  * Reduced notification volume by filtering non-actionable relay messages so only critical relay notices (Notice, Closed, Auth) generate user-facing notifications, resulting in a cleaner, more relevant alert stream and fewer distractions.
* **Tests**
  * Added tests validating the new notification filtering to ensure benign, high-volume messages are no longer surfaced while important notices still trigger alerts.

### DIFF
--- a/src/relay_control/sessions/notifications.rs
+++ b/src/relay_control/sessions/notifications.rs
@@ -36,13 +36,15 @@ impl RelayNotification {
         relay_url: RelayUrl,
         message: RelayMessage<'static>,
     ) -> Option<Self> {
-        // This normalization is telemetry-oriented. Only actionable relay
-        // messages (Notice, Closed, Auth) produce notifications. High-volume
-        // happy-path messages (OK, EOSE, Count, NegMsg, NegErr) are filtered
-        // out to avoid polluting the relay_events table with non-actionable data.
+        // Only actionable relay messages (Notice, Closed, Auth) produce
+        // notifications. High-volume happy-path messages (OK, EOSE, Count,
+        // NegMsg, NegErr) are filtered out to avoid polluting the
+        // relay_events table with non-actionable data.
         match message {
-            RelayMessage::Event { .. }
-            | RelayMessage::Ok { .. }
+            // Events are handled by the RelayPoolNotification::Event branch
+            // in the pool notification loop, not through this path.
+            RelayMessage::Event { .. } => None,
+            RelayMessage::Ok { .. }
             | RelayMessage::EndOfStoredEvents(_)
             | RelayMessage::Count { .. }
             | RelayMessage::NegMsg { .. }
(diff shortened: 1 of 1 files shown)


src/relay_control/sessions/notifications.rs
```diff
@@ -59,31 +65,87 @@ impl RelayNotification {
...
+    }
+
+    #[test]
+    fn count_message_is_filtered() {
+        let result = RelayNotification::from_message(
+            test_relay_url(),
+            RelayMessage::count(SubscriptionId::new("sub1"), 42),
+        );
+        assert_eq!(result, None);
+    }
```
jgmontoya, line 108: `NegMsg` and `NegErr` are also filtered but not tested. Maybe we could collapse all of them to one test that does all?

src/relay_control/sessions/notifications.rs
```diff
@@ -36,11 +36,17 @@ impl RelayNotification {
         relay_url: RelayUrl,
         message: RelayMessage<'static>,
     ) -> Option<Self> {
-        // This normalization is telemetry-oriented. For message types such as
-        // `OK` and `EOSE` we intentionally collapse structured payloads into a
-        // coarse notification shape instead of preserving every field.
+        // This normalization is telemetry-oriented. Only actionable relay
```
jgmontoya: I would remove the "This normalization is telemetry-oriented." part as it's a bit confusing: we're no longer doing normalization, just dropping some cases.

src/relay_control/sessions/notifications.rs
```diff
@@ -36,11 +36,17 @@ impl RelayNotification {
...
+        // messages (Notice, Closed, Auth) produce notifications. High-volume
+        // happy-path messages (OK, EOSE, Count, NegMsg, NegErr) are filtered
+        // out to avoid polluting the relay_events table with non-actionable data.
         match message {
-            RelayMessage::Event { .. } => None,
+            RelayMessage::Event { .. }
```
jgmontoya: we could clarify perhaps that this one is filtered for a different reason (handled by the Event branch in the pool notification loop)

erskingardner (approved):

PR #646, opened 2026-03-16
fix: drain inbox planes in single lock then deactivate sequentially

### LINKED ISSUE #608: shutdown_all deactivates accounts sequentially
In <mod.rs:321>-332, shutdown_all iterates account pubkeys and calls deactivate_account_subscriptions sequentially. Each deactivation involves async unsubscribe operations and session shutdown. For many active accounts this could be slow.
These deactivations are independent of each other and could be parallelized.
Suggested fix: Use futures::future::join_all to deactivate all accounts concurrently.
Files: src/relay_control/<mod.rs:321>-332

mubarakcoded:
## Summary

- Drain all inbox planes in a single write-lock acquisition instead of acquiring the lock per account
- Deactivate planes sequentially after draining, following the same pattern as `reset_for_tests`

## Problem

`shutdown_all` called `deactivate_account_subscriptions` per account, each acquiring a write lock on `account_inbox_planes`. This caused repeated lock acquisition overhead.

## Approach

Drain the entire map in one lock, then iterate over the collected entries. Dropped `join_all` since `group_plane.remove_account` and `ephemeral.remove_account_scope` both acquire their own internal locks, so concurrent calls would contend rather than run in parallel.

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo clippy -- -D warnings`
- [x] Integration tests (`just precommit`)

Closes #608

## Summary by CodeRabbit

* **Chores**
  * Streamlined system shutdown to deactivate planes directly, improving reliability and making shutdowns faster and more deterministic.

* **Bug Fixes**
  * Made internal cleanup and reset routines always available, ensuring complete cleanup of ephemeral state and more consistent reset behavior during shutdown and tests.

### DIFF
--- a/src/relay_control/mod.rs
+++ b/src/relay_control/mod.rs
@@ -334,13 +334,23 @@ impl RelayControlPlane {
     /// Deactivates all account subscriptions. Called during full data teardown.
     #[perf_instrument("relay")]
     pub(crate) async fn shutdown_all(&self) {
-        let planes: Vec<_> = self.account_inbox_planes.write().await.drain().collect();
+        let planes: Vec<_> = self
+            .account_inbox_planes
+            .write()
+            .await
+            .drain()
+            .map(|(_, plane)| plane)
+            .collect();
 
-        for (pubkey, plane) in planes {
+        for plane in planes {
             plane.deactivate().await;
-            self.group_plane.remove_account(&pubkey).await;
-            self.ephemeral.remove_account_scope(&pubkey).await;
         }
+
+        // Reset group and ephemeral planes independently to ensure any
+        // orphaned entries (e.g. from partial activation failures) are
+        // cleaned up, not just accounts that had inbox planes.
+        self.group_plane.reset().await;
+        self.ephemeral.remove_all_account_scopes().await;
     }
 
     #[perf_instrument("relay")]
--- a/src/relay_control/ephemeral.rs
+++ b/src/relay_control/ephemeral.rs
@@ -110,7 +110,6 @@ impl EphemeralPlane {
         self.executor.remove_account_scope(account_pubkey).await;
     }
 
-    #[cfg(feature = "integration-tests")]
     #[perf_instrument("relay")]
     pub(crate) async fn remove_all_account_scopes(&self) {
         self.executor.remove_all_account_scopes().await;
--- a/src/relay_control/ephemeral_executor.rs
+++ b/src/relay_control/ephemeral_executor.rs
@@ -124,7 +124,6 @@ impl EphemeralExecutor {
         }
     }
 
-    #[cfg(feature = "integration-tests")]
     pub(crate) async fn remove_all_account_scopes(&self) {
         let _span = perf_span!("relay::ephemeral_remove_all_account_scopes");
         let account_keys = self
--- a/src/relay_control/groups.rs
+++ b/src/relay_control/groups.rs
@@ -368,7 +368,6 @@ impl GroupPlane {
         }
     }
 
-    #[cfg(feature = "integration-tests")]
     #[perf_instrument("relay")]
     pub(crate) async fn reset(&self) {
         let pubkeys = self


src/relay_control/mod.rs
```diff
@@ -334,15 +334,12 @@ impl RelayControlPlane {
...
-            .read()
-            .await
-            .keys()
-            .cloned()
-            .collect();
-        for pubkey in pubkeys {
-            self.deactivate_account_subscriptions(&pubkey).await;
+        let planes: Vec<_> = self.account_inbox_planes.write().await.drain().collect();
+
+        for (pubkey, plane) in planes {
```
erskingardner: The PR description explains why `join_all` was dropped (internal locks in `group_plane` and `ephemeral` would contend under concurrent calls), but this reasoning is not present in the code. Future maintainers looking at the sequential loop may reach for `join_all` as an "obvious optimization." A brief comment explaining why sequential is correct here would prevent regression:

```rust
// Sequential, not concurrent: group_plane.remove_account and
// ephemeral.remove_account_scope each acquire their own internal
// write locks, so concurrent calls would contend without benefit.
for (pubkey, plane) in planes {
```

src/relay_control/mod.rs
```diff
@@ -334,15 +334,12 @@ impl RelayControlPlane {
...
-        let pubkeys: Vec<_> = self
-            .account_inbox_planes
-            .read()
-            .await
-            .keys()
-            .cloned()
-            .collect();
-        for pubkey in pubkeys {
-            self.deactivate_account_subscriptions(&pubkey).await;
+        let planes: Vec<_> = self.account_inbox_planes.write().await.drain().collect();
```
erskingardner: Agent came back with this and it's got a good point. 

We should make sure that however we're shutting down/cleaning up all these planes and subs, it needs to be orderly and atomic across all the plans for a given account. 

> The new implementation uses `account_inbox_planes` as the sole source of truth for which accounts need cleanup. But `group_plane` and `ephemeral` can have accounts that are not in `account_inbox_planes`:
> 
> - **Partial activation failure path in `activate_account_subscriptions`**: The docstring acknowledges activation is not atomic. If `group_plane.update_account` succeeds but inbox `plane.activate` subsequently fails on the _first_ activation attempt, the error path cleans up group_plane. However, re-entrant calls to `activate_account_subscriptions` call `deactivate_account_subscriptions` first (atomically removing the inbox entry), then attempt a fresh activation. If inbox activation fails on _re-entry_, the group plane is updated but inbox plane is not inserted — leaving the group_plane entry orphaned.
> - **`sync_account_group_subscriptions`** (`src/whitenoise/accounts/setup.rs:703`) is called from the background refresh path and updates `group_plane` for an account. If this runs concurrently with shutdown and completes after the drain, the group_plane entry will be missed entirely.
> 
> The old code called `deactivate_account_subscriptions` per pubkey, which unconditionally calls both `group_plane.remove_account` and `ephemeral.remove_account_scope` regardless of whether an inbox plane entry was found. The new code ties those cleanup calls directly to inbox plane membership.
> 
> **Impact**: Orphaned group_plane subscriptions could remain active post-shutdown, potentially continuing to receive relay events into a torn-down system. For the `delete_all_data` context where `shutdown_all` is called, this is more of a resource leak than a correctness hazard — but it diverges from the previous semantics.
> 
> **Recommendation**: Consider whether `shutdown_all` should instead call the public `group_plane.reset()` and `ephemeral.remove_all_account_scopes()` equivalents (the pattern used in `reset_for_tests`) after draining inbox planes, rather than iterating per-account from the inbox plane's key set. That would cleanly guarantee all three planes are fully drained:

Another thing that the agent said aligns with this: 

> ### Consider extracting a private helper to reduce the duplication between `deactivate_account_subscriptions` and `shutdown_all`

> 
> **`src/relay_control/mod.rs:320–346`**
> 
> Both methods share the same three-step sequence: `plane.deactivate()`, `group_plane.remove_account()`, `ephemeral.remove_account_scope()`. If the invariant holds (all accounts in group/ephemeral are always in inbox), a small private helper that does the three cleanup steps given a pubkey and plane would reduce drift risk. Not blocking, but reduces copy/paste maintenance surface.

PR #647, opened 2026-03-17
pagination in streams too

dannym-arx:
marmot

## What

Brings pagination into the message streaming path, so the Flutter side can request a partial initial load and fetch older pages on demand (exactly the pattern a chat app needs for infinite scroll).

Three connected changes:

**Bug fix:** `find_messages_by_group_paginated` was missing `AND am.deletion_event_id IS NULL`. Deleted messages were leaking through into paginated results. Fixed and covered by a new test.

**`subscribe_to_group_messages` gets a `limit` parameter.** The initial snapshot now uses the same paginated query as `fetch_aggregated_messages_for_group`, so the Flutter side opens a chat with at most N messages and fetches older ones on demand via the existing pagination API. Without this, the initial load always returns the full history, defeating the point of pagination.

**Wire protocol updated.** `MessagesSubscribe` gains an optional `limit` field (defaults to `None`, which gives 50, via `#[serde(default)]`). Old clients that don't send the field still work. The CLI `messages subscribe` command gets `--limit` too.

## Tests added

**Pagination DB layer** (8 new tests on `find_messages_by_group_paginated`):
- Group isolation: messages from group A don't bleed into group B
- Deleted messages excluded: the bug fix is regression-locked
- Non-kind-9 rows excluded: kind-7 reactions don't appear
- Uppercase cursor ID canonicalized: `ABCD...` and `abcd...` produce the same page
- Single-message group: works at the minimum
- Exactly 50 messages: boundary case for the default limit
- `limit=1` traversal: steps through every message one at a time, no skips or duplicates
- Cursor at the oldest message: strict `<` boundary returns empty

**Stream subscription** (14 new tests on `subscribe_to_group_messages`):
- `limit` caps the snapshot to N newest
- `limit` larger than history returns everything
- `None` defaults to 50, caps at 50 when more exist
- Snapshot is always oldest-first
- Drain loop exits cleanly on empty channel
- Deduplication by message ID (no double copies)
- Updates receiver delivers future messages
- Updates receiver doesn't replay the snapshot
- Empty group with limit: receiver still works
- Multiple subscribers on the same group get the same updates (broadcast fan-out)
- Updates isolated per group

**Protocol** (3 new tests):
- `limit` round-trips correctly
- Missing `limit` field in JSON deserialises as `None` (backward compat)
- Explicit `null` deserialises as `None`

**Scroll-up and live message scenarios** (4 end-to-end tests):
- User opens chat, scrolls back through history pages, receives new messages during scrolling. New messages land at the bottom via the updates receiver, not in any historical page.
- History exhausts cleanly on the third page
- Live updates arrive regardless of how many history pages have been fetched
- Tied timestamps at a page boundary produce no overlap or gaps

sorry, pepi

## Summary by CodeRabbit

* **New Features**
  * Added a CLI --limit option for message subscriptions to control initial snapshot size (default 50, capped at 200)
  * Subscription API accepts an optional limit and returns paginated initial snapshots

* **Bug Fixes**
  * Exclude deleted messages from initial snapshots
  * Deterministic oldest-first ordering, stronger deduplication, and fail-fast on receiver lag during subscription init

* **Tests**
  * Expanded tests for pagination, serialization/backward-compatibility, ordering, deduplication, streaming, fan-out, and per-group isolation

### DIFF
--- a/src/integration_tests/test_cases/message_streaming/verify_initial_messages.rs
+++ b/src/integration_tests/test_cases/message_streaming/verify_initial_messages.rs
@@ -132,27 +132,32 @@ impl TestCase for VerifyInitialMessagesTestCase {
             tracing::info!("✓ Message '{}' correctly has no reactions", key);
         }
 
-        // Verify deleted messages are NOT in the initial snapshot
+        // Verify deleted messages ARE in the initial snapshot as tombstones (is_deleted = true)
         for key in &self.expected_deleted {
             let msg_id = context.get_message_id(key)?;
-            let found = subscription
+            let msg = subscription
                 .initial_messages
                 .iter()
-                .any(|m| &m.id == msg_id);
+                .find(|m| &m.id == msg_id);
 
             assert!(
-                !found,
-                "Expected message '{}' to be excluded from initial snapshot because it is deleted, but it was found",
+                msg.is_some(),
+                "Expected deleted message '{}' to appear in initial snapshot as a tombstone, but it was absent",
+                key
+            );
+            assert!(
+                msg.unwrap().is_deleted,
+                "Expected message '{}' to have is_deleted = true, but it was false",
                 key
             );
             tracing::info!(
-                "✓ Message '{}' is correctly excluded from the initial snapshot",
+                "✓ Message '{}' is correctly present in the initial snapshot as a tombstone",
                 key
             );
         }
 
         tracing::info!(
-            "✓ Verified {} initial messages ({} with reactions, {} without reactions, {} deleted)",
+            "✓ Verified {} initial messages ({} with reactions, {} without reactions, {} deleted tombstones)",
             subscription.initial_messages.len(),
             self.expected_with_reactions.len(),
             self.expected_no_reactions.len(),


src/whitenoise/streaming.rs
```diff
@@ -30,51 +30,90 @@ impl Whitenoise {
...
+        //    on the initial snapshot.
         loop {
             match updates.try_recv() {
                 Ok(update) => {
-                    // Apply update: insert or replace by message ID
-                    messages_map.insert(update.message.id.clone(), update.message);
+                    if update.trigger == message_streaming::UpdateTrigger::NewMessage
+                        || messages_map.contains_key(&update.message.id)
+                    {
+                        messages_map.insert(update.message.id.clone(), update.message);
```
erskingardner, line 85: I _believe_ this is means that if you're 3 pages back in messages and a new message arrives it's going to immediately insert itself in the list? I can see how that might be intentional but it feels like it would dependent on how the UI works? 

> When `NewMessage` updates arrive in the drain window they are unconditionally inserted into
> the map. This is correct for messages whose ID is already in the map (replacing a stale DB
> row). But for genuinely new messages that arrived *just after* the SQL cut-off, every such
> message is added regardless of the requested `limit`. In a group receiving a burst of
> messages the snapshot handed back to Flutter can silently exceed the limit that was
> requested — breaking the invariant Flutter relies on to know where to place the cursor for
> the first scroll-up fetch.
> 
> The fix is to track whether each drain-window `NewMessage` is actually newer than the oldest
> message in the current map (i.e. it truly belongs in this page), or to simply apply the
> limit as a post-step trim on `initial_messages` before returning.

src/whitenoise/database/aggregated_messages.rs
```diff
@@ -326,6 +326,7 @@ impl AggregatedMessage {
                ON am.message_id = mds.message_id AND am.mls_group_id = mds.mls_group_id
              WHERE am.kind = 9 AND am.mls_group_id = ?
                AND (am.created_at < ? OR (am.created_at = ? AND am.message_id < ?))
+               AND am.deletion_event_id IS NULL
```
erskingardner: Don't we still show the deleted messages in the aggregated list so that we can show "Message delete" in the right spot? 

Either way - we should match on how we're filtering between the paginated endpoint and the plain endpoint

src/whitenoise/streaming.rs
dannym-arx: I'm unsure of what to do. 

I could just ignore new messages but then you wouldn't see new messages; I could keep the new ones in a queue, until flutter sends an event that the user is at the bottom or something? or we could keep it as is and let flutter handle it

I don't like any of those options to be honest haha

Issue #607, opened 2026-03-10
Group plane tears down subscriptions before establishing new ones


erskingardner:
In <groups.rs:89>-95, update_account unconditionally unsubscribes all existing subscriptions for an account before establishing new ones. This creates a coverage gap where group messages arriving between teardown and setup are missed.
Compare this to the discovery plane (<discovery.rs:142>-146) which builds all new subscriptions first, then retires stale ones — a much better pattern that avoids coverage gaps.
Suggested fix: The group plane should subscribe with the new spec first (using the new subscription IDs), then unsubscribe the old subscription IDs that weren’t reused. This matches the discovery plane’s approach.
Files: src/relay_control/<groups.rs:89>-95, src/relay_control/<discovery.rs:142>-146 (reference for correct pattern)

PR #648, opened 2026-03-17
Fixing preventing coverage gaps during account resubscribe

sergey3bv:
Addressing #607 

## Summary by CodeRabbit

* **Bug Fixes**
  * More resilient group subscription updates with targeted cleanup, improved cancellation handling, and ensuring relay records exist before syncing.

* **Behavior Changes**
  * Larger resubscribe lookback to cover teardown gaps; initial activation/lookback adjustments to reduce unnecessary refetches.

* **Refactor**
  * Centralized filter construction for consistent, idempotent subscription updates.

* **Tests**
  * Expanded tests for filter serialization, subscription/state behaviors, and cancellation cases.

### DIFF
--- a/src/relay_control/groups.rs
+++ b/src/relay_control/groups.rs
@@ -111,7 +111,7 @@ impl GroupPlane {
         let normalized_groups = Self::normalize_group_specs(group_specs);
         let subscriptions = Self::build_relay_set_subscriptions(&normalized_groups);
 
-        // On any failure below, remove the accounts entry so that
+        // On any failure below, remove the account entry so that
         // has_active_subscription() returns false and recovery is triggered.
         // (The unsubscribe above already tore down the previous subscription,
         // so leaving stale relay info in the map would make the health check
@@ -120,15 +120,6 @@ impl GroupPlane {
         for subscription in &subscriptions {
             if subscription.relays.is_empty() || subscription.group_ids.is_empty() {
                 continue;
-            }
-
-            let mut filter = Filter::new().kind(Kind::MlsGroupMessage).custom_tags(
-                SingleLetterTag::lowercase(Alphabet::H),
-                subscription.group_ids.iter(),
-            );
-
-            if let Some(since) = since {
-                filter = filter.since(since);
             }
 
             if let Err(e) = self
@@ -140,6 +131,13 @@ impl GroupPlane {
                     .await;
                 self.accounts.write().await.remove(&pubkey);
                 return Err(e);
+            }
+            let mut filter = Filter::new().kind(Kind::MlsGroupMessage).custom_tags(
+                SingleLetterTag::lowercase(Alphabet::H),
+                subscription.group_ids.iter(),
+            );
+            if let Some(since) = since {
+                filter = filter.since(since);
             }
             if let Err(e) = self
                 .session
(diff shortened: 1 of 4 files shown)


src/whitenoise/mod.rs
```diff
@@ -570,6 +570,111 @@ impl Whitenoise {
...
+            Self::setup_global_users_subscriptions(whitenoise_ref),
+            Self::setup_accounts_subscriptions(whitenoise_ref),
+        );
+        global_result?;
+        accounts_result?;
+        Ok(())
+    }
+
+    #[perf_instrument("whitenoise")]
+    async fn setup_global_users_subscriptions(whitenoise_ref: &Whitenoise) -> Result<()> {
```
erskingardner: I think the merge with master messed up something here. These methods and constants are now in the `subscriptions.rs` file since the refactor we merged earlier today.

src/whitenoise/mod.rs
```diff
@@ -570,6 +570,111 @@ impl Whitenoise {
...
+        for account in accounts {
+            let inbox_relays = account.effective_inbox_relays(whitenoise_ref).await?;
+            // Setup subscriptions for this account
+            match whitenoise_ref
+                .setup_subscriptions(&account, &inbox_relays)
+                .await
+            {
+                Ok(()) => {
+                    tracing::debug!(
+                        target: "whitenoise::initialize_whitenoise",
```
erskingardner: we should use `whitenoise::setup_accounts_subscriptions` as the log target

src/whitenoise/mod.rs
```diff
@@ -570,6 +570,111 @@ impl Whitenoise {
...
+                        "Successfully set up subscriptions for account: {}",
+                        account.pubkey.to_hex()
+                    );
+                }
+                Err(e) => {
+                    tracing::warn!(
+                        target: "whitenoise::initialize_whitenoise",
```
erskingardner: we should use `whitenoise::setup_accounts_subscriptions` as the log target

src/relay_control/groups.rs
```diff
@@ -251,6 +242,19 @@ impl GroupPlane {
         ))
     }
 
+    fn build_filter(subscription: &GroupRelaySetSubscription, since: Option<Timestamp>) -> Filter {
```
erskingardner: This doesn't feel like a valuable refactor here - we're not reusing that filter anywhere else.

src/whitenoise/accounts/setup.rs
```diff
@@ -756,7 +756,29 @@ impl Whitenoise {
...
+                "Skipping group subscription sync because the account was cancelled",
+            );
+            return Ok(());
+        }
+
+        self.relay_control
+            .sync_account_group_subscriptions(
+                account.pubkey,
+                &group_specs,
+                account.since_timestamp(Self::RESUBSCRIBE_BUFFER_SECS),
```
erskingardner: This change looks correct for the background refresh path: it now uses `RESUBSCRIBE_BUFFER_SECS` directly when rebuilding group subscriptions. My only concern is consistency: `sync_group_subscriptions()` (above) still uses the older `10s` lookback, so the wider overlap window is not applied uniformly across all group-subscription rebuild paths yet. If the goal is to avoid message gaps any time we tear down and rebuild these subscriptions, it may be worth centralizing that lookback behavior so every rebuild entry point uses the same buffer.

PR #654, opened 2026-03-17
Perf/login

jgmontoya:
## Summary by CodeRabbit

* **Tests**
  * Added two new login performance benchmark scenarios; registry and parsing updated (case-insensitive lookup) and test counts adjusted.

* **Refactor**
  * Made relay publishing and account activation steps concurrent; introduced background key-package publish and improved error handling.
  * Expanded performance instrumentation across login flows for finer-grained tracing.

### DIFF
--- a/src/whitenoise/accounts/login.rs
+++ b/src/whitenoise/accounts/login.rs
@@ -313,35 +313,18 @@ impl Whitenoise {
             .relays_or(RelayType::Nip65, &default_relays)
             .to_vec();
 
-        // For each relay type: publish defaults only for the ones that were
-        // missing. Already-found lists were already persisted to the DB by
-        // try_discover_relay_lists in step 1, so no second write is needed.
+        // Persist missing relay lists to the local database, then publish
+        // them to the network concurrently. DB writes are fast (sub-ms) and
+        // use the same connection, so they run sequentially. The publishes
+        // are independent Nostr events (kinds 10002, 10050, 10051) targeting
+        // the same relay set, so they run concurrently — same pattern as
+        // setup_relays_for_existing_account.
+        let _publish_span = perf_span!("accounts::login_publish_defaults::publish_missing");
         for relay_type in [RelayType::Nip65, RelayType::Inbox, RelayType::KeyPackage] {
             if !discovered.found(relay_type) {
-                // Missing — assign defaults in the DB and publish to the network.
                 user.add_relays(&default_relays, relay_type, &self.database)
                     .await
                     .map_err(LoginError::from)?;
-                if let Err(error) = self
-                    .publish_relay_list(
-                        &default_relays,
-                        relay_type,
-                        &publish_to_relays,
-                        keys.clone(),
-                    )
-                    .await
-                {
-                    if discovered.nip65.is_none() || relay_type == RelayType::Nip65 {
-                        return Err(LoginError::from(error));
-                    }
-
-                    tracing::warn!(
-                        target: "whitenoise::accounts",
-                        pubkey = %pubkey,
-                        ?relay_type,
-                        "Failed to publish default relay list to preserved NIP-65 relays; continuing login with local relay state: {error}"
-                    );
-                }
             } else {
                 tracing::debug!(
                     target: "whitenoise::accounts",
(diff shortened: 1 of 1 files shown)


src/whitenoise/accounts/login.rs
```diff
@@ -348,6 +334,73 @@ impl Whitenoise {
...
+            if let Err(error) = result {
+                if discovered.nip65.is_none() {
+                    return Err(LoginError::from(error));
+                }
+                tracing::warn!(
+                    target: "whitenoise::accounts",
+                    pubkey = %pubkey,
+                    ?relay_type,
+                    "Failed to publish default relay list to preserved NIP-65 relays; continuing login with local relay state: {error}"
+                );
```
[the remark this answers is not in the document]
jgmontoya: Good catch. discovered is cloned from step 1 state and never updated after the publish, so discovered.nip65.is_none() stays true for fresh accounts even after a successful NIP-65 publish, making inbox/key-package failures incorrectly fatal.                                             
                                                                                                     
The suggested fix of gating on nip65_result.is_err() can't quite work because nip65_result is consumed by the map_err()? on line 381. But it's also unnecessary: if we get past that line, NIP-65 is guaranteed to be on the network (either discovered in step 1 or just published). The guard can never correctly fire at that point, so I removed it entirely along with the now-stale comment.

src/whitenoise/accounts/login.rs
```diff
@@ -1055,50 +1116,67 @@ impl Whitenoise {
...
-                    RelayType::KeyPackage,
-                    &nip65_urls,
-                    signer.clone(),
-                )
-                .await?;
-        }
+        let (r1, r2, r3) = tokio::join!(nip65_publish, inbox_publish, key_package_publish);
+        r1?;
+        r2?;
+        r3?;
```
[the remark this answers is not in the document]
jgmontoya: Good catch — same issue as the other path. After r1?, NIP-65 is guaranteed on the network, so
  inbox/key-package failures shouldn't abort login. It's also worse here because                     
  activate_account_without_publishing runs before the publish, so a fatal r2/r3 returns an error for
  an already-activated account. Replaced r2?; r3?; with the same warn-and-continue loop used in the  
  multi-step path.                                                                                 
                  
  Worth noting: there's no retry mechanism for failed relay list publishes today (re-login
  short-circuits for existing accounts). I'll create an issue for a relay_list_maintenance scheduled 
  task to self-heal this, following the KeyPackageMaintenance pattern.

src/whitenoise/accounts/login.rs
```diff
@@ -348,6 +334,68 @@ impl Whitenoise {
...
+            } else {
+                Ok(())
+            }
+        };
+
+        let (nip65_result, inbox_result, kp_result) =
+            tokio::join!(nip65_publish, inbox_publish, key_package_publish);
+
+        // NIP-65 publish failure is always fatal.
+        nip65_result.map_err(LoginError::from)?;
```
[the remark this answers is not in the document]
jgmontoya: Keeping the concurrent design. The happy path (all three succeed) is the overwhelmingly common
  case, and tokio::join! gives us max(t1, t2, t3) vs t1 + max(t2, t3) — each publish involves relay  
  connections and network round-trips, so that's a meaningful difference on a perf branch.         
                                                                                                     
  The failure scenario (NIP-65 fails, 10050/10051 succeed) is harmless in practice: without NIP-65,
  nobody discovers the user, so orphaned 10050/10051 events are inert. They're replaceable events    
  that get overwritten on the next successful publish.

src/integration_tests/benchmarks/scenarios/login_multistep_performance.rs
```diff
@@ -0,0 +1,84 @@
...
+            warmup_iterations: 1,
+            cooldown_between_iterations: Duration::from_millis(500),
+        }
+    }
+
+    async fn setup(&mut self, _context: &mut ScenarioContext) -> Result<(), WhitenoiseError> {
+        let config = self.config();
+        let total_iterations = (config.iterations + config.warmup_iterations) as usize;
+
+        tracing::info!(
```
mubarakcoded, line 55: This needs target: "whitenoise::benchmarks" right?

src/integration_tests/benchmarks/scenarios/login_multistep_performance.rs
```diff
@@ -0,0 +1,84 @@
...
+            "Setting up login-multistep benchmark: {} iterations (no relay lists to publish)",
+            total_iterations
+        );
+
+        // Generate keypairs only — no relay lists published so login_start
+        // returns NeedsRelayLists for each.
+        let prepared_keys: Vec<Keys> = (0..total_iterations).map(|_| Keys::generate()).collect();
+
+        tracing::info!(
```
mubarakcoded, line 64: same here

src/integration_tests/benchmarks/scenarios/login_start_performance.rs
```diff
@@ -0,0 +1,111 @@
...
+            cooldown_between_iterations: Duration::from_millis(500),
+        }
+    }
+
+    async fn setup(&mut self, context: &mut ScenarioContext) -> Result<(), WhitenoiseError> {
+        let config = self.config();
+        let total_iterations = (config.iterations + config.warmup_iterations) as usize;
+        let relay_urls: Vec<String> = context.dev_relays.iter().map(|s| s.to_string()).collect();
+
+        tracing::info!(
```
mubarakcoded, line 59: same here

src/integration_tests/benchmarks/scenarios/login_start_performance.rs
```diff
@@ -0,0 +1,111 @@
...
+            .await?;
+
+            // Publish all 3 relay lists so login_start finds them
+            publish_relay_lists(&test_client, relay_urls.clone()).await?;
+
+            test_client.disconnect().await;
+            prepared_keys.push(keys);
+
+            if (i + 1) % 5 == 0 || i + 1 == total_iterations {
+                tracing::info!("Prepared {}/{} test accounts", i + 1, total_iterations);
```
mubarakcoded, line 84: and here, this one fires every 5 iterations so worth getting the target right

src/integration_tests/benchmarks/scenarios/login_start_performance.rs
```diff
@@ -0,0 +1,111 @@
...
+            }
+        }
+
+        // Brief pause to let relays index the published events
+        tokio::time::sleep(Duration::from_secs(2)).await;
+
+        tracing::info!(
```
mubarakcoded, line 91: and here too

PR #666, opened 2026-03-18
perf: remove unnecessary side effects from new-chat user resolution

### LINKED ISSUE #660: Remove unnecessary side effects and extra work from new-chat user resolution
## Summary
Cold new-chat user resolution appears to do extra work on the critical path that is not always required for starting a DM.

Two likely contributors:
- refreshing global discovery subscriptions after one-off user resolution
- syncing metadata before key-package fetch in member resolution even when metadata is not required to create the MLS group

## Relevant code paths
- `src/whitenoise/users.rs` (`sync_user_blocking`, `background_fetch_user_data`)
- `src/whitenoise/mod.rs` (`refresh_global_subscription_for_user` comment notes full replace / not cheap)
- `src/whitenoise/groups.rs` (`resolve_member_key_package`)

## Expected outcome
- Identify and remove non-essential side effects from the critical path for first-time chat starts.
- Avoid global subscription refresh for one-off user discovery unless it is truly required.
- Avoid blocking key-package resolution on metadata sync when the caller only needs the key package.

dannym-arx:
marmot

Remove unnecessary side effects from new-chat user resolution

Three classes of redundant work were blocking the critical path for first-time DM starts.

**Global subscription refresh removed from one-off user lookups.** \`sync_user_blocking\` and \`background_fetch_user_data\` both called \`refresh_global_subscription_for_user\` after resolving a single user. That function does a full replace of every watched-user and follow-list subscription batch (not cheap, not warranted for a one-off lookup). The refresh still fires at the right moments: account login, and when a followed user's relay list actually changes.

**Metadata sync removed from \`resolve_member_key_package\`.** Creating an MLS group requires a key package, not a display name. The prior code fetched metadata for any newly seen member before fetching the key package, adding a network round-trip that contributed nothing to group creation.

**\`add_members_to_group\` now syncs relay lists synchronously for new users.** The prior code called \`background_fetch_user_data\` (async) then immediately read the user's key package relays: a race it always lost. Changed to a synchronous \`update_relay_lists\` call so relay data is available before the key package fetch. Metadata sync is omitted here too since it's not needed to add a member.

Closes #660

## Summary by CodeRabbit

* **Refactor**
  * Simplified new-user and group-member setup to avoid mutating users during creation and reduce redundant background synchronization.
  * Switched some background metadata syncs to a synchronous relay-list update and emit a warning if that update fails.
  * Background user-fetch now treats newly created users differently, preventing automatic global-subscription refreshes during initial setup to reduce noisy refreshes.

### DIFF
--- a/src/whitenoise/follows.rs
+++ b/src/whitenoise/follows.rs
@@ -24,7 +24,7 @@ impl Whitenoise {
         let (user, newly_created) = User::find_or_create_by_pubkey(pubkey, &self.database).await?;
 
         if newly_created {
-            self.background_fetch_user_data(&user).await?;
+            self.background_fetch_user_data(&user, true).await?;
         }
 
         account.follow_user(&user, &self.database).await?;
--- a/src/whitenoise/users.rs
+++ b/src/whitenoise/users.rs
@@ -372,6 +372,16 @@ impl Whitenoise {
                     e
                 );
             }
+            // Add the new user to the discovery subscription batch so future subscription-driven
+            // updates (metadata changes, relay list changes) are delivered automatically.
+            if let Err(e) = self.refresh_global_subscription_for_user().await {
+                tracing::warn!(
+                    target: "whitenoise::users::sync_user_blocking",
+                    "Failed to refresh global subscription for new user {}: {}",
+                    user_clone.pubkey,
+                    e
+                );
+            }
         }
 
         if let Err(e) = user_clone.sync_metadata(self).await {
@@ -388,7 +398,7 @@ impl Whitenoise {
 
     async fn sync_user_background(&self, user: &User, is_new: bool) -> Result<()> {
         if is_new {
-            if let Err(e) = self.background_fetch_user_data(user).await {
+            if let Err(e) = self.background_fetch_user_data(user, true).await {
                 tracing::warn!(
                     target: "whitenoise::users::sync_user_background",
                     "Failed to start background fetch for new user {}: {}",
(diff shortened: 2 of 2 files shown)


src/whitenoise/groups.rs
```diff
@@ -89,29 +89,19 @@ impl Whitenoise {
...
-                    target: "whitenoise::accounts::groups::create_group",
-                    "Failed to sync metadata for new user {}: {}",
-                    user.pubkey,
-                    e
-                );
-            }
+        let (user, created) = User::find_or_create_by_pubkey(pk, &self.database).await?;
+        if created && let Err(e) = user.update_relay_lists(self).await {
+            tracing::warn!(
+                target: "whitenoise::accounts::groups::create_group",
```
jgmontoya, line 100: Got this one from my agent:

> After this change, a user created for the first time through `resolve_member_key_package` has their relay lists fetched but is never enrolled in the discovery subscription batch (`refresh_global_subscription_for_user`). The PR description says the refresh "still fires at the right moments: account login, and when a followed user's relay list actually changes" — but `resolve_member_key_package` is neither of those. A user you add to a group is not necessarily a follow, and the relay-change handler only fires for users already subscribed.
> 
> This means the client won't receive future metadata updates or relay list changes for users who are first encountered as MLS group members, not as follows.
> 
> The pre-PR code routed through `background_fetch_user_data`, which always called `refresh_global_subscription_for_user`. The new `add_members_to_group` path (correctly) fixes the race by calling `update_relay_lists` synchronously — but it also needs to enqueue `refresh_global_subscription_for_user` afterwards (it can be async / fire-and-forget). The `resolve_member_key_package` path has the same gap.

dannym-arx: ... too late merged already, but @coderabbitai please open an issue for this, so I'll solve it

PR #675, opened 2026-03-19
Make user metadata state explicit

erskingardner:
## Summary
- add `metadata_known_at` to `users`, backfill clearly non-empty legacy metadata as known, and plumb the new state through the DB and `User` model
- keep `find_or_create_user_by_pubkey(..., UserSyncMode)` intact while switching internal metadata resolution from TTL/`updated_at` refreshes to explicit known-vs-unknown behavior
- stop targeted discovery from writing `users.metadata` directly by routing discovery catch-up through the guarded event-processing path, and make metadata fetch fail closed on invalid kind-0 events
- update unit/integration coverage for blank metadata, stale/older relay results, and unknown metadata behavior; include the small supporting test/bootstrap fixes needed to keep the suite green

## Testing
- `just precommit-quick`

## Summary by CodeRabbit

* **New Features**
  * Persist a per-user "metadata known" state and surface it across user records and sync flows
  * Introduce bounded targeted discovery to fetch metadata/relay info when metadata is unknown
  * Expose test helper to observe processing events in tests

* **Bug Fixes**
  * Prefer semantically valid metadata events; never fall back to invalid ones
  * Prevent older discovery results from overwriting newer processed metadata
  * Ensure metadata is marked known when observed or accepted

* **Tests**
  * Added/updated tests for discovery, metadata handling, uploads, and CLI key-package waiting

* **Chores**
  * Database migration adding nullable metadata_known_at column

### DIFF
--- a/src/integration_tests/test_cases/user_discovery/find_or_create_user_preserves_newer_processed_metadata.rs
+++ b/src/integration_tests/test_cases/user_discovery/find_or_create_user_preserves_newer_processed_metadata.rs
@@ -1,8 +1,47 @@
+use std::time::Duration;
+
+use async_trait::async_trait;
+use nostr_sdk::{Client, EventBuilder, EventId, Filter, Keys, Kind, Metadata, Timestamp};
+
 use crate::WhitenoiseError;
 use crate::integration_tests::core::test_clients::create_test_client;
 use crate::integration_tests::core::*;
-use async_trait::async_trait;
-use nostr_sdk::{EventBuilder, Keys, Metadata, Timestamp};
+
+const LOG_TARGET: &str = "integration_tests::test_cases::user_discovery::find_or_create_user_preserves_newer_processed_metadata";
+
+async fn wait_for_latest_metadata_event(
+    client: &Client,
+    pubkey: nostr_sdk::PublicKey,
+    expected_event_id: EventId,
+    description: &str,
+) -> Result<(), WhitenoiseError> {
+    // Kind-0 is replaceable, so older events are not guaranteed to remain
+    // queryable by exact ID once the relay has reconciled them. Poll the same
+    // author+kind lookup shape used by targeted discovery instead.
+    retry(
+        30,
+        Duration::from_millis(100),
+        || async {
+            let events = client
+                .fetch_events(
+                    Filter::new().author(pubkey).kind(Kind::Metadata),
+                    Duration::from_secs(1),
+                )
+                .await?;
+
+            if events.iter().any(|event| event.id == expected_event_id) {
+                Ok(())
+            } else {
+                Err(WhitenoiseError::Other(anyhow::anyhow!(
+                    "Latest metadata query does not yet return expected event {}",
+                    expected_event_id.to_hex()
+                )))
+            }
+        },
+        description,
+    )
+    .await
+}
 
 /// Tests that targeted discovery cannot overwrite newer already-processed metadata.
 ///
@@ -42,6 +81,7 @@ impl TestCase for FindOrCreateUserPreservesNewerProcessedMetadataTestCase {
     async fn run(&self, context: &mut ScenarioContext) -> Result<(), WhitenoiseError> {
         let test_pubkey = self.test_keys.public_key();
         tracing::info!(
+            target: LOG_TARGET,
             "Testing targeted discovery ordering for legacy unknown metadata on pubkey: {}",
             test_pubkey
         );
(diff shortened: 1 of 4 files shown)


src/integration_tests/test_cases/user_discovery/find_or_create_user_unknown_metadata_no_result.rs
jgmontoya, line 1: We should update the file name

src/integration_tests/test_cases/user_discovery/find_or_create_user_preserves_newer_processed_metadata.rs
jgmontoya, line 1: we should update this file name as well

src/whitenoise/users/key_package.rs
jgmontoya, line 1: I think the change on this file is completely unnecessary tbh, it would only make a difference if the app is explicitly built with an empty set of relays on the discovery plane, I'd assume if someone does that it's gonna be broken regardless

src/whitenoise/users.rs
```diff
@@ -340,182 +409,160 @@ impl Whitenoise {
...
+    pub(crate) async fn background_fetch_user_data(&self, user: &User, is_new: bool) -> Result<()> {
+        #[cfg(test)]
+        {
+            self.background_fetch_user_data_inner(user.clone(), is_new)
+                .await?;
+            Ok(())
+        }
+
+        #[cfg(not(test))]
+        {
```
jgmontoya: I don't like this one bit

src/whitenoise/users.rs
```diff
@@ -340,182 +409,160 @@ impl Whitenoise {
...
-                tracing::warn!(
-                    target: "whitenoise::users::sync_user_background",
-                    "Failed to start background fetch for new user {}: {}",
-                    user.pubkey,
-                    e
-                );
-            }
-        } else if user.needs_metadata_refresh() {
-            // For existing users, only sync if metadata is stale
+        if user.metadata_is_unknown() {
```
jgmontoya, line 461: Previously, when `is_new` was `true`, this method always called `background_fetch_user_data`, which for new users would call `refresh_global_subscription_for_user`.
                                                                                                                                                                                                             
Now, `refresh_global_subscription_for_user` only runs if `metadata_is_unknown()` is also `true`. In practice this is fine today, a newly created user always has `metadata_known_at: None`, so the condition is always met when `is_new` is `true`.                                                                                                                                                                            
                            
But it's a bit uncomfortable that subscription setup for new users is gated behind a metadata resolution check. Those are two independent concerns: "this user needs to be added to discovery subscriptions" vs "this user's metadata hasn't been resolved yet." Coupling them means that if anything ever changes how `metadata_known_at` gets set during creation, we'd silently stop subscribing to new users.

src/whitenoise/event_processor/event_handlers/handle_metadata.rs
```diff
@@ -49,3 +59,133 @@ impl Whitenoise {
...
+        let user = whitenoise
+            .find_user_by_pubkey(&keys.public_key())
+            .await
+            .unwrap();
+        assert_eq!(user.metadata.name, Some("newer".to_string()));
+        assert_eq!(user.metadata_known_at, first_known_at);
+    }
+
+    #[tokio::test]
+    async fn already_current_metadata_marks_unknown_user_as_known() {
```
jgmontoya: Name is a bit confusing, how about `already_processed_event_still_marks_unknown_user_as_known` ?

src/whitenoise/users.rs
```diff
@@ -274,11 +226,128 @@ impl Whitenoise {
...
+                }
+                Kind::Metadata
+                | Kind::RelayList
+                | Kind::InboxRelays
+                | Kind::MlsKeyPackageRelays => Some(event),
+                _ => None,
+            })
+            .collect();
+
+        filtered_events.sort_unstable_by_key(|event| (event.created_at, event.id));
```
jgmontoya: is this sort needed?

src/whitenoise/users.rs
```diff
@@ -57,80 +65,22 @@ pub struct User {
...
-                event.created_at
-            );
-        } else {
-            // No metadata found — record that we checked so TTL is respected
-            // for empty-profile users, preventing repeated fruitless fetches.
-            self.touch_updated_at(&whitenoise.database).await?;
-        }
+    pub fn metadata_is_unknown(&self) -> bool {
+        self.metadata_known_at.is_none()
+    }
```
jgmontoya, line 80: Why not just use `!user.metadata_is_known()` ?

Not that it really matters much though

erskingardner: doesn't matter. leaving.

src/whitenoise/users/key_package.rs
erskingardner: yeah - it apparently was triggering a flaky test 🤷‍♂️

it doesn't hurt in any case.

src/whitenoise/users.rs
erskingardner: this is now decoupled.

PR #676, opened 2026-03-20
feat: add per-user user stream

erskingardner:
## Summary
- add a per-user streaming module keyed by `PublicKey`
- add race-free `subscribe_to_user` with an immediate local snapshot and background resolution for unknown metadata
- emit `UserCreated`, `MetadataChanged`, and `LocalMetadataChanged` from the scoped write paths only
- add unit tests for the manager/types and the new subscription/update paths

## Scope
- additive only for single-user subscriptions
- leaves search/list APIs untouched
- does not batch users or do public API cleanup

## Testing
- `just precommit-quick`

## Summary by CodeRabbit

* **New Features**
  * Subscribe to a user to receive an initial snapshot plus real-time updates for user creation, metadata changes, and local metadata edits; subscriptions deliver the latest merged state, preserve ordering, and drain queued updates so the snapshot reflects the most recent change.

* **Tests**
  * Expanded test coverage for subscriptions, update delivery, snapshot correctness, triggers, and stream lifecycle.

### DIFF
--- a/src/whitenoise/mod.rs
+++ b/src/whitenoise/mod.rs
@@ -1047,7 +1047,7 @@ impl Whitenoise {
                 Err(broadcast::error::TryRecvError::Empty) => break,
                 Err(broadcast::error::TryRecvError::Lagged(n)) => {
                     tracing::warn!(
-                        target: "whitenoise::mod",
+                        target: "whitenoise",
                         "subscription drain lagged by {n} user updates"
                     );
                     continue;


dannym-arx (commented):
I'm not sure if a per user stream makes sense, couldn't we have a generic user update stream?

the step I want to avoid is the flutter side having to choose which user streams to subscribe to, and rather move that logic to the rust side

the flutter side would just get a generic user updated event out of the stream, which it can choose to act upon or ignore, depending on what it needs

PR #681, opened 2026-03-23
Refactor/async discovery rebuilds

jgmontoya:
## Summary by CodeRabbit

* **Performance Improvements**
  * Deterministic batching and hash-based per-batch updates skip unchanged work, reducing relay activity.

* **New Features**
  * Background discovery sync worker with debounced rebuilds and a wait-for-completion API.
  * Scheduled subscription health check every 15 minutes.
  * Exposed watched-user count and ability to retire all discovery subscriptions.

* **Reliability / UX**
  * Account activation, follow, and user-sync flows request async rebuilds instead of blocking on immediate refresh.

### DIFF
--- a/justfile
+++ b/justfile
@@ -234,8 +234,24 @@ precommit-quick:
     @just _run-quiet "check-fmt"    "fmt"
     @just _run-quiet "check-docs"   "docs"
     @just _run-quiet "check-clippy" "clippy"
-    @just _run-quiet "test"         "tests"
+    @just _run-quiet "test-unit"    "tests"
     @echo "PRECOMMIT PASSED"
+
+# Unit tests only (no integration-tests feature, no e2e binaries).
+# Used by precommit-quick for fast local validation without Docker.
+test-unit:
+    #!/usr/bin/env bash
+    set -euo pipefail
+    if command -v cargo-nextest &> /dev/null; then
+        echo "Running unit tests with nextest (parallel)..."
+        cargo nextest run --features cli --all-targets \
+            --filter-expr 'not binary(cli_e2e) and not binary(cli_relay_control_e2e)'
+        cargo test --features cli --doc
+    else
+        echo "Running unit tests with cargo test..."
+        cargo test --features cli --all-targets
+        cargo test --features cli --doc
+    fi
 
 # Quick pre-commit with verbose output
 precommit-quick-verbose: check test


src/whitenoise/discovery_sync_worker.rs
```diff
@@ -0,0 +1,314 @@
...
+use tokio::time::timeout;
+
+use crate::whitenoise::error::Result;
+
+const DEBOUNCE_WINDOW: Duration = Duration::from_millis(500);
+
+pub(crate) struct DiscoverySyncWorker {
+    notify: Notify,
+    generation: watch::Sender<u64>,
+}
```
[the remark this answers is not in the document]
jgmontoya: ▎ Both behaviors are by design.

  ▎ "Early return on unrelated rebuilds" — The worker always rebuilds the entire discovery state (all watched users + all follow lists). Any rebuild that completes after a signal satisfies every waiting
  caller — there's no "my specific rebuild" because the output is identical regardless of which signal triggered it. Multiple callers coalescing to one rebuild is the design intent.

  ▎ "Indefinite hang" — The only caller is setup_all_subscriptions during startup, which runs in the same initialize_whitenoise function immediately after the worker is spawned. The worker loop only exits
  on shutdown signal, at which point no code calls this method. The map_err handles the sender-dropped case, though that path is unreachable in practice.

  ▎ Per-request tracking would add significant complexity for scenarios that can't occur in the actual call graph.

src/whitenoise/event_processor/event_handlers/handle_contact_list.rs
```diff
@@ -147,13 +147,7 @@ impl Whitenoise {
...
             );
 
-            if let Err(error) = whitenoise.refresh_all_global_subscriptions().await {
-                tracing::warn!(
-                    target: "whitenoise::handle_contact_list",
-                    "Failed to refresh discovery subscriptions before catch-up: {}",
-                    error
-                );
-            }
+            whitenoise.discovery_sync_worker.request_rebuild();
```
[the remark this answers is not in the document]
jgmontoya: ▎ This is pre-existing behavior, not a regression. The original code also had refresh_all_global_subscriptions() inside the tokio::spawn, after the same empty-list guard at line 116-118. The placement is
   unchanged — only the call was swapped from an inline await to a fire-and-forget signal.

  ▎ The 15-minute SubscriptionHealthCheck self-heals any drift from this edge case.

src/whitenoise/mod.rs
```diff
@@ -1359,10 +1353,28 @@ impl Whitenoise {
...
+            tracing::info!(
+                target: "whitenoise::is_global_subscriptions_operational",
+                db_users = db_user_count,
+                plane_users = plane_user_count,
+                "Watched-user count mismatch between DB and discovery plane",
+            );
+            return Ok(false);
+        }
+
+        Ok(true)
```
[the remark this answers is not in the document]
jgmontoya: ▎ Valid observation — the parity check only covers watched-user subscriptions. In practice, the existing has_discovery_subscriptions() check on line 1355 already validates that both
  public_subscription_ids and follow_list_subscription_ids are non-empty, catching the most common failure mode (no subscriptions at all).

  ▎ Follow-list subscriptions are typically a handful accounts. Adding hash-based health checks would couple the health predicate to internal state representation for a scenario that self-heals within one
  SubscriptionHealthCheck interval (15 min). If this becomes a real issue we can revisit, but the complexity trade-off isn't justified right now.

src/relay_control/discovery.rs
```diff
@@ -17,6 +20,47 @@ use crate::{
 
 const MAX_USERS_PER_PUBLIC_DISCOVERY_SUBSCRIPTION: usize = 500;
 
+fn batch_count_for(user_count: usize) -> usize {
+    user_count.div_ceil(MAX_USERS_PER_PUBLIC_DISCOVERY_SUBSCRIPTION)
+}
```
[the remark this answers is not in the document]
jgmontoya: ▎ The 500 limit is a soft relay capacity target, not a protocol constraint — empirical testing confirms relays handle well beyond this. Pubkeys are uniformly distributed secp256k1 curve points, so
  tail_bytes % batch_count gives near-perfect uniformity. Empirically, 1000 users / 2 batches splits within ~2-3% of 500/500.

  ▎ The suggested fix (sequential fill / min-heap) would destroy the key property this PR introduces: per-batch stability. With capacity-aware bucketing, adding one user can cascade users across multiple
  batches — exactly the O(N) relay churn that hash-based assignment eliminates. The trade-off (tiny variance in batch sizes vs ~90% reduction in relay churn) is explicit and intentional.

src/relay_control/discovery.rs
```diff
@@ -1,3 +1,6 @@
+use std::collections::hash_map::DefaultHasher;
+use std::hash::{Hash, Hasher};
```
erskingardner, line 2: My agent left this note. I'm not sure that it's really important in this case, an extra rebuild on start up is probably fine, but the note about adding a comment somewhere to explain might be helpful in case this causes strange tests failures in the future: 

> `DefaultHasher` is the standard library's *internal* hasher for `HashMap`. Its algorithm (currently SipHash 1-3) is seeded randomly at process start for DOS-resistance and is explicitly **not guaranteed to be stable across Rust releases**. The `previous_user_set_hash` and `previous_batch_hashes` in `DiscoveryPlaneState` are held only in memory within a single process run, so cross-restart stability doesn't matter here. However, if the Rust toolchain is ever updated and the hash output changes for the same input, the in-memory skip logic would silently re-subscribe everything after first use — which is harmless but surprising. More importantly, the random seed means that unit tests like `hash_is_stable_for_same_sorted_input` only verify stability within a single test run, not across runs or versions.
>
> A more explicit alternative is `std::hash::BuildHasherDefault::<std::collections::hash_map::DefaultHasher>` or any deterministic hasher (e.g. `fnv`, `rustc-hash`, or even a simple XOR-fold over fixed-width byte slices). Since the hash is only used in-process for change detection, the random seed is the primary concern rather than cross-process stability. At minimum, document the limitation in the function's doc comment so future readers understand the guarantee is only intra-process.

src/relay_control/discovery.rs
```diff
@@ -101,34 +151,39 @@ impl DiscoveryPlane {
...
+    /// batches via hash-based assignment.
     #[perf_instrument("relay")]
-    pub(crate) async fn sync(
+    pub(crate) async fn sync_watched_users(
         &self,
         watched_users: &[PublicKey],
-        follow_list_accounts: &[(PublicKey, Option<Timestamp>)],
         public_since: Option<Timestamp>,
     ) -> Result<()> {
-        // Short-circuit: no relays configured — retire any existing subscriptions.
```
erskingardner, line 111: It feels like it would be hard to configure this without relays though since we default to a few but worth a quick thought - I'm happy if you just resolve this and move on though

> The original `sync()` function had an explicit early-return guard:
> 
> ```rust
> // Short-circuit: no relays configured — retire any existing subscriptions.
> if self.config.relays.is_empty() { ... return Ok(()); }
> ```
> 
> This guard was removed in the refactor. Both `sync_watched_users` and `sync_follow_lists` now skip to `self.start().await?` when no relay is connected, but `start()` calls `ensure_relays_connected(&self.config.relays)` — which, with an empty slice, either silently succeeds (no-op) or behaves unexpectedly depending on the `RelaySession` implementation. In a deployment with zero configured relays, the previous code retired existing subscriptions and returned early. The new code may leave stale subscriptions in place indefinitely (they never hash-match any new batch, so the hash-skip path never fires). Add an early exit for `self.config.relays.is_empty()` in both functions, calling `retire_all()` or returning `Ok(())` as appropriate.

src/whitenoise/mod.rs
```diff
@@ -1250,7 +1245,7 @@ impl Whitenoise {
                 target: "whitenoise::ensure_global_subscriptions",
                 "Global subscriptions not operational, refreshing..."
             );
-            Self::setup_global_users_subscriptions(self).await?;
+            self.discovery_sync_worker.request_rebuild();
```
erskingardner: This feels like it could bite us no? Especially the relays degraded then triggering lots of repeat rebuilds? 

> ```rust
> pub async fn ensure_global_subscriptions(&self) -> Result<()> {
>     let is_operational = self.is_global_subscriptions_operational().await?;
> 
>     if !is_operational {
>         self.discovery_sync_worker.request_rebuild();  // fire-and-forget
>     }
> 
>     Ok(())
> }
> ```
> 
> The caller (`ensure_all_subscriptions`, the scheduled health-check) receives `Ok(())` immediately even when a rebuild was requested. This means the health-check task reports success before the rebuild has actually run. This is arguably intentional (non-blocking), but it creates an observable problem: if the health check fires, triggers a rebuild, and the next health check fires before the rebuild completes (e.g. under relay pressure), the parity check will fail again and another rebuild will be queued. Under sustained relay degradation this can queue multiple back-to-back rebuilds.
> 
> More importantly, the doc comment on `ensure_all_subscriptions` says *"triggers a rebuild for any that have drifted"* — callers who read this expect the rebuild to have started at minimum. That part is true. But returning `Ok(())` with the rebuild still in-flight is a semantic mismatch. Consider at minimum renaming to make the fire-and-forget nature explicit, or documenting the async gap more prominently.

src/whitenoise/discovery_sync_worker.rs
```diff
@@ -0,0 +1,314 @@
...
+            ))
+        })?;
+        Ok(())
+    }
+
+    /// Runs the worker loop using `sync_discovery_subscriptions` from the
+    /// global Whitenoise instance. This is the production entry point.
+    pub(crate) async fn run(&self, shutdown_rx: watch::Receiver<bool>) {
+        self.run_with(shutdown_rx, || async {
+            crate::whitenoise::Whitenoise::get_instance()?
```
erskingardner: > `DiscoverySyncWorker` is a field of `Whitenoise`, yet its production entry point reaches back to `Whitenoise::get_instance()` to get the `&'static Whitenoise` reference needed to call `sync_discovery_subscriptions`. This is a design smell: the worker is *owned by* the instance but obtains access to it through the global singleton rather than through a `&'static Whitenoise` passed at construction or spawn time. The `run_with` escape hatch solves the testability problem neatly, but the awkward singleton back-reference warrants a comment explaining why it's structured this way. A cleaner alternative (if the lifetime allows) would be to pass `whitenoise_ref: &'static Whitenoise` into `run()` directly, matching the pattern used by `start_scheduled_tasks`.

src/relay_control/discovery.rs
jgmontoya: I don't think this is something to worry about honestly, subscriptions have to be rebuilt upon process start anyway

src/relay_control/discovery.rs
jgmontoya: The code actually works correctly without the guard, subscribe_with_id_to with an empty relay slice is a no-op, state commits normally, and the hash guard prevents redundant work on subsequent calls

src/whitenoise/mod.rs
jgmontoya: The back-to-back rebuild scenario can't happen, the health check fires every 15 minutes and a rebuild takes seconds. Even in the theoretical case, `Notify::notify_one()` stores at most one permit, so concurrent signals coalesce into at most 2 rebuilds (tested in rapid_signals_coalesce). There's no unbounded queue.
                                                                                                                                                                                                             
The behavioral change from inline-await to fire-and-forget is intentional, `ensure_global_subscriptions` is called from a best-effort periodic health check, not from code that needs to observe the rebuild's completion. The old code also tolerated rebuild failures (every call site used `if let Err(e) = ... { warn!(...) }`).

src/whitenoise/discovery_sync_worker.rs
```diff
@@ -0,0 +1,313 @@
...
+        tracing::debug!(
+            target: "whitenoise::discovery_sync",
+            "Discovery sync rebuild requested"
+        );
+        self.notify.notify_one();
+    }
+
+    pub(crate) async fn request_rebuild_and_wait(&self) -> Result<()> {
+        let mut rx = self.generation.subscribe();
+        let current = *rx.borrow_and_update();
```
mubarakcoded, line 35: A snapshot is taken before the notify, which opens a race. If a rebuild is already in progress when this is called, it'll complete and advance the generation past the current one. wait_for returns immediately, before the rebuild triggered by our notify_one() on line 36 ever runs. Caller thinks its changes are synced, but they're not. I suggest moving the snapshot after the notify so any in-flight rebuild is already in the past

src/whitenoise/discovery_sync_worker.rs
```diff
@@ -0,0 +1,313 @@
...
+    async fn run_with<F, Fut>(&self, mut shutdown_rx: watch::Receiver<bool>, work: F)
+    where
+        F: Fn() -> Fut,
+        Fut: Future<Output = Result<()>>,
+    {
+        loop {
+            tokio::select! {
+                _ = shutdown_rx.changed() => break,
+                _ = self.notify.notified() => {
+                    loop {
```
mubarakcoded, line 66: Once we enter this inner loop, shutdown_rx is never checked again. A shutdown signal during an active rebuild (or while waiting on the 500ms debounce) is silently ignored until the loop naturally exits. 
Consider folding the shutdown arm into the inner select

src/whitenoise/discovery_sync_worker.rs
jgmontoya: I don't think there's a real race here because the `request_rebuild_and_wait` is only used during startup. Even if it did happen, the next trigger would immediately correct any 'staleness' of the data

src/whitenoise/discovery_sync_worker.rs
jgmontoya: I don't think it's worth it to check for shutdown mid-rebuild, but I added a check within the inner loop so we don't need to wait for the debounce

PR #685, opened 2026-03-24
feat(groups): keep removed groups visible with real-time RemovedFromGroup trigger

### LINKED ISSUE #683: fix: emit RemovedFromGroup stream event when admin removes a member
# Leave Group — Real-time Removal Plan

## Problem

When an admin removes a member, the removed user's client processes the MLS commit and MDK marks the group inactive — but no stream event is emitted. The group stays visible in Flutter's chat list until the user manually refreshes. Since Flutter is stream-driven, there is no poll interval that catches it.

---

### Changes

**1. New `ChatListUpdateTrigger` variant**
- Add `RemovedFromGroup` to `ChatListUpdateTrigger` in `chat_list_streaming/types.rs`
- Semantically distinct from `ChatArchiveChanged` (which is a user-initiated archive action)

**2. Emit on removal commit**
- In `handle_mls_message.rs`, after processing a `MessageProcessingResult::Commit`, check whether the account is still an active MDK member
- If not: emit `RemovedFromGroup` via the existing `emit_chat_list_update` mechanism — no DB write needed, MDK already marks the group inactive
- Existing `background_refresh_account_group_subscriptions` call stays in place

### Estimated size
~50–80 LOC across 3 files.

---

## PR 2 — Flutter handler (whitenoise)

**Depends on:** PR 1 merged and published.

### Changes
- Handle `RemovedFromGroup` trigger in `use_chat_list.dart` — remove the group from the chat map
- Navigate away if the user is currently inside the removed group's chat screen

mubarakcoded:
## Summary
- Add `removed_at` column to `accounts_groups` (migration 0038) to track involuntary removal separately from `archived_at` (user-initiated)
- `mark_removed_atomic()`: single `UPDATE … WHERE removed_at IS NULL` that sets `removed_at` and `user_confirmation = true` atomically — eliminates check-then-update race and ensures a pending invite removed before acceptance is no longer surfaced as pending
- `mark_as_removed()` on `Whitenoise`: idempotent, emits `RemovedFromGroup` only when the row actually changed
- `ChatListUpdateTrigger::RemovedFromGroup`: semantically distinct from `ChatArchiveChanged`; routed by `item.archived_at` so archived-stream subscribers receive the update too
- `ChatListItem.removed_at`: exposed on the stream payload so Flutter can render read-only removed state without a separate query
- `visible_groups()`: drop MDK `Active`-only filter — `AccountGroup` is source of truth; inactive MDK groups included only when `membership.is_removed()`, preventing unrelated inactive groups from reappearing
- `handle_mls_message`: after `Commit`, call `mdk.get_group()` to check membership; if inactive call `mark_as_removed()` and propagate errors (non-best-effort); skip background image cache sync for removed groups

Known gap: MDK deduplicates commits so a retry after a persistent `mark_as_removed` failure will not re-run the removal check — startup reconciliation is a follow-up.

Closes #683

## Testing
- `just precommit-quick`

## Summary by CodeRabbit

* **New Features**
  * Admins can mark users as removed from group conversations; removed users remain visible/read-only and show a removal timestamp.
  * Migration adds a nullable removal timestamp for group memberships.

* **Bug Fixes**
  * Removal state is persisted and cannot be cleared by resaves or re-processing.
  * Pending membership logic now excludes removed entries.
  * Chat-list streams emit explicit "removed from group" updates to the correct active or archived channel.

* **Tests**
  * Added tests for removal behavior, idempotency, persistence, visibility, and chat-list update emission.

### DIFF
--- a/src/whitenoise/accounts_groups.rs
+++ b/src/whitenoise/accounts_groups.rs
@@ -64,8 +64,11 @@ impl AccountGroup {
     }
 
     /// Returns true if this group is pending user confirmation.
+    ///
+    /// A removed group is never pending even if `user_confirmation` was `None` at the
+    /// time of removal: an admin commit that removes the user supersedes the pending state.
     pub fn is_pending(&self) -> bool {
-        self.user_confirmation.is_none()
+        self.user_confirmation.is_none() && self.removed_at.is_none()
     }
 
     /// Returns true if the user has accepted this group.
@@ -399,7 +402,12 @@ impl Whitenoise {
             .ok_or(WhitenoiseError::GroupNotFound)?;
 
         let Some(updated) = account_group.mark_removed(self).await? else {
-            return Ok(account_group);
+            // Another task won the atomic update — reload the current row so
+            // we return the persisted removed_at rather than the stale pre-load value.
+            let current = AccountGroup::get(self, &account.pubkey, mls_group_id)
+                .await?
+                .ok_or(WhitenoiseError::GroupNotFound)?;
+            return Ok(current);
         };
 
         self.emit_chat_list_update(
@@ -1385,9 +1393,8 @@ mod tests {
 
         assert!(removed.is_removed());
         assert!(removed.removed_at.is_some());
-        // user_confirmation must be set to true — a pending invite that gets a
-        // removal commit before the user accepts must not surface as pending
-        assert_eq!(removed.user_confirmation, Some(true));
+        // A group removed before the user accepted must not surface as pending.
+        assert!(!removed.is_pending());
     }
 
     #[tokio::test]
--- a/src/whitenoise/chat_list_streaming/types.rs
+++ b/src/whitenoise/chat_list_streaming/types.rs
@@ -13,9 +13,9 @@ pub enum ChatListUpdateTrigger {
     /// Emitted to both active and archived channels so each can add/remove accordingly.
     ChatArchiveChanged,
     /// The account was involuntarily removed from this group by an admin.
-    /// The group stays in the active chat list (read-only) until the user
-    /// explicitly archives or deletes it. The item's `removed_at` field is set.
-    /// Emitted only to the active channel.
+    /// The group stays visible (read-only) until the user explicitly archives or
+    /// deletes it. The item's `removed_at` field is set. Routed to the active
+    /// channel when not archived, or the archived channel if already archived.
     RemovedFromGroup,
 }
 
--- a/src/whitenoise/database/accounts_groups.rs
+++ b/src/whitenoise/database/accounts_groups.rs
@@ -221,7 +221,7 @@ impl AccountGroup {
         let rows = sqlx::query_as::<_, AccountGroupRow>(
             "SELECT *
              FROM accounts_groups
-             WHERE account_pubkey = ? AND user_confirmation IS NULL",
+             WHERE account_pubkey = ? AND user_confirmation IS NULL AND removed_at IS NULL",
         )
         .bind(account_pubkey.to_hex())
         .fetch_all(&database.pool)
@@ -294,7 +294,7 @@ impl AccountGroup {
                -- archived_at is intentionally excluded: archive/unarchive use
                -- update_archived_at() so save() never clobbers user preference.
                -- removed_at is intentionally excluded: removal is set by
-               -- update_removed_at() when an admin removal commit is detected.
+               -- mark_removed_atomic() when an admin removal commit is detected.
                -- Write-once: preserve existing dm_peer_pubkey if already set.
                -- Many code paths construct AccountGroup without knowing the DM peer,
                -- so we only fill this on first write and never overwrite a correct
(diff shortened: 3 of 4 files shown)


src/whitenoise/database/accounts_groups.rs
```diff
@@ -362,6 +368,63 @@ impl AccountGroup {
...
+        )
+        .bind(now_ms)
+        .bind(now_ms)
+        .bind(id)
+        .fetch_optional(&database.pool)
+        .await?;
+
+        Ok(row.map(Into::into))
+    }
+
```
jgmontoya: I don't really understand why we need both methods.

jgmontoya: I really don't like that we're settings `user_confirmation = 1` here. The user has not confirmed anything by calling this so it's odd. I think that elsewhere we should add the check for removed_at if it applies rather than reusing `user_confirmation` for this

src/whitenoise/database/accounts_groups.rs
```diff
@@ -1549,4 +1616,134 @@ mod tests {
...
+        assert!(account_group.removed_at.is_none());
+
+        let now = Utc::now();
+        let updated = account_group
+            .update_removed_at(Some(now), &whitenoise.database)
+            .await
+            .unwrap();
+
+        assert!(updated.removed_at.is_some());
+        assert!(updated.is_removed());
```
jgmontoya: this `is_removed()` call caught me a bit off-guard, because it's a method defined on accounts_groups and not database/accounts_groups.

I'd remove that line from this test.

src/whitenoise/database/accounts_groups.rs
```diff
@@ -1549,4 +1616,134 @@ mod tests {
...
+            .unwrap();
+        assert!(removed.is_removed());
+
+        // Clear removed
+        let cleared = removed
+            .update_removed_at(None, &whitenoise.database)
+            .await
+            .unwrap();
+        assert!(cleared.removed_at.is_none());
+        assert!(!cleared.is_removed());
```
jgmontoya: same with this one

src/whitenoise/database/accounts_groups.rs
```diff
@@ -1549,4 +1616,134 @@ mod tests {
...
+        let found = AccountGroup::find_by_account_and_group(
+            &account.pubkey,
+            &group_id,
+            &whitenoise.database,
+        )
+        .await
+        .unwrap()
+        .unwrap();
+
+        assert!(found.is_removed());
```
jgmontoya: same here

src/whitenoise/database/accounts_groups.rs
```diff
@@ -1549,4 +1616,134 @@ mod tests {
...
+            AccountGroup::find_or_create(&account.pubkey, &group_id, None, &whitenoise.database)
+                .await
+                .unwrap();
+
+        // Mark as removed
+        let removed = ag
+            .update_removed_at(Some(Utc::now()), &whitenoise.database)
+            .await
+            .unwrap();
+        assert!(removed.is_removed());
```
jgmontoya: and here

src/whitenoise/database/accounts_groups.rs
```diff
@@ -1549,4 +1616,134 @@ mod tests {
...
+            &account.pubkey,
+            &group_id,
+            &whitenoise.database,
+        )
+        .await
+        .unwrap()
+        .unwrap();
+
+        assert!(
+            found.is_removed(),
```
jgmontoya: and here

src/whitenoise/event_processor/event_handlers/handle_mls_message.rs
```diff
@@ -218,8 +219,32 @@ impl Whitenoise {
                     "Processed commit for group {}",
                     hex::encode(mls_group_id.as_slice())
                 );
+
+                // Detect self-removal: if the commit left us inactive, persist the
```
jgmontoya: why "self-removal" ? this would be any removal, no?

src/whitenoise/database/accounts_groups.rs
mubarakcoded: Removed update_removed_at entirely. It was only ever called in tests (the production path always went through mark_removed_atomic, so the two-method situation was already an artifact. Tests that previously required a removed row for setup now call mark_removed_atomic directly; the three tests that were exercising update_removed_at directly are removed.

mubarakcoded: Agreed, that was the wrong layer to fix it. Dropped user_confirmation = 1 from mark_removed_atomic.

src/whitenoise/accounts_groups.rs
```diff
@@ -177,6 +193,20 @@ impl AccountGroup {
...
+    pub(crate) async fn mark_removed(
+        &self,
+        whitenoise: &Whitenoise,
+    ) -> Result<Option<Self>, WhitenoiseError> {
+        let updated = self
+            .mark_removed_atomic(&whitenoise.database)
+            .await
+            .map_err(WhitenoiseError::from)?;
+        Ok(updated)
+    }
```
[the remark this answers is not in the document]
mubarakcoded: The CASE patch isn't needed - we already solved this more cleanly in the previous commit. mark_removed_atomic no longer writes user_confirmation at all, so a declined invite (Some(false)) is naturally preserved. The pending-invite case is handled at the right layer instead: is_pending() now guards with removed_at.is_none() and find_pending_for_account excludes removed_at IS NOT NULL rows. Added the regression test you asked for in bb98e4b.

src/whitenoise/chat_list_streaming/types.rs
```diff
@@ -12,6 +12,11 @@ pub enum ChatListUpdateTrigger {
     /// `archived_at` field indicates direction: `Some` = archived, `None` = unarchived.
     /// Emitted to both active and archived channels so each can add/remove accordingly.
     ChatArchiveChanged,
+    /// The account was involuntarily removed from this group by an admin.
+    /// The group stays in the active chat list (read-only) until the user
+    /// explicitly archives or deletes it. The item's `removed_at` field is set.
+    /// Emitted only to the active channel.
```
jgmontoya: I think it's also emitted for the archived, no?

PR #686, opened 2026-03-25
Fixing bugs after refactor

### LINKED ISSUE #684: Pre-existing logic bugs surfaced during PR #586 structural refactor
## Background

During the structural refactor in PR #586 (split `login.rs` / `mod.rs` into focused submodules), several pre-existing logic issues were identified that were carried over verbatim from the original code. None of these were introduced by the refactor. This issue tracks them for a follow-up fix.

Related PR: https://github.com/marmot-protocol/whitenoise-rs/pull/586
Flagged by: @dannym-arx (https://github.com/marmot-protocol/whitenoise-rs/pull/586#discussion_r2976284205)

---

## Issue 1 — Missing active-session short-circuit in `login_external_signer_start`

**File:** `src/whitenoise/accounts/login_external_signer.rs` (around line 41)

`login_external_signer_start` always proceeds into `setup_external_signer_account_without_relays`, even if the account already has an active session. Unlike `login_start` / `login_with_external_signer`, there is no early-return guard. This can re-activate or migrate a live local-key session to `AccountType::External` and delete its stored private key before the staged login has succeeded.

**Fix:** Add the same active-session short-circuit used in `login_start`: check `Account::find_by_pubkey` + `!pending_logins.contains_key` + `background_task_cancellation.contains_key` and return `LoginStatus::Complete` early.

---

## Issue 2 — Serialized and fatal 10050/10051 publishes in `login_external_signer_publish_default_relays`

**File:** `src/whitenoise/accounts/login_external_signer.rs` (around line 172)

The loop over `[RelayType::Nip65, RelayType::Inbox, RelayType::KeyPackage]` serializes the three relay-list publishes and treats Inbox (10050) / KeyPackage (10051) failures as fatal whenever `discovered.nip65` is `None`. This diverges from the established login behavior (see `publish_relay_lists_with_signer` which fires all three concurrently via `tokio::join!`) where only NIP-65/10002 failure should gate completion.

**Fix:** Align with `publish_relay_lists_with_signer`: run all three concurrently via `tokio::join!`, propagate only the NIP-65 result as fatal, and demote 10050/10051 failures to warnings.

---

## Issue 3 — Stale DB state after `login_with_custom_relay` retry in `login_publish_default_relays`

**File:** `src/whitenoise/accounts/login_multistep.rs` (around line 195)

After an incomplete `login_with_custom_relay` retry, `try_discover_relay_lists` may have already synced that relay's empty result into the DB while the pending snapshot still preserves previously found lists. `login_publish_default_relays` only calls `add_relays` for missing kinds, so it can complete activation against a DB that is missing relay rows that exist in the stash (because the last custom-relay attempt wrote empty rows over them).

**Fix:** Call `sync_discovered_relay_lists` (or equivalent) from the stash before `add_relays` fills defaults, so the DB is consistent with the stash before activation.

---

## Issue 4 — `login_cancel` removes stash before destructive cleanup succeeds

**File:** `src/whitenoise/accounts/login_multistep.rs` (around line 417)

`login_cancel` removes the `pending_logins` entry before `account.delete()` runs. If the delete fails, the partial account remains in the DB but the stash is gone, so a retry of `login_cancel` becomes a no-op and the best-effort keychain cleanup never runs.

**Fix:** Hold the removed stash value locally; restore it into `pending_logins` if `account.delete()` returns an error, or perform all destructive steps before removing the stash.

---

## Issue 5 — `get_signer_for_account` selects signer by map presence, not account type

**File:** `src/whitenoise/signer.rs` (around line 140)

`get_signer_for_account` decides which signer to use based solely on whether an entry exists in the `external_signers` map, rather than checking `account.account_type`. This makes stale-cleanup state observable: a leftover external signer entry can hijack a later local login for the same pubkey, and a stale local key can let an external-type account keep working after its external signer is removed.

**Fix:** Gate the external-signer path on `account.account_type == AccountType::External` in addition to the map lookup.

---

## Issue 6 — `TryRecvError::Lagged` during initial snapshot drain in streaming subscriptions

**File:** `src/whitenoise/streaming.rs` (around line 73)

When the initial broadcast drain encounters `TryRecvError::Lagged(n)`, the code logs a warning and continues. However, once lagged, some updates that arrived after the DB snapshot are already gone. Continuing breaks the "initial snapshot + subsequent updates" guarantee and can permanently omit messages or chat-list items until the client resubscribes.

**Fix:** On `Lagged`, rebuild the snapshot from storage (re-query the DB) and restart the drain loop, rather than continuing with an incomplete merge.

sergey3bv:
Should close #684 

## Summary by CodeRabbit

* **New Features**
  * Added archived chat-list subscription and a notifications subscription.

* **Bug Fixes**
  * Faster login for already-active accounts by skipping redundant setup.
  * More reliable relay persistence and concurrent publishing with clear fatal vs non-fatal failures.
  * External-signing accounts now require a registered signer (no local-key fallback).
  * Improved subscription lag recovery by rebuilding state from storage.
  * Safer login cancel flow preserves retryability.
  * Initialization skips accounts with relay-resolution errors instead of aborting all setup.

* **Tests**
  * Updated tests to enforce account-type-specific signer behavior.

### DIFF
--- a/src/whitenoise/mod.rs
+++ b/src/whitenoise/mod.rs
@@ -758,6 +758,8 @@ impl Whitenoise {
 
 #[cfg(test)]
 pub mod test_utils {
+    use std::time::Duration;
+
     use super::*;
     use crate::whitenoise::accounts_groups::AccountGroup;
     use crate::whitenoise::group_information::GroupInformation;
@@ -1026,6 +1028,7 @@ pub mod test_utils {
             })
             .clone();
 
+        wait_for_test_relays().await;
         Whitenoise::initialize_whitenoise(config).await.unwrap();
         Whitenoise::get_instance().unwrap()
     }
@@ -1076,7 +1079,10 @@ pub mod test_utils {
         whitenoise: &Whitenoise,
         publisher_accounts: &[&Account],
     ) {
-        tokio::time::timeout(std::time::Duration::from_secs(5), async {
+        const OUTER_TIMEOUT: Duration = Duration::from_secs(60);
+        const PER_FETCH_TIMEOUT: Duration = Duration::from_secs(8);
+
+        tokio::time::timeout(OUTER_TIMEOUT, async {
             loop {
                 let mut all_available = true;
 
@@ -1088,13 +1094,17 @@ pub mod test_utils {
                             .unwrap(),
                     );
 
-                    match whitenoise
-                        .relay_control
-                        .fetch_user_key_package(publisher_account.pubkey, &relay_urls)
-                        .await
-                    {
-                        Ok(Some(_)) => {}
-                        Ok(None) | Err(_) => {
+                    let fetch_result = tokio::time::timeout(
+                        PER_FETCH_TIMEOUT,
+                        whitenoise
+                            .relay_control
+                            .fetch_user_key_package(publisher_account.pubkey, &relay_urls),
+                    )
+                    .await;
+
+                    match fetch_result {
+                        Ok(Ok(Some(_))) => {}
+                        Ok(Ok(None)) | Ok(Err(_)) | Err(_) => {
                             all_available = false;
                             break;
                         }
@@ -1105,7 +1115,7 @@ pub mod test_utils {
                     break;
                 }
 
-                tokio::time::sleep(std::time::Duration::from_millis(25)).await;
+                tokio::time::sleep(Duration::from_millis(25)).await;
             }
         })
         .await
--- a/src/whitenoise/push_notifications.rs
+++ b/src/whitenoise/push_notifications.rs
@@ -2531,12 +2531,10 @@ mod tests {
             .await
             .unwrap();
 
-        tokio::time::pause();
         let settings = whitenoise
             .update_notifications_enabled(&admin_account, false)
             .await
             .unwrap();
-        tokio::time::resume();
         assert!(!settings.notifications_enabled);
 
         let cached_after_disable = GroupPushToken::find_by_account_and_group(
--- a/src/whitenoise/scheduled_tasks/tasks/key_package_maintenance.rs
+++ b/src/whitenoise/scheduled_tasks/tasks/key_package_maintenance.rs
@@ -424,7 +424,7 @@ mod tests {
     use crate::whitenoise::database::published_key_packages::PublishedKeyPackage;
     use crate::whitenoise::key_packages::has_encoding_tag;
     use crate::whitenoise::relays::Relay;
-    use crate::whitenoise::test_utils::create_mock_whitenoise;
+    use crate::whitenoise::test_utils::{create_mock_whitenoise, wait_for_key_package_publication};
 
     /// Publishes a key package without the encoding tag for testing outdated package rotation.
     async fn publish_outdated_key_package(
@@ -465,6 +465,26 @@ mod tests {
         Ok(event_id)
     }
 
+    async fn first_tracked_key_package(
+        whitenoise: &Whitenoise,
+        account: &Account,
+        events: &[Event],
+    ) -> PublishedKeyPackage {
+        for event in events {
+            if let Some(pkg) = PublishedKeyPackage::find_by_event_id(
+                &account.pubkey,
+                &event.id.to_hex(),
+                &whitenoise.database,
+            )
+            .await
+            .unwrap()
+            {
+                return pkg;
+            }
+        }
+        panic!("expected a relay key package with a published_key_packages row");
+    }
+
     #[tokio::test]
     async fn test_execute_deletes_outdated_packages_when_valid_exists() {
         let (whitenoise, _data_temp, _logs_temp) = create_mock_whitenoise().await;
@@ -472,6 +492,7 @@ mod tests {
 
         // Create account (this publishes a valid key package automatically)
         let account = whitenoise.create_identity().await.unwrap();
+        wait_for_key_package_publication(whitenoise, &[&account]).await;
         let kp_relays = account.key_package_relays(whitenoise).await.unwrap();
         tokio::time::sleep(Duration::from_millis(500)).await;
 
@@ -581,6 +602,7 @@ mod tests {
         let whitenoise: &'static Whitenoise = Box::leak(Box::new(whitenoise));
 
         let account = whitenoise.create_identity().await.unwrap();
+        wait_for_key_package_publication(whitenoise, &[&account]).await;
         tokio::time::sleep(Duration::from_millis(500)).await;
 
         let before = whitenoise
@@ -589,14 +611,7 @@ mod tests {
             .unwrap();
         assert_eq!(before.len(), 1, "Should start with exactly one key package");
 
-        let tracked = PublishedKeyPackage::find_by_event_id(
-            &account.pubkey,
-            &before[0].id.to_hex(),
-            &whitenoise.database,
-        )
-        .await
-        .unwrap()
-        .unwrap();
+        let tracked = first_tracked_key_package(whitenoise, &account, &before).await;
 
         let mdk = whitenoise.create_mdk_for_account(account.pubkey).unwrap();
         mdk.delete_key_package_from_storage_by_hash_ref(&tracked.key_package_hash_ref)
@@ -638,6 +653,7 @@ mod tests {
         let whitenoise: &'static Whitenoise = Box::leak(Box::new(whitenoise));
 
         let account = whitenoise.create_identity().await.unwrap();
+        wait_for_key_package_publication(whitenoise, &[&account]).await;
         tokio::time::sleep(Duration::from_millis(500)).await;
 
         let before = whitenoise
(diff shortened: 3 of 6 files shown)


src/whitenoise/streaming.rs
```diff
@@ -30,90 +29,68 @@ impl Whitenoise {
...
+                        &self.database,
+                    )
+                    .await
+                    .map_err(|e| {
+                        WhitenoiseError::from(anyhow::anyhow!(
+                            "Failed to read cached messages: {}",
+                            e
+                        ))
+                    })?;
+                    messages_map = refetched.into_iter().map(|m| (m.id.clone(), m)).collect();
```
erskingardner, line 110: This is where there is a conflict - but we're currently adding pagination to messages so it's likely to change under you again here. @dannym-arx thoughts?

src/whitenoise/streaming.rs
```diff
@@ -134,19 +111,43 @@ impl Whitenoise {
...
         })
     }
 
+    #[cfg(test)]
     fn drain_user_updates(
-        mut initial_user: User,
+        mut initial_user: crate::whitenoise::users::User,
         updates: &mut broadcast::Receiver<user_streaming::UserUpdate>,
-    ) -> Result<User> {
+    ) -> Result<crate::whitenoise::users::User> {
```
dannym-arx: this looks less readable than before, I'd say we should keep the import

src/whitenoise/signer.rs
```diff
@@ -118,8 +118,13 @@ impl Whitenoise {
     ///
     /// Returns an error if no signer is available for the account.
     pub(crate) fn get_signer_for_account(&self, account: &Account) -> Result<Arc<dyn NostrSigner>> {
-        // First check for a registered external signer
-        if let Some(external_signer) = self.get_external_signer(&account.pubkey) {
+        // Only use the external-signer path when the account is actually typed
+        // as External. A stale map entry must not hijack a local-key account,
+        // and a missing map entry must not silently fall through to local keys
+        // for an External account.
+        if account.uses_external_signer()
```
erskingardner: This is subtle and I had to poke the agent a few times to get clear on it myself but in the case we had (past tense) a local key signed in, someone signs out (but the signout doesn't go as planned) then we still could have a case where the external signer is registered but we keep using the local signer. 

> 
>  Yeah — it protects half of the problem, not all of it.
> 
>  The new code does this:
> 
>  ```rust
>    if account.uses_external_signer()
>        && let Some(external_signer) = self.get_external_signer(&account.pubkey)
>    {
>        return Ok(external_signer);
>    }
> 
>    // otherwise fall back to local keys
>    let keys = self.secrets_store.get_nostr_keys_for_pubkey(&account.pubkey)?;
>  ```
> 
>
>  So the behavior is:
> 
>  - Local account + stale external signer entry → now uses local keys.
>  ✅ fixed
>  - External account + registered external signer → uses external signer.
>  ✅ fine
>  - External account + no registered external signer + stale local key still present → falls through to local keys.
>  ❌ still wrong
> 
>  That last case is the one I meant.
> 
>  To fully close it, the logic needs to be:
> 
>  ```rust
>    if account.uses_external_signer() {
>        return self
>            .get_external_signer(&account.pubkey)
>            .ok_or(...missing external signer...);
>    }
> 
>    // only local accounts may use secrets_store
>    let keys = self.secrets_store.get_nostr_keys_for_pubkey(&account.pubkey)?;
>  ```
> 
>  Why I think this matters: the codebase itself treats local-key removal for external accounts as best effort (let _ = remove_private_key_for_pubkey(...)), so a stale local key is explicitly possible. In that state, the current PR would still let an External account silently
>  operate with local keys.
> 
>  So: your PR fixed “stale external signer hijacks local account”, but not “external account silently falls back to local keys when signer is missing.”

src/whitenoise/streaming.rs
```diff
@@ -260,7 +271,17 @@ impl Whitenoise {
...
+                        target: "whitenoise",
+                        "subscription drain lagged by {n} messages; rebuilding snapshot from storage"
+                    );
+                    // Some updates were dropped — rebuild the snapshot from the DB
+                    // so the "initial snapshot + subsequent updates" guarantee holds.
+                    let refetched = self.get_archived_chat_list(account).await?;
+                    items_map = refetched
+                        .into_iter()
+                        .map(|item| (item.mls_group_id.clone(), item))
+                        .collect();
```
dannym-arx, line 315: this code is a 1 to 1 duplication of the code on 210, could you perhaps move it to a separate function?

src/whitenoise/accounts/login_external_signer.rs
```diff
@@ -132,41 +151,85 @@ impl Whitenoise {
...
+                    relay_type,
+                );
+            }
+        }
+
+        // Publish all missing relay lists concurrently — same pattern as
+        // publish_relay_lists_with_signer. Only NIP-65 failure is fatal;
+        // 10050/10051 failures are demoted to warnings.
+        let signer_arc = std::sync::Arc::new(signer.clone());
+        let nip65_publish = async {
```
erskingardner, line 177: From the agent - we just need to do a quick fetch again - the way we do with local signers. 

> login_external_signer_publish_default_relays has the same stash/DB divergence bug that you fixed in the local-key flow. After an incomplete login_external_signer_with_custom_relay retry, pending_logins can preserve relay lists from earlier probes while try_discover_relay_lists
>  has already synced the DB to the most recent probe. Because this loop skips writes for discovered.found(...), external login can complete with preserved NIP-65/Inbox/KeyPackage relays missing from the DB. It needs the same sync_discovered_relay_lists(&account, &discovered) step
>  before filling defaults.

src/whitenoise/streaming.rs
dannym-arx: I think that fixing the conflict will be enough, I don't think that this section of the code is going to change

erskingardner: But do you think that this needs to use pagination when it does the rebuild/refetch on lag?

dannym-arx: probably not, unless the lag is larger than the default page size (50)

src/whitenoise/signer.rs
```diff
@@ -51,7 +51,7 @@ impl Whitenoise {
         // Rebuild account subscriptions now that signing/decryption is available.
         // Only inbox_relays is needed by setup_subscriptions; do not gate
         // recovery on the nip65 lookup which is not used here.
-        match account.inbox_relays(self).await {
+        match account.effective_inbox_relays(self).await {
```
erskingardner, line 54: Either the comment needs to be changed or the method. 

> The comment was written to justify using `inbox_relays` (bare, no NIP-65 fallback) rather than
> `effective_inbox_relays`. The PR silently changes the call to `effective_inbox_relays` — which does
> a NIP-65 lookup when no inbox relays are found — without updating the comment. The comment now
> actively contradicts the code: it says "do not gate recovery on the nip65 lookup" but the code
> does exactly that via the fallback inside `effective_inbox_relays`.

src/whitenoise/streaming.rs
```diff
@@ -226,68 +280,53 @@ impl Whitenoise {
...
+                    tracing::warn!(
+                        target: "whitenoise",
+                        "subscription drain lagged by {n} messages; rebuilding snapshot from storage"
+                    );
+                    let refetched = if archived {
+                        self.get_archived_chat_list(account).await?
+                    } else {
+                        self.get_chat_list(account).await?
+                    };
+                    *items_map = refetched
```
erskingardner, line 312: This is a narrow race window sort of issue... I think it's likely worth looking into though. 

> When `Lagged` is handled by a full DB refetch, execution then `continue`s back to the top of the
> loop and processes the next buffered update. If that next update is a delta that arrived _after_
> the DB snapshot was taken, the delta is correctly applied on top of the snapshot — fine. But if
> that next update is for an item that the DB snapshot _already_ includes at a newer version, the
> buffered update (which could be stale relative to the DB read) overwrites the fresh DB row in
> `items_map`. The original non-lag path didn't have this issue because it read from the DB first
> and then applied updates. This is a narrow race window, but it's a class of bug that the
> subscribe-then-fetch design was meant to eliminate. Worth a careful look at whether the DB read
> truly happens-after all updates that are currently buffered.

src/whitenoise/accounts/login_multistep.rs
```diff
@@ -402,10 +411,11 @@ impl Whitenoise {
                     e
                 );
             }
-            account
-                .delete(&self.database)
-                .await
-                .map_err(LoginError::from)?;
+            if let Err(e) = account.delete(&self.database).await {
+                // Restore the stash so a retry of login_cancel can succeed.
+                self.pending_logins.insert(stash.0, stash.1);
```
erskingardner: I really don't like tuple indices like this. Can you destructure to named vars on line 386 above instead so it's more obvious? 

`let Some((stash_pubkey, stash_discovered)) = self.pending_logins.remove(pubkey) else { ... }`

erskingardner:
@sergey3bv failing checks still here. I'm going to close for now. Feel free to reopen if you want to pick this back up. tbh, i'd prefer to see fixes like these each in their own PR instead of bundled together.

PR #688, opened 2026-03-26
MIP-05 PR1: add push notification registration persistence

erskingardner:
## Summary
- add the new `push_notifications` module with `PushPlatform`, `PushRegistration`, `GroupPushToken`, and app-facing `Whitenoise` APIs for registration lookup, upsert, and clear
- add local persistence for device registrations and per-group leaf-index token cache via new `push_registrations` and `group_push_tokens` tables plus database access modules
- add unit tests covering registration upsert/replacement, clear semantics, token cache scoping, and the explicit behavior that disabling notifications does not clear stored registration state

## Out Of Scope For This PR
- MLS token gossip (`447` / `448` / `449`)
- notification request publication (`446 -> 1059`)
- refresh / cleanup hardening

## Verification
- `just fmt`
- `just precommit-quick`
  - `fmt`, `docs`, and `clippy` pass
  - unit tests are blocked by an unrelated existing failure in `whitenoise::scheduled_tasks::tasks::relay_list_maintenance::tests::test_execute_is_noop_when_relay_lists_already_present`
  - the same failing test reproduces on clean `master` in `/Users/jeff/code/whitenoise-rs`, so this branch is not introducing that red

## Summary by CodeRabbit

* **New Features**
  * Added push notification infrastructure supporting APNS and FCM platforms.
  * Implemented account-level push registration management for device handling.
  * Added group-level push token management for notification delivery.
  * Push registration state persists independently of notification settings.

### DIFF
--- a/db_migrations/0039_push_notifications.sql
+++ b/db_migrations/0039_push_notifications.sql
@@ -26,7 +26,9 @@ CREATE TABLE group_push_tokens (
     leaf_index INTEGER NOT NULL CHECK (leaf_index >= 0),
     server_pubkey TEXT NOT NULL,
     relay_hint TEXT,
-    encrypted_token TEXT NOT NULL CHECK (length(encrypted_token) > 0),
+    encrypted_token TEXT NOT NULL CHECK (
+        length(trim(encrypted_token, ' ' || char(9) || char(10) || char(13))) > 0
+    ),
     created_at INTEGER NOT NULL,
     updated_at INTEGER NOT NULL,
     FOREIGN KEY (account_pubkey) REFERENCES accounts(pubkey) ON DELETE CASCADE
--- a/src/whitenoise/database/group_push_tokens.rs
+++ b/src/whitenoise/database/group_push_tokens.rs
@@ -1,5 +1,7 @@
 //! Database operations for cached per-group push tokens.
 
+use core::fmt;
+
 use chrono::{DateTime, Utc};
 use mdk_core::prelude::GroupId;
 use nostr_sdk::{PublicKey, RelayUrl};
@@ -8,7 +10,6 @@ use super::{Database, utils::parse_timestamp};
 use crate::perf_instrument;
 use crate::whitenoise::push_notifications::GroupPushToken;
 
-#[derive(Debug)]
 struct GroupPushTokenRow {
     account_pubkey: PublicKey,
     mls_group_id: GroupId,
@@ -18,6 +19,21 @@ struct GroupPushTokenRow {
     encrypted_token: String,
     created_at: DateTime<Utc>,
     updated_at: DateTime<Utc>,
+}
+
+impl fmt::Debug for GroupPushTokenRow {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("GroupPushTokenRow")
+            .field("account_pubkey", &self.account_pubkey)
+            .field("mls_group_id", &self.mls_group_id)
+            .field("leaf_index", &self.leaf_index)
+            .field("server_pubkey", &self.server_pubkey)
+            .field("relay_hint", &self.relay_hint)
+            .field("encrypted_token", &"<redacted>")
+            .field("created_at", &self.created_at)
+            .field("updated_at", &self.updated_at)
+            .finish()
+    }
 }
 
 impl<'r, R> sqlx::FromRow<'r, R> for GroupPushTokenRow
@@ -413,5 +429,46 @@ mod tests {
         assert_eq!(updated.created_at, expected_created_at);
         assert_eq!(updated.server_pubkey, second_server);
         assert_eq!(updated.encrypted_token, "ciphertext-two");
+    }
+
+    #[tokio::test]
+    async fn test_group_push_tokens_upsert_rejects_whitespace_only_encrypted_token() {
+        let (whitenoise, _data_temp, _logs_temp) = create_mock_whitenoise().await;
+        let account = whitenoise.create_identity().await.unwrap();
+        let mls_group_id = make_group_id(44);
+        let server_pubkey = Keys::generate().public_key();
+
+        let error = GroupPushToken::upsert(
+            &account.pubkey,
+            &mls_group_id,
+            1,
+            &server_pubkey,
+            None,
+            " \n\t\r ",
+            &whitenoise.database,
+        )
+        .await
+        .unwrap_err();
+
+        assert!(matches!(error, sqlx::Error::Database(_)));
+    }
+
+    #[test]
+    fn test_group_push_token_row_debug_redacts_encrypted_token() {
+        let row = GroupPushTokenRow {
+            account_pubkey: Keys::generate().public_key(),
+            mls_group_id: make_group_id(55),
+            leaf_index: 2,
+            server_pubkey: Keys::generate().public_key(),
+            relay_hint: Some(RelayUrl::parse("wss://push.example.com").unwrap()),
+            encrypted_token: "ciphertext-value".to_string(),
+            created_at: Utc::now(),
+            updated_at: Utc::now(),
+        };
+
+        let debug_output = format!("{row:?}");
+
+        assert!(debug_output.contains("<redacted>"));
+        assert!(!debug_output.contains("ciphertext-value"));
     }
 }
--- a/src/whitenoise/database/push_registrations.rs
+++ b/src/whitenoise/database/push_registrations.rs
@@ -1,5 +1,6 @@
 //! Database operations for local push registrations.
 
+use core::fmt;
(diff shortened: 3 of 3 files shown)


db_migrations/0039_push_notifications.sql
```diff
@@ -0,0 +1,32 @@
...
+CREATE INDEX idx_push_registrations_server_pubkey
+    ON push_registrations(server_pubkey);
+
+CREATE TABLE group_push_tokens (
+    account_pubkey TEXT NOT NULL,
+    group_id BLOB NOT NULL,
+    leaf_index INTEGER NOT NULL CHECK (leaf_index >= 0),
+    server_pubkey TEXT NOT NULL,
+    relay_hint TEXT,
+    encrypted_token TEXT NOT NULL CHECK (length(encrypted_token) > 0),
```
jgmontoya: Not sure if a real concern (application code prevents it) but a pure-blank/whitespace string passes this validation. We could use `CHECK (trim(raw_token) != '')` to match the rust validation exactly

db_migrations/0039_push_notifications.sql
```diff
@@ -0,0 +1,32 @@
...
+    updated_at INTEGER NOT NULL,
+    PRIMARY KEY (account_pubkey, group_id, leaf_index),
+    FOREIGN KEY (account_pubkey) REFERENCES accounts(pubkey) ON DELETE CASCADE
+);
```
jgmontoya, line 35: I think this tables should have an autoincrement id as primary key to follow the project's conventions

jgmontoya, line 35: In the future, when we have multi-device, we might want to have multiple tokens registered to a single account

db_migrations/0039_push_notifications.sql
```diff
@@ -0,0 +1,32 @@
...
+    last_shared_at INTEGER,
+    FOREIGN KEY (account_pubkey) REFERENCES accounts(pubkey) ON DELETE CASCADE
+);
+
+CREATE INDEX idx_push_registrations_server_pubkey
+    ON push_registrations(server_pubkey);
+
+CREATE TABLE group_push_tokens (
+    account_pubkey TEXT NOT NULL,
+    group_id BLOB NOT NULL,
```
jgmontoya: we've used `mls_group_id` in other places (for example, in the `group_information` table, I'd use the same to avoid confusion

src/whitenoise/database/group_push_tokens.rs
```diff
@@ -0,0 +1,357 @@
...
+    pub(crate) async fn upsert(
+        account_pubkey: &PublicKey,
+        group_id: &GroupId,
+        leaf_index: u32,
+        server_pubkey: &PublicKey,
+        relay_hint: Option<&RelayUrl>,
+        encrypted_token: &str,
+        database: &Database,
+    ) -> Result<Self, sqlx::Error> {
+        let _span = perf_span!("db::group_push_tokens_upsert");
```
jgmontoya: I think we should use `#[perf_instrument("db::...")]` on the methods rather than this

src/whitenoise/database/group_push_tokens.rs
```diff
@@ -0,0 +1,357 @@
...
+        Ok(row.into())
+    }
+
+    pub(crate) async fn delete(
+        account_pubkey: &PublicKey,
+        group_id: &GroupId,
+        leaf_index: u32,
+        database: &Database,
+    ) -> Result<bool, sqlx::Error> {
+        let _span = perf_span!("db::group_push_tokens_delete");
```
jgmontoya: same here

src/whitenoise/database/group_push_tokens.rs
```diff
@@ -0,0 +1,357 @@
...
+
+        Ok(result.rows_affected() > 0)
+    }
+
+    pub(crate) async fn find_by_account_and_group(
+        account_pubkey: &PublicKey,
+        group_id: &GroupId,
+        database: &Database,
+    ) -> Result<Vec<Self>, sqlx::Error> {
+        let _span = perf_span!("db::group_push_tokens_find_by_group");
```
jgmontoya: and here

src/whitenoise/database/group_push_tokens.rs
```diff
@@ -0,0 +1,357 @@
...
+        .await?;
+
+        Ok(rows.into_iter().map(Into::into).collect())
+    }
+
+    pub(crate) async fn group_ids_for_account(
+        account_pubkey: &PublicKey,
+        database: &Database,
+    ) -> Result<Vec<GroupId>, sqlx::Error> {
+        let _span = perf_span!("db::group_push_tokens_group_ids_for_account");
```
jgmontoya: and here

src/whitenoise/database/push_registrations.rs
```diff
@@ -0,0 +1,415 @@
...
+        }
+    }
+}
+
+impl PushRegistration {
+    pub(crate) async fn find_by_account_pubkey(
+        account_pubkey: &PublicKey,
+        database: &Database,
+    ) -> Result<Option<Self>, sqlx::Error> {
+        let _span = perf_span!("db::push_registrations_find");
```
jgmontoya: and here

src/whitenoise/database/push_registrations.rs
```diff
@@ -0,0 +1,415 @@
...
+
+    pub(crate) async fn upsert(
+        account_pubkey: &PublicKey,
+        platform: PushPlatform,
+        raw_token: &str,
+        server_pubkey: &PublicKey,
+        relay_hint: Option<&RelayUrl>,
+        database: &Database,
+    ) -> Result<Self, sqlx::Error> {
+        let _span = perf_span!("db::push_registrations_upsert");
```
jgmontoya: and here

src/whitenoise/database/push_registrations.rs
```diff
@@ -0,0 +1,415 @@
...
+        .await?;
+
+        Ok(row.into())
+    }
+
+    pub(crate) async fn delete_by_account_pubkey(
+        account_pubkey: &PublicKey,
+        database: &Database,
+    ) -> Result<bool, sqlx::Error> {
+        let _span = perf_span!("db::push_registrations_delete");
```
jgmontoya: and here

src/whitenoise/push_notifications.rs
```diff
@@ -0,0 +1,305 @@
...
+
+impl FromStr for PushPlatform {
+    type Err = String;
+
+    fn from_str(value: &str) -> std::result::Result<Self, Self::Err> {
+        match value.to_ascii_lowercase().as_str() {
+            "apns" => Ok(Self::Apns),
+            "fcm" => Ok(Self::Fcm),
+            _ => Err(format!("Invalid push platform: {value}")),
+        }
```
jgmontoya: do we need the `to_ascii_lowercase()`? I think the only values that can get here come from the db which has `CHECK (platform IN ('apns', 'fcm'))`

Also, if a value failed here because of casing, the error message would still show the lowercase value which could be confusing

db_migrations/0039_push_notifications.sql
```diff
@@ -0,0 +1,32 @@
...
+    created_at INTEGER NOT NULL,
+    updated_at INTEGER NOT NULL,
+    last_shared_at INTEGER,
+    FOREIGN KEY (account_pubkey) REFERENCES accounts(pubkey) ON DELETE CASCADE
+);
+
+CREATE INDEX idx_push_registrations_server_pubkey
+    ON push_registrations(server_pubkey);
+
+CREATE TABLE group_push_tokens (
```
jgmontoya, line 22: This table is missing a `created_at` field

src/whitenoise/database/push_registrations.rs
```diff
@@ -0,0 +1,415 @@
...
+use nostr_sdk::{PublicKey, RelayUrl};
+
+use super::{
+    Database,
+    utils::{parse_optional_timestamp, parse_timestamp},
+};
+use crate::perf_span;
+use crate::whitenoise::push_notifications::{PushPlatform, PushRegistration};
+
+#[derive(Debug)]
```
jgmontoya: Not sure if a big deal or not: should we be mindful and avoid logging the raw_token or is it not really that important?

db_migrations/0039_push_notifications.sql
erskingardner: This is local to the device though - push_registrations are just your own. you should only have one per account on each device (which is also why there is a lack of auto-incrementing id). I can probably get behind the auto-incrementing id + an index to enforce uniqueness though simply to stick with convention.

erskingardner: nevermind - i was thinking about the other table.

PR #690, opened 2026-03-26
Adding fail early if tests run without required docker containers

sergey3bv:
Trying to improve developer experience if tests happened to run without required docker containers. @erskingardner, @jgmontoya, could you please take a look

## Summary by CodeRabbit

* **New Features**
  * Added a centralized preflight check that verifies required local test services are reachable and shows an actionable hint if not.

* **Refactor**
  * Removed scattered relay-readiness logic in favor of the centralized preflight routine; readiness is now consistently enforced early in startup.

* **Tests**
  * Integration tests and benchmarks run the preflight at startup and will fail fast with clear guidance when local services are unavailable.

### DIFF
--- a/src/whitenoise/mod.rs
+++ b/src/whitenoise/mod.rs
@@ -40,6 +40,8 @@ pub mod message_aggregator;
 pub mod message_streaming;
 pub mod messages;
 pub mod notification_streaming;
+#[cfg(any(test, feature = "integration-tests", feature = "benchmark-tests"))]
+pub(crate) mod preflight;
 pub mod push_notifications;
 pub mod relays;
 pub mod scheduled_tasks;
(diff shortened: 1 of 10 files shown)


src/whitenoise/mod.rs
erskingardner, line 790: If I'm not mistaken, I think we need your preflight call here too. Many of our unit tests call these mock whitenoise methods still.

src/whitenoise/preflight.rs
```diff
@@ -0,0 +1,59 @@
...
+    for endpoint in endpoints {
+        let deadline = Instant::now() + RETRY_WINDOW;
+        let mut last_error = "timed out".to_string();
+        let mut connected = false;
+
+        loop {
+            let result =
+                tokio::time::timeout(CONNECTION_TIMEOUT, TcpStream::connect(endpoint)).await;
+            match result {
+                Ok(Ok(_)) => {
```
erskingardner: For relays - we should not only check that we can connect but that the connection successfully upgrades to a websocket connection. obviously for blossom that doesn't apply. The previous checks waited for the websocket upgrade.

src/whitenoise/mod.rs
```diff
@@ -751,6 +736,8 @@ pub mod test_utils {
     use nostr_sdk::{Keys, PublicKey, RelayUrl};
     use tempfile::TempDir;
 
+    const REQUIRED_TEST_RELAYS: &[&str] = &["127.0.0.1:8080", "127.0.0.1:7777"];
```
erskingardner, line 765: We should also include the blossom `:3000` endpoint here I think.

erskingardner:
@sergey3bv sorry but I'm actually not really seeing that this PR is very additive to the developer experience - and at the cost of several hundred lines of extra code. Maybe I'm misunderstanding what you were trying to do originally but I'm leaning towards closing this PR.

erskingardner:
> Well, I wanted to fix the issue I have ran into myself: if docker containers are not running, tests will stale indefinitely. If the issue is not worth solving, I am fine closing it.

It's just that the cure is worse than the disease. Asking Claude gave me this suggestion: 

> The simplest approach: add a **TCP port probe with a ~1-2 second timeout** at the earliest point in test setup, before any real work happens.
> 
> The services listen on three ports:
> - `localhost:8080` (nostr-rs-relay)
> - `localhost:7777` (strfry)  
> - `localhost:3000` (blossom)
> 
> Right now, tests have 30-60 second timeouts baked in (daemon startup waits 30s, `wait_for_group` 60s, etc.), so without Docker they'll just hang for ages before failing.
> 
> **Most surgical option:** Add a quick connectivity check in the two test entry points:
> 
> 1. **Integration tests** — at the top of `src/bin/integration_test.rs` before building `WhitenoiseConfig`, do a `TcpStream::connect_timeout` to each relay/blossom port with a 1-2s timeout. If any fail, bail with a clear message like `"Docker services not running — run 'docker compose up -d' first"`.
> 
> 2. **CLI E2E tests** — in `tests/support/cli.rs` inside `DaemonUnderTest::start()` (line ~33 area), same check before spawning the daemon binary.
> 
> That's ~10-15 lines of code total, no new dependencies (just `std::net::TcpStream`), and it turns a multi-minute hang into a sub-2-second failure with an actionable error message. No changes to any actual test logic or timeouts needed.

PR #691, opened 2026-03-26
feat: add mute/unmute chat support

### LINKED ISSUE #687: Mute/Unmute Chat
### What it does
Suppresses notifications for a chat. Messages still arrive and appear in the conversation — the user just doesn't get push notifications. Local-only, no MDK/protocol involvement.

### Durations
1 hour, 8 hours, 1 day, 1 week, Always (until manually unmuted)

### whitenoise-rs
- **DB migration:** Add `muted_until INTEGER DEFAULT NULL` to `accounts_groups` — `NULL` = not muted, timestamp = muted until, far-future = muted forever
- **Core logic:** `mute_chat(account, group_id, until)` and `unmute_chat(account, group_id)` — follows the same idempotent pattern as `archive_chat()`
- **Notification suppression:** Check `is_muted()` before emitting push notifications in `emit_new_message_notification()` — messages still delivered, just no notification
- **Stream update:** New `ChatListUpdateTrigger::ChatMuteChanged` so the UI reacts in real time
- **ChatListItem:** Add `muted_until` field so Flutter knows the mute state

### Bridge
- Expose `mute_chat()` / `unmute_chat()` functions
- Add `mutedUntil` to `ChatSummary`
- Add `ChatListUpdateTrigger::ChatMuteChanged`

### Flutter
- Mute/unmute action in chat context menu
- Duration picker bottom sheet (1h, 8h, 1d, 1w, Always)
- Mute icon on chat list tile for muted chats
- Handle `chatMuteChanged` stream trigger

### Mute expiry
Checked at read time — `is_muted()` compares `muted_until` against current time. No background timers needed. Expired mutes naturally stop suppressing notifications; the UI self-corrects on next render.

mubarakcoded:
## Summary

Implements mute/unmute chat functionality (closes #687).

- **Migration 0040**: Adds `muted_until` column to `accounts_groups` (nullable timestamp, supports timed + forever mutes via `MUTE_FOREVER` sentinel year 9999)
- **Centralized `MuteDuration` enum** in `accounts_groups.rs` with preset variants (`OneHour`, `EightHours`, `OneDay`, `OneWeek`, `Forever`) and a `Custom(DateTime<Utc>)` escape hatch for arbitrary expiry timestamps. Single source of truth — CLI re-exports via `pub use`, library consumers import from `src/lib.rs`.
- **Domain layer**: `is_muted()`, `is_muted_forever()`, `mute()`, `unmute()` helpers on `AccountGroup`; `mute_chat()` takes `MuteDuration`, calls `to_expiry()` internally, validates `Custom` timestamps are in the future
- **Notification suppression**: Muted chats skip `emit_new_message_notification` with fail-open semantics (DB error → not muted)
- **Scheduled task**: `MuteExpiryCleanup` runs every 60s, clears expired `muted_until` rows and emits `ChatMuteChanged` triggers
- **Streaming**: New `ChatMuteChanged` trigger + `muted_until` field on `ChatListItem` for real-time UI updates
- **CLI**: `MuteChat`/`UnmuteChat` protocol variants, `Mute`/`Unmute` subcommands with `MuteDuration` (1h/8h/1d/1w/forever), serde aliases for `24h`/`7d`/`one_day`/`one_week`
- **Merged duplicate DB queries** in notification streaming per review feedback
- **18 new unit tests** across domain, notifications, CLI, and scheduler layers

## Design decisions

- `MuteDuration` lives in the domain (`accounts_groups.rs`) not the CLI — `to_expiry()` is called inside `Whitenoise::mute_chat()`, so callers never deal with raw timestamps
- `Custom(DateTime<Utc>)` variant allows arbitrary expiry while keeping the typed API — past timestamps are rejected with `InvalidInput`
- `save()` intentionally excludes `muted_until` to prevent giftwrap re-processing from clobbering user preferences (same pattern as `archived_at`, `removed_at`)
- `mute_chat()` always overwrites (not idempotent) per agreed feedback
- `unmute_chat()` has no early-return guard — always clears DB to handle stale expired timestamps
- Notification filtering is fail-open: DB lookup error → treat as not muted (don't drop real notifications)

## Test plan

- [x] `just precommit-quick` passes (fmt, docs, clippy, tests)
- [x] `just precommit` with Docker for full integration tests
- [ ] Manual verification of mute/unmute via CLI
- [x] Flutter app integration after whitenoise crate update

### DIFF
--- a/src/whitenoise/accounts_groups.rs
+++ b/src/whitenoise/accounts_groups.rs
@@ -48,7 +48,7 @@ pub struct AccountGroup {
     pub removed_at: Option<DateTime<Utc>>,
     /// When this chat's mute expires.
     /// - `None` = not muted (notifications enabled)
-    /// - `Some(timestamp)` = muted until that time; use [`MUTE_FOREVER`] for
+    /// - `Some(timestamp)` = muted until that time; use `MUTE_FOREVER` for
     ///   "always muted until manually unmuted"
     pub muted_until: Option<DateTime<Utc>>,
     pub created_at: DateTime<Utc>,
@@ -57,7 +57,7 @@ pub struct AccountGroup {
 
 /// Sentinel timestamp representing "muted forever" (9999-12-31T23:59:59Z).
 /// Use this instead of crafting ad-hoc far-future timestamps.
-pub const MUTE_FOREVER: DateTime<Utc> = {
+pub(crate) const MUTE_FOREVER: DateTime<Utc> = {
     match DateTime::<Utc>::from_timestamp(253_402_300_799, 0) {
         Some(dt) => dt,
         None => unreachable!(),
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -53,7 +53,7 @@ pub use whitenoise::account_settings::AccountSettings;
 pub use whitenoise::app_settings::{AppSettings, Language, ThemeMode};
 
 // Groups and relays
-pub use whitenoise::accounts_groups::{AccountGroup, MUTE_FOREVER, MuteDuration};
+pub use whitenoise::accounts_groups::{AccountGroup, MuteDuration};
 pub use whitenoise::group_information::{GroupInformation, GroupType};
 pub use whitenoise::groups::{GroupWithInfoAndMembership, GroupWithMembership};
 pub use whitenoise::relays::{Relay, RelayType};


src/cli/protocol.rs
```diff
@@ -1,7 +1,65 @@
...
 
+use chrono::{DateTime, Duration, Utc};
 use serde::{Deserialize, Serialize};
 
+use crate::whitenoise::accounts_groups::MUTE_FOREVER;
+
+/// Mute duration for the `mute_chat` CLI command.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum MuteDuration {
```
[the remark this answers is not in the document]
mubarakcoded: fixed

src/whitenoise/notification_streaming/mod.rs
```diff
@@ -51,6 +51,15 @@ impl Whitenoise {
             return;
         }
 
+        if self.is_chat_muted(&account.pubkey, group_id).await {
```
jgmontoya: On the previous block, we check for `!self.is_group_accepted` which queries the db (`AccountGroup::find_by_account_and_group`), then this method does it again. We should extract both to a new private method, that queries once and then checks for both fields so we don't have to query multiple times for the same record

src/whitenoise/accounts_groups.rs
```diff
@@ -86,6 +100,19 @@ impl AccountGroup {
...
+    pub fn is_muted(&self) -> bool {
+        match self.muted_until {
+            Some(until) => Utc::now() < until,
+            None => false,
+        }
+    }
+
+    /// Returns true if this chat is muted forever (until manually unmuted).
+    pub fn is_muted_forever(&self) -> bool {
+        self.muted_until == Some(MUTE_FOREVER)
```
jgmontoya: Perhaps this should be a `>=` instead of an `==`

jgmontoya (changes requested):

db_migrations/0040_add_muted_until_to_accounts_groups.sql
erskingardner, line 1: I just merged my PR with a 0039 migration - I think you'll need to update this one after a rebase.

erskingardner (approved):
I'm not mad about the slight mixing of types with strings, dates, and enums for durations but overall I think this is fine to merge. FWIW, I think I would have preferred to see a centralized concept of MuteDuration which had helper methods that would be used by the public API and the CLI. This would have given us a single place to reason about what default durations are available and we handle custom mute durations.

src/whitenoise/accounts_groups.rs
```diff
@@ -43,10 +46,93 @@ pub struct AccountGroup {
...
+    /// - `Some(timestamp)` = muted until that time; use [`MUTE_FOREVER`] for
+    ///   "always muted until manually unmuted"
+    pub muted_until: Option<DateTime<Utc>>,
     pub created_at: DateTime<Utc>,
     pub updated_at: DateTime<Utc>,
 }
 
+/// Sentinel timestamp representing "muted forever" (9999-12-31T23:59:59Z).
+/// Use this instead of crafting ad-hoc far-future timestamps.
+pub const MUTE_FOREVER: DateTime<Utc> = {
```
jgmontoya: Why do we need to have this public?

PR #693, opened 2026-03-27
fix: recover missing key package state on startup

erskingardner:
## Summary
- teach key package maintenance to only treat live, locally tracked packages as healthy and republish when restored sessions have no usable local key package state
- add a regression test for the missing-local-key-material case
- make the relay-list maintenance no-op test wait for background relay-list publishes so it stops flaking

## Testing
- just precommit-quick

Fixes marmot-protocol/whitenoise#491

## Summary by CodeRabbit

* **Bug Fixes**
  * Key-package maintenance now checks for live tracked key material before deleting; when no live local state exists it will republish to avoid accidental loss.

* **Tests**
  * Replaced fixed sleeps with retry/polling for reliable async convergence; added regression tests for republish scenarios and improved relay discovery polling.
  * Added a test-only helper to insert/track published key packages to stabilize integration tests.

### DIFF
--- a/src/integration_tests/test_cases/shared/create_accounts.rs
+++ b/src/integration_tests/test_cases/shared/create_accounts.rs
@@ -1,6 +1,9 @@
+use std::time::Duration;
+
+use async_trait::async_trait;
+
 use crate::WhitenoiseError;
 use crate::integration_tests::core::*;
-use async_trait::async_trait;
 
 pub struct CreateAccountsTestCase {
     account_names: Vec<String>,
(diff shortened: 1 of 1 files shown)


src/integration_tests/test_cases/scheduler/key_package_maintenance.rs
```diff
@@ -185,6 +237,11 @@ async fn publish_backdated_key_package(
     client.send_event(&event).await?;
     client.disconnect().await;
 
+    context
+        .whitenoise
+        .track_published_key_package_for_testing(&account.pubkey, &hash_ref, &event_id.to_hex())
```
mubarakcoded, line 242: Just making sure I understand correctly - publish_backdated_key_package now correctly calls track_published_key_package_for_testing so maintenance treats the package as locally owned. But publish_outdated_key_package (used in other tests) doesn't call it. Is that intentional? If the outdated package isn't in the PublishedKeyPackage table, find_live_published_key_packages returns None for it, and it will not be counted as live local state. This seems fine, as outdated packages are handled in a separate branch before the live-state check. Just confirming the intent rather than assuming.

erskingardner: Yes, that’s intentional. `publish_backdated_key_package()` creates a still-valid key package with a custom old timestamp, so we need to call `track_published_key_package_for_testing()` so maintenance recognizes it as locally owned/live state. 
 
`publish_outdated_key_package()` is different: it intentionally creates a legacy/outdated package without the encoding tag, and maintenance handles those in the earlier outdated-package branch before any `find_live_published_key_packages()` check. In that branch, outdated packages are excluded from the live-state lookup, so not inserting them into `PublishedKeyPackage` does not affect the behavior under test.

So the asymmetry is deliberate: backdated valid packages need tracking; outdated packages don’t for these test paths.

PR #694, opened 2026-03-28
Implement MIP-05 token gossip orchestration

erskingardner:
## Summary
- bump MDK to the upstream MIP-05-ready master ref and enable the `mdk-core` `mip05` feature
- implement PR2 MIP-05 token gossip orchestration in `whitenoise-rs`, including incoming/outgoing `447/448/449`, welcome-triggered sharing, notification preference hooks, delayed in-memory `447 -> 448` suppression, and stale leaf pruning on commits
- harden post-accept token-update handling so cache reconciliation failures log instead of surfacing after MLS state has already advanced, and attempt share/remove publishes across all active groups before reporting partial failures
- expand tests for token exchange, fallback suppression, welcome-triggered sharing, and multi-group share/remove behavior, plus small follow-up cleanup in helper naming/duplication

## Testing
- `just precommit-quick`

## Summary by CodeRabbit

* **New Features**
  * MIP-05 token support added for improved push-notification workflows
  * Push token sharing, delayed responses, and per-group token reconciliation when joining/leaving groups or updating registrations
  * Automatic publishing/removal of tokens based on notification settings and registration changes

* **Bug Fixes**
  * Improved error handling with non-fatal warnings for push token sync and welcome flows
  * More robust handling of token message types to prevent unintended caching or dispatching

### DIFF
--- a/src/whitenoise/push_notifications.rs
+++ b/src/whitenoise/push_notifications.rs
@@ -846,10 +846,11 @@ impl Whitenoise {
         group_id: &GroupId,
         token_tag: Option<&TokenTag>,
     ) -> Result<()> {
+        let mdk = self.create_mdk_for_account(account.pubkey)?;
+        let leaf_index = mdk.own_leaf_index(group_id)?;
+
         match token_tag {
             Some(token_tag) => {
-                let mdk = self.create_mdk_for_account(account.pubkey)?;
-                let leaf_index = mdk.own_leaf_index(group_id)?;
                 GroupPushToken::upsert(
                     &account.pubkey,
                     group_id,
@@ -863,8 +864,6 @@ impl Whitenoise {
                 .await?;
             }
             None => {
-                let mdk = self.create_mdk_for_account(account.pubkey)?;
-                let leaf_index = mdk.own_leaf_index(group_id)?;
                 GroupPushToken::delete(&account.pubkey, group_id, leaf_index, &self.database)
                     .await?;
             }


src/whitenoise/push_notifications.rs
```diff
@@ -129,23 +138,486 @@ impl Whitenoise {
...
+
+        if publish_failures.is_empty() {
+            Ok(())
+        } else {
+            Err(WhitenoiseError::Configuration(format!(
+                "failed to share push token to one or more groups: {}",
+                publish_failures.join(", ")
+            )))
+        }
+    }
```
dannym-arx, line 593: super minor, but maybe add a blank line after this function, like we do everywhere else?

src/whitenoise/event_processor/event_handlers/handle_mls_message.rs
```diff
@@ -1326,6 +1735,90 @@ mod tests {
...
+            whitenoise
+                .handle_giftwrap(member, giftwrap)
+                .await
+                .expect("member should process welcome successfully");
+        }
+
+        group_id
+    }
+
+    async fn wait_for_published_event_count(
```
dannym-arx: this function is duplicated in push_notifications.rs

src/whitenoise/event_processor/event_handlers/handle_mls_message.rs
```diff
@@ -1326,6 +1735,90 @@ mod tests {
...
+                return new_count;
+            }
+
+            tokio::time::sleep(std::time::Duration::from_millis(100)).await;
+        }
+
+        panic!("timed out waiting for new published event");
+    }
+
+    async fn count_published_events_for_account(
```
dannym-arx: this function is duplicated

src/whitenoise/push_notifications.rs
```diff
@@ -129,23 +138,486 @@ impl Whitenoise {
...
+
+    #[perf_instrument("push_notifications")]
+    pub(crate) async fn handle_received_push_group_message(
+        &self,
+        account: &Account,
+        message: &mdk_core::prelude::message_types::Message,
+        sender_leaf_index: Option<u32>,
+    ) -> Result<bool> {
+        if !Self::is_push_group_message_kind(message.kind) {
```
dannym-arx: did you mean to have an Ok(false) here? shouldn't we just let parse_group_message handle it and return an err?

src/whitenoise/push_notifications.rs
```diff
@@ -129,23 +138,486 @@ impl Whitenoise {
...
+            GroupPushToken::find_by_account_and_group(&account.pubkey, group_id, &self.database)
+                .await?;
+
+        for token in cached_tokens {
+            if active_leaf_map.contains_key(&token.leaf_index) {
+                continue;
+            }
+
+            GroupPushToken::delete(&account.pubkey, group_id, token.leaf_index, &self.database)
+                .await?;
```
dannym-arx: minor speed improvement; collect the stale leaf indices first, then run a single `DELETE WHERE leaf_index IN (...)`

something like: 

```rust
let stale_indices: Vec<u32> = cached_tokens
    .iter()
    .filter(|t| !active_leaf_map.contains_key(&t.leaf_index))
    .map(|t| t.leaf_index)
    .collect();
if !stale_indices.is_empty() {
    GroupPushToken::delete_batch(&account.pubkey, group_id, &stale_indices, &self.database).await?;
}

if !stale_indices.is_empty() {
    GroupPushToken::delete_batch(&account.pubkey, group_id, &stale_indices, &self.database).await?;
}
```

we don't have delete_batch of course, you would have to add it

src/whitenoise/push_notifications.rs
```diff
@@ -129,23 +138,486 @@ impl Whitenoise {
...
+                &account.pubkey,
+                mls_group_id,
+                token.leaf_index,
+                &token.token_tag.server_pubkey,
+                Some(&token.token_tag.relay_hint),
+                &token.token_tag.encrypted_token.to_base64(),
+                &self.database,
+            )
+            .await?;
+        }
```
dannym-arx: we should wrap this in an sqlx transaction because if any of those fail, it leaves the db in an incomplete state

src/whitenoise/push_notifications.rs
erskingardner: I kept Ok(false) here intentionally. This helper is being used as a discriminator for “is this one of the MIP-05 control message kinds we own?” rather than as a pure parser wrapper, so non-MIP-05 MLS application messages are a normal control-flow case, not an error. Letting parse_group_message drive that branch would turn expected non-push messages into failures and make the call site less clear.

src/whitenoise/push_notifications.rs
erskingardner: I left this as-is for now. If we do this (and I don't really think that we need to tbh) I’d prefer to treat the batch delete as a small follow-up.

src/whitenoise/push_notifications.rs
dannym-arx: makes sense to me

src/whitenoise/event_processor/event_handlers/handle_giftwrap.rs
```diff
@@ -343,6 +343,19 @@ impl Whitenoise {
             );
         }
 
+        if let Err(error) = whitenoise
+            .share_local_push_token_to_group(account, group_id)
+            .await
+        {
```
jgmontoya, line 350: I think this should be done in the background and not synchronously in the event processor

src/whitenoise/event_processor/event_handlers/handle_mls_message.rs
```diff
@@ -64,70 +64,100 @@ impl Whitenoise {
...
-            match message.kind {
-                Kind::ChatMessage => {
-                    let msg = self.cache_chat_message(&group_id, &message).await?;
-                    let group_name = mdk.get_group(&group_id).ok().flatten().map(|g| g.name);
-                    Whitenoise::spawn_new_message_notification_if_enabled(
-                        account, &group_id, &msg, group_name,
-                    );
-                    self.emit_message_update(&group_id, UpdateTrigger::NewMessage, msg);
-                    self.emit_chat_list_update(
+        if let Some((group_id, inner_event, message)) = Self::extract_message_details(&outcome) {
```
jgmontoya: The comment (`// Extract and store media references synchronously for application messages.`) is now far from the code and is making things confusing.

src/whitenoise/event_processor/event_handlers/handle_mls_message.rs
```diff
@@ -64,70 +64,100 @@ impl Whitenoise {
...
+                    let media_manager = mdk.media_manager(group_id.clone());
+                    self.media_files()
+                        .parse_imeta_tags_from_event(&inner_event, &media_manager)?
+                };
 
-                    for (trigger, msg) in self.cache_deletion(&group_id, &message).await? {
-                        self.emit_message_update(&group_id, trigger, msg);
+                self.media_files()
+                    .store_parsed_media_references(&group_id, &account.pubkey, parsed_references)
+                    .await?;
```
jgmontoya: I think part of the problem is this method now being too large, could we split it / refactor it so that it's easier to follow & understand, has less anidation, etc..?

src/whitenoise/push_notifications.rs
```diff
@@ -128,24 +137,592 @@ impl Whitenoise {
...
+            .bind(account.pubkey.to_hex())
+            .bind(mls_group_id.as_slice())
+            .bind(i64::from(token.leaf_index))
+            .bind(token.token_tag.server_pubkey.to_hex())
+            .bind(token.token_tag.relay_hint.to_string())
+            .bind(token.token_tag.encrypted_token.to_base64())
+            .bind(now)
+            .bind(now)
+            .execute(&mut *tx)
+            .await?;
```
jgmontoya: We shouldn't have SQL queries outside of the `whitenoise/database/` dir. Let's move this to an appropriate method on the `GrupPushToken` struct to follow our conventions

src/whitenoise/event_processor/event_handlers/handle_mls_message.rs
```diff
@@ -64,70 +64,100 @@ impl Whitenoise {
...
-                    Whitenoise::spawn_new_message_notification_if_enabled(
-                        account, &group_id, &msg, group_name,
-                    );
-                    self.emit_message_update(&group_id, UpdateTrigger::NewMessage, msg);
-                    self.emit_chat_list_update(
+        if let Some((group_id, inner_event, message)) = Self::extract_message_details(&outcome) {
+            let is_push_group_message = message.kind
+                == Kind::from(mdk_core::mip05::TOKEN_REQUEST_KIND)
+                || message.kind == Kind::from(mdk_core::mip05::TOKEN_LIST_RESPONSE_KIND)
+                || message.kind == Kind::from(mdk_core::mip05::TOKEN_REMOVAL_KIND);
```
jgmontoya: This could perhaps use `Self::is_push_group_message_kind(message.kind)` (implemented in `src/whitenoise/push_notifications.rs`)

src/whitenoise/push_notifications.rs
```diff
@@ -158,12 +735,144 @@ fn validate_raw_token(raw_token: &str) -> Result<()> {
...
+    use crate::whitenoise::{
+        relays::Relay,
+        test_utils::{
+            count_published_events_for_account, create_mock_whitenoise,
+            create_nostr_group_config_data, setup_multiple_test_accounts,
+            wait_for_key_package_publication, wait_for_published_event_count,
+        },
+    };
+
+    async fn setup_two_member_group(
```
jgmontoya: I think we have something similar already implemented for other tests (see `src/whitenoise/event_processor/account_event_processor.rs` and `src/whitenoise/event_processor/event_handlers/handle_mls_message.rs`)

Perhaps we can consolidate all of them in a single implementation and move it to `test_utils`?

src/whitenoise/event_processor/event_handlers/handle_mls_message.rs
mubarakcoded: Agreed.

src/whitenoise/event_processor/event_handlers/handle_giftwrap.rs
erskingardner: This is already on the background welcome-finalization path in the current code. `handle_giftwrap()` spawns `background_finalize_welcome()`, and the `share_local_push_token_to_group()` call happens inside `finalize_welcome_with_instance()`, so it’s no longer running synchronously in the event processor loop.

src/whitenoise/push_notifications.rs
```diff
@@ -128,24 +138,564 @@ impl Whitenoise {
...
+
+    #[perf_instrument("push_notifications")]
+    async fn merge_token_request(
+        &self,
+        account: &Account,
+        mls_group_id: &GroupId,
+        leaf_index: u32,
+        request: mdk_core::mip05::TokenRequest,
+    ) -> Result<()> {
+        let Some(token) = request.tokens.into_iter().next() else {
```
mubarakcoded, line 762: This only extracts the first token from the request, and then the function only upserts that single   
 token to the database. If the request contains multiple tokens, all the others are silently discarded.

erskingardner: Correct - my agent also flagged that - but that is the correct shape. The only place this is called is when you receive a user's 447 event. These should only ever contain a single token (given you're sending your token for the device that you're currently on).

mubarakcoded: Agreed.

src/whitenoise/push_notifications.rs
```diff
@@ -128,24 +303,583 @@ impl Whitenoise {
...
+                })?;
+
+                GroupPushToken::delete_by_member_pubkey(
+                    &account.pubkey,
+                    &message.mls_group_id,
+                    &message.event.pubkey,
+                    &self.database,
+                )
+                .await?;
+            }
```
jgmontoya, line 461: This validates the `sender_leaf_index` (`.ok_or_else`) but it then doesn't really use the `sender_leaf_index` at all

src/whitenoise/push_notifications.rs
```diff
@@ -128,24 +303,583 @@ impl Whitenoise {
...
+                    )
+                })?;
+
+                GroupPushToken::delete_by_member_pubkey(
+                    &account.pubkey,
+                    &message.mls_group_id,
+                    &message.event.pubkey,
+                    &self.database,
+                )
+                .await?;
```
jgmontoya: This will delete all the tokens for a given pubkey, just double-checking whether that's correct/the expected behavior

src/whitenoise/event_processor/event_handlers/handle_mls_message.rs
```diff
@@ -63,213 +65,314 @@ impl Whitenoise {
...
-                    &account.pubkey,
-                    group_id,
-                    &relay_urls,
-                )
-                .await?;
+    async fn handle_push_application_message(
+        &self,
+        account: &Account,
+        outcome: &MessageProcessingOutcome,
+        _group_id: &GroupId,
```
jgmontoya: why receive the group_id if it's not used?

src/whitenoise/push_notifications.rs
erskingardner: I double checked the behavior and we have two ways of deleting - either by leaf_index (when we only want to remove a single token) and by pubkey (when a user is removed from the group).

src/whitenoise/push_notifications.rs
```diff
@@ -128,24 +312,586 @@ impl Whitenoise {
...
+            }
+            None => {
+                let mdk = self.create_mdk_for_account(account.pubkey)?;
+                let leaf_index = mdk.own_leaf_index(group_id)?;
+                GroupPushToken::delete(&account.pubkey, group_id, leaf_index, &self.database)
+                    .await?;
+            }
+        }
+
```
jgmontoya, line 871: Here we could take mdk and the leaf_index out of the match / branches

```rs
let mdk = self.create_mdk_for_account(account.pubkey)?;
let leaf_index = mdk.own_leaf_index(group_id)?;
match token_tag {
  Some(token_tag) => {
    GroupPushToken::upsert(...
  }
  None => {
    GroupPushToken::delete(...
  }
}
```

But perhaps even better would be to get the mdk instance as a param, because I see that for example `share_push_token_to_joined_groups` already creates an mdk instance, and this (`sync_local_group_push_token_cache`) runs for every group, so we could avoid doing the mdk initialization for every account

erskingardner: I hoisted the mdk and leaf_index out for now - and created https://github.com/marmot-protocol/whitenoise-rs/issues/698 to track the other potential improvement.

PR #695, opened 2026-03-30
Add fetch_messages_unread_with_minimum

dannym-arx:
marmot

Adds `fetch_messages_unread_with_minimum`: a new public API method that returns all unread messages for a chat, but guarantees a minimum number of messages even when the unread count is low.

## What it does

The effective fetch size is `max(unread_count, minimum)`. Two cases:

- **More unreads than the minimum:** returns every unread message since the read marker. You get the full unread backlog.
- **Fewer unreads than the minimum:** the floor kicks in. You get at least a full page of the newest messages regardless of read state.

The default minimum is 50. Values above 200 are clamped.

## How it works

A single SQL CTE computes the unread count and uses `MAX(cnt, minimum)` as the LIMIT in the same query. One round-trip, no separate COUNT call. The unread boundary uses the same `created_at > marker_created_at` logic as the existing `count_unread_for_group`.

Messages come back oldest-first, same as `fetch_aggregated_messages_for_group`.

## Testing

Four unit tests covering:

- Unread count exceeds minimum: all unreads returned
- Unread count below minimum: floor applied, newest N messages returned
- No read marker: all messages treated as unread
- Unknown account: rejected with `AccountNotFound`

## Summary by CodeRabbit

* **New Features**
  * Fetch unread group messages with a configurable minimum threshold (defaults to 50) so callers receive at least that many messages when available.
  * Deleted messages are returned as tombstones in results.

* **Behavior**
  * Results are selected newest-first but delivered oldest-first.
  * Missing read markers are treated as "all unread".
  * Unknown accounts return an explicit error.

* **Tests**
  * Added integration tests for unread/minimum behavior, tombstone inclusion, and error cases.

### DIFF
--- a/src/whitenoise/database/aggregated_messages.rs
+++ b/src/whitenoise/database/aggregated_messages.rs
@@ -3193,8 +3193,14 @@ mod tests {
 
         // Insert 3 messages: seeds 1, 2, 3
         for i in 1u8..=3 {
-            insert_message_at(i, author, base_ts + u64::from(i), &group_id, &whitenoise.database)
-                .await;
+            insert_message_at(
+                i,
+                author,
+                base_ts + u64::from(i),
+                &group_id,
+                &whitenoise.database,
+            )
+            .await;
         }
 
         // Delete seed-2 message
--- a/src/whitenoise/messages.rs
+++ b/src/whitenoise/messages.rs
@@ -707,7 +707,8 @@ impl Whitenoise {
             read_marker.as_ref(),
             minimum_val,
             &self.database,
-        ).await?)
+        )
+        .await?)
     }
 
     /// Fetch a single aggregated message by its event ID.


src/whitenoise/messages.rs
```diff
@@ -674,6 +674,51 @@ impl Whitenoise {
...
+        &self,
+        pubkey: &PublicKey,
+        group_id: &GroupId,
+        minimum: Option<u32>,
+    ) -> Result<Vec<ChatMessage>> {
+        Account::find_by_pubkey(pubkey, &self.database).await?;
+
+        let read_marker = self
+            .get_last_read_message_id(
+                &Account::find_by_pubkey(pubkey, &self.database).await?,
```
erskingardner: You're looking up the account twice here.

src/whitenoise/database/aggregated_messages.rs
```diff
@@ -347,6 +347,113 @@ impl AggregatedMessage {
...
+        //
+        // The unread count uses the same timestamp-subquery logic as
+        // `count_unread_for_group`: messages whose `created_at` is strictly after
+        // the read marker's `created_at` are considered unread.  When there is no
+        // read marker every non-deleted kind-9 message is unread.
+        let rows: Vec<AggregatedMessageRow> = match read_marker {
+            Some(marker_id) => {
+                sqlx::query_as(
+                    "WITH unread_count AS (
+                       SELECT COUNT(*) AS cnt
```
erskingardner: The CTE correctly excludes deleted messages (`deletion_event_id IS NULL`) when counting unreads, but the outer `SELECT` that fetches the actual rows does **not** include this filter. This means deleted messages can be returned in the result set, inflating the count and surfacing tombstoned content to callers.

Compare to `find_messages_by_group_paginated`, which omits the deletion filter too — but `count_unread_for_group` (the existing authoritative unread counter) does apply it. The CTE count and the actual returned rows need to be in sync on this filter.

src/whitenoise/database/aggregated_messages.rs
```diff
@@ -347,6 +347,113 @@ impl AggregatedMessage {
...
+                .bind(minimum_val) // MAX(cnt, minimum)
+                .fetch_all(&database.pool)
+                .await?
+            }
+            None => {
+                // No read marker: every non-deleted kind-9 message is unread.
+                // The effective limit is MAX(total_count, minimum).
+                sqlx::query_as(
+                    "WITH unread_count AS (
+                       SELECT COUNT(*) AS cnt
```
erskingardner: Same issue as above exists here too.

src/whitenoise/database/aggregated_messages.rs
```diff
@@ -347,6 +347,113 @@ impl AggregatedMessage {
...
+        // Single-query approach: a CTE computes the unread count, then the outer
+        // SELECT uses MAX(unread_count, minimum) as the LIMIT.  This avoids a
+        // separate COUNT round-trip and keeps the logic in one atomic read.
+        //
+        // The unread count uses the same timestamp-subquery logic as
+        // `count_unread_for_group`: messages whose `created_at` is strictly after
+        // the read marker's `created_at` are considered unread.  When there is no
+        // read marker every non-deleted kind-9 message is unread.
+        let rows: Vec<AggregatedMessageRow> = match read_marker {
+            Some(marker_id) => {
```
erskingardner: The `Some(marker_id)` and `None` branches are nearly identical SQL, differing only in whether the CTE includes the `created_at >` condition. SQLite's `COALESCE` trick (already used in the marker subquery's fallback) could potentially unify them into one query, reducing surface area. That said, the current split is readable and the duplication is manageable — this is a style/maintenance preference rather than a correctness issue.

src/whitenoise/messages.rs
```diff
@@ -674,6 +674,46 @@ impl Whitenoise {
...
+        let minimum_val = minimum.unwrap_or(50);
+
+        AggregatedMessage::find_messages_with_unread_minimum(
+            group_id,
+            read_marker.as_ref(),
+            minimum_val,
+            &self.database,
+        )
+        .await
+        .map_err(|e| {
```
mubarakcoded: This already returns a WhitenoiseError, as every other AggregatedMessage::find_* call in this file does .await? Should be the same here to avoid losing the original variant

src/whitenoise/messages.rs
```diff
@@ -2739,6 +2779,203 @@ mod tests {
...
+        let result = whitenoise
+            .fetch_messages_unread_with_minimum(&stranger, &group.mls_group_id, None)
+            .await;
+
+        assert!(matches!(
+            result.unwrap_err(),
+            WhitenoiseError::AccountNotFound
+        ));
+    }
+
```
mubarakcoded, line 2975: Can we add a quick test that actually hits the DB with None and verifies we get the default 50 behaviour?

PR #699, opened 2026-03-31
Feat/leave group

jgmontoya:
## Summary by CodeRabbit

* **New Features**
  * Added a CLI command to voluntarily relinquish admin status in MLS groups.
  * Group departures now record whether a user left voluntarily vs was removed; chat lists and notifications reflect this state.

* **Bug Fixes**
  * Prevents attempts to leave a group already departed from and improves related error handling and post-leave updates.

### DIFF
--- a/src/whitenoise/notification_streaming/mod.rs
+++ b/src/whitenoise/notification_streaming/mod.rs
@@ -232,7 +232,9 @@ impl Whitenoise {
             .await
         {
             Ok(Some(ag)) => {
-                if !ag.is_accepted() {
+                if ag.is_removed() {
+                    Some("departed")
+                } else if !ag.is_accepted() {
                     Some("unaccepted")
                 } else if ag.is_muted() {
                     Some("muted")
@@ -581,7 +583,7 @@ mod tests {
         whitenoise: &Whitenoise,
         account_pubkey: &PublicKey,
         group_id: &GroupId,
-    ) {
+    ) -> AccountGroup {
         let ag = AccountGroup {
             id: None,
             account_pubkey: *account_pubkey,
@@ -598,7 +600,7 @@ mod tests {
             created_at: Utc::now(),
             updated_at: Utc::now(),
         };
-        ag.save(&whitenoise.database).await.unwrap();
+        ag.save(&whitenoise.database).await.unwrap()
     }
 
     #[tokio::test]
@@ -954,6 +956,24 @@ mod tests {
                 .should_suppress_notification(&account.pubkey, &group_id)
                 .await,
             Some("unknown")
+        );
+    }
+
+    #[tokio::test]
+    async fn test_should_suppress_notification_suppresses_departed_group() {
+        let (whitenoise, _data_temp, _logs_temp) = create_mock_whitenoise().await;
+        let account = whitenoise.create_identity().await.unwrap();
+        let group_id = GroupId::from_slice(&[35u8; 32]);
+
+        // Create an accepted group, then mark it as left (voluntary departure)
+        let ag = create_accepted_account_group(&whitenoise, &account.pubkey, &group_id).await;
+        ag.mark_left(&whitenoise).await.unwrap();
+
+        assert_eq!(
+            whitenoise
+                .should_suppress_notification(&account.pubkey, &group_id)
+                .await,
+            Some("departed")
         );
     }
 


src/whitenoise/groups.rs
```diff
@@ -767,32 +767,81 @@ impl Whitenoise {
...
     #[perf_instrument("groups")]
     pub async fn leave_group(&self, account: &Account, group_id: &GroupId) -> Result<()> {
+        // Guard: can't leave a group you've already departed from
+        let account_group = AccountGroup::get(self, &account.pubkey, group_id)
+            .await?
+            .ok_or(WhitenoiseError::GroupNotFound)?;
+
+        if account_group.is_removed() {
+            return Err(WhitenoiseError::AlreadyDepartedFromGroup);
+        }
```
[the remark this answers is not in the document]
jgmontoya: The scenario where an MLS group exists without an `AccountGroup` row only occurs in test fixtures `create_mdk_group_without_auto_accept` deliberately deletes the row for test setup. In production, the row is always created during create_group() (line 192) or welcome processing (line 170).

If the row is genuinely missing, something is corrupted and `GroupNotFound` is the correct signal. Making the lookup optional would add a dead code path for a test-only scenario and weaken the guard against corrupted state.

src/whitenoise/database/accounts_groups.rs
dannym-arx, line 316: I think this is more of a meta comment, because we're doing this elsewhere in the code, but what do you think about moving long queries (such as this one) in separate files, so that they're easier to read, and we get syntax highlighting when reviewing?

src/whitenoise/notification_streaming/mod.rs
mubarakcoded, line 241: I think we need ag.is_removed() as a check here (returning something like Some("departed")), and probably a test for it too, it's the only suppression scenario that doesn't have one yet.

src/whitenoise/groups.rs
```diff
@@ -767,32 +767,81 @@ impl Whitenoise {
...
     /// * `account` - The account that wants to leave the group
     /// * `group_id` - The ID of the group to leave
+    ///
+    /// # Errors
+    /// * `AlreadyDepartedFromGroup` if the account has already departed
+    /// * MDK errors if the caller is still an admin (must call `self_demote()` first)
+    /// * Relay errors if the proposal cannot be published
     #[perf_instrument("groups")]
     pub async fn leave_group(&self, account: &Account, group_id: &GroupId) -> Result<()> {
+        // Guard: can't leave a group you've already departed from
```
erskingardner, line 810: A cheap check here is probably sensible. 

> If an admin calls leave_group without first calling self_demote, the MDK will presumably reject the SelfRemove proposal. The error message from MDK may be opaque. Consider adding a guard:
> 
> ```rs
> if self.is_account_group_admin(account, group_id).await? {
>     return Err(WhitenoiseError::MustSelfDemoteFirst); // or similar
> }
> ```
>
> This would give users a clear error rather than a cryptic MDK failure.

db_migrations/0042_add_self_removed_to_accounts_groups.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE accounts_groups ADD COLUMN self_removed INTEGER NOT NULL DEFAULT 0 CHECK (self_removed IN (0, 1));
```
erskingardner, line 1: Definitely should fix. 

> The bare ALTER TABLE ... ADD COLUMN will fail if run twice. If your migration runner guarantees exactly-once execution this is fine, but if not, consider:
> 
> ```sql
> ALTER TABLE accounts_groups ADD COLUMN IF NOT EXISTS self_removed INTEGER NOT NULL DEFAULT 0 CHECK (self_removed IN (0, 1));
> ```
>
> (SQLite added IF NOT EXISTS for ADD COLUMN in 3.35.0.)

src/whitenoise/database/accounts_groups.rs
jgmontoya: Separation of concerns vs locality of behavior...

This query in particular is not really all that long/hard to read, but it's "polluted" by all those comments... (I'd like to change it actually, I'd make .save be able to modify basically any field, but that's for another time).

For actually long or particularly hard to understand queries, I'd be ok with moving them to their own file as long as we can then reference them with understandable names

db_migrations/0042_add_self_removed_to_accounts_groups.sql
jgmontoya: This is not needed (we've done this a bunch of times before) and I think SQLite does not support `IF NOT EXISTS` on `ALTER TABLE ADD COLUMN`

src/whitenoise/groups.rs
jgmontoya: MDK already handles this clearly (At `mdk-core/src/groups.rs`):

```rs
if self.is_leaf_node_admin(group_id, own_leaf)? {
    return Err(Error::Group(
        "Admins must self-demote before leaving. \
         Use self_demote() first."
            .to_string(),
    ));
}
```

I could add the additional check at the whitenoise-rs level but it feels redundant. wdyt?

erskingardner: yeah - fair - it's ok without.

PR #701, opened 2026-04-01
[codex] publish MIP-05 notifications after chat delivery

erskingardner:
## Summary

Publish MIP-05 notification requests only after a real kind `9` chat message is accepted by at least one relay, and keep notification delivery strictly best-effort so chat success never turns into chat failure.

## What Changed

- added a structured `MessagePublishResult` so the caller can distinguish successful chat delivery from failed publication without relying on a bare boolean
- gated notification-request publication in the background send path so only successful kind `9` chat sends trigger follow-up `446 -> 1059` publication
- added post-delivery notification batching that:
  - loads cached `group_push_tokens`
  - skips unusable rows with warnings instead of failing the send
  - dedupes identical encrypted tokens per notification server
  - uses MDK's `build_notification_batches` for batching and chunking
- resolved notification-server inbox relays in this order:
  - cached inbox relays
  - synced kind `10050` inbox relays using batch relay hints as query seeds
  - hint fallback for the current attempt only when no authoritative inbox relays are available
- exposed production relay-sync context/helpers so notification-server relay discovery can reuse the existing user relay-list machinery
- updated stale push-notification doc comments while touching the module

## Behavioral Notes

- notification publication is attempted only after at least one relay accepts the chat message
- notification publication failures only log warnings and never change the chat delivery status from sent to failed
- sender-owned member pubkeys are excluded from recipient selection so the sender's other devices do not receive notifications for their own message
- relay hints are not persisted as authoritative inbox relays unless a real kind `10050` relay list is fetched

## Validation

- `just test-unit`
  - result: `1668 passed, 3 skipped`

## Summary by CodeRabbit

* **New Features**
  * Best-effort push-notification publishing after group chat delivery; notifications are attempted in background for chat messages.

* **Improvements**
  * Chat delivery decoupled from notification publishing; notification failures no longer mark chats as failed.
  * Improved relay resolution, token validation and deduplication, and relay-sync orchestration via a dedicated sync context.
  * Follow-list updates now include an explicit creation timestamp when published.

* **Bug Fixes**
  * More explicit publish-result reporting and refined delivery-status handling to avoid false failures.

* **Tests**
  * Expanded coverage: relay dedupe/order, recipient selection, multi-event publishing, APNS validation, and publish-result behavior.

### DIFF
--- a/src/whitenoise/push_notifications.rs
+++ b/src/whitenoise/push_notifications.rs
@@ -1202,26 +1202,6 @@ impl Whitenoise {
         publish_push_group_message_with(&self.config, &self.relay_control, account, group_id, rumor)
             .await
     }
-
-    #[perf_instrument("push_notifications")]
-    pub(crate) async fn publish_notification_requests_after_delivery(
-        &self,
-        account: &Account,
-        group_id: &GroupId,
-    ) -> Result<()> {
-        let user_relay_sync = self.user_relay_sync_context();
-        let ephemeral = self.relay_control.ephemeral();
-
-        publish_notification_requests_after_delivery_with(
-            &self.config,
-            &self.database,
-            &user_relay_sync,
-            &ephemeral,
-            account.pubkey,
-            group_id,
-        )
-        .await
-    }
 }
 
 fn validate_raw_token(raw_token: &str) -> Result<()> {
@@ -1333,6 +1313,25 @@ mod tests {
             created_at: Utc::now(),
             updated_at: Utc::now(),
         }
+    }
+
+    async fn publish_notification_requests_after_delivery_for_test(
+        whitenoise: &Whitenoise,
+        account: &Account,
+        group_id: &GroupId,
+    ) -> Result<()> {
+        let user_relay_sync = whitenoise.user_relay_sync_context();
+        let ephemeral = whitenoise.relay_control.ephemeral();
+
+        publish_notification_requests_after_delivery_with(
+            &whitenoise.config,
+            &whitenoise.database,
+            &user_relay_sync,
+            &ephemeral,
+            account.pubkey,
+            group_id,
+        )
+        .await
     }
 
     async fn wait_for_user_inbox_relays_on_network(
@@ -1772,10 +1771,13 @@ mod tests {
             &member_account,
         )
         .await;
-        whitenoise
-            .publish_notification_requests_after_delivery(&admin_account, &group_id)
-            .await
-            .unwrap();
+        publish_notification_requests_after_delivery_for_test(
+            &whitenoise,
+            &admin_account,
+            &group_id,
+        )
+        .await
+        .unwrap();
 
         let cached_tokens = GroupPushToken::find_by_account_and_group(
             &admin_account.pubkey,


src/whitenoise/messages.rs
```diff
@@ -122,8 +125,28 @@ impl Whitenoise {
                 )
                 .await;
 
+            if kind == 9
```
dannym-arx: shouldn't we also do it for reactions (7)? many apps including signal notify users for reactions. I think we should do the same (tho with a setting for the user to not trigger it upon receipt)

src/whitenoise/push_notifications.rs
```diff
@@ -291,6 +299,310 @@ async fn publish_push_group_message_with(
...
+
+fn dedupe_relay_urls(relay_urls: &[RelayUrl]) -> Vec<RelayUrl> {
+    let mut deduped = Vec::with_capacity(relay_urls.len());
+    let mut seen = HashSet::with_capacity(relay_urls.len());
+
+    for relay_url in relay_urls {
+        if seen.insert(relay_url.clone()) {
+            deduped.push(relay_url.clone());
+        }
+    }
```
dannym-arx, line 327: do we need the double clone? just clone into seen and push a reference-derived clone only on succes

src/whitenoise/push_notifications.rs
```diff
@@ -291,6 +299,304 @@ async fn publish_push_group_message_with(
...
+        if !seen_tokens.insert((token.server_pubkey, encrypted_token.clone())) {
+            skipped_duplicate_token += 1;
+            continue;
+        }
+
+        recipient_tokens.push(TokenTag {
+            encrypted_token,
+            server_pubkey: token.server_pubkey,
+            relay_hint,
+        });
```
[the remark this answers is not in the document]
dannym-arx: I agree with coderabbit, but I'm not sure if this actually matters

src/whitenoise/push_notifications.rs
```diff
@@ -952,15 +1284,759 @@ impl PushRegistration {
...
+            &group_id,
+            &whitenoise.database,
+        )
+        .await
+        .unwrap();
+        assert!(cached_tokens.is_empty());
+    }
+
+    #[tokio::test]
+    async fn test_publish_notification_requests_after_chat_delivery_publishes_1059_with_expected_446_tokens()
```
dannym-arx: that's a long name lol

perhaps shorten it and leave a comment?

src/relay_control/ephemeral.rs
```diff
@@ -514,27 +557,14 @@ impl EphemeralPlane {
...
-            }
             Ok(output) => {
-                tracing::warn!(
-                    target: "whitenoise::messages::delivery",
-                    "Publish completed but no relay accepted the message (failed: {:?})",
-                    output.failed.keys().collect::<Vec<_>>(),
+                debug_assert!(
+                    !output.success.is_empty(),
+                    "publish_event_to returned Ok with no accepted relays"
                 );
```
dannym-arx: I think this should be a standard assert

src/whitenoise/push_notifications.rs
```diff
@@ -291,6 +299,310 @@ async fn publish_push_group_message_with(
...
+            Err(error) => {
+                tracing::warn!(
+                    target: "whitenoise::push_notifications",
+                    server_pubkey = %server_pubkey.to_hex(),
+                    relay_source = NotificationRelaySource::Synced.as_str(),
+                    relay_hint_count = deduped_hints.len(),
+                    error = %error,
+                    "Failed to sync notification server inbox relays from relay hints"
+                );
+                server_user.relays(RelayType::Inbox, database).await?
```
dannym-arx: this doesn't make sense to me, didn't we check those already on 432?

src/whitenoise/messages.rs
erskingardner: yeah - I think you're right - it's probably better to send on all but decide what you see on the receiving end.

src/whitenoise/messages.rs
```diff
@@ -122,8 +125,28 @@ impl Whitenoise {
...
+                && let Some(_publish_output) = publish_result.output()
```
mubarakcoded: Small nit - MessagePublishResult already has a succeeded() method for this. Since _publish_output is never used, this could be if kind == 9 && publish_result.succeeded()

src/whitenoise/database/group_push_tokens.rs
```diff
@@ -124,8 +124,6 @@ impl From<GroupPushTokenRow> for GroupPushToken {
     }
 }
 
-// PR 1 only persists the cache locally; gossip/publication flows consume these
-// helpers in follow-up PRs, while tests exercise them now.
 #[allow(dead_code)]
```
erskingardner: Do we need this dead_code line now?

src/whitenoise/push_notifications.rs
```diff
@@ -291,6 +299,304 @@ async fn publish_push_group_message_with(
...
+        }
+
+        if active_member_pubkey == sender_member_pubkey {
+            continue;
+        }
+
+        let Some(relay_hint) = token.relay_hint.clone() else {
+            skipped_missing_hint += 1;
+            continue;
+        };
```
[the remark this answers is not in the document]
erskingardner: We're going to handle this in a separate issue: #703

src/whitenoise/push_notifications.rs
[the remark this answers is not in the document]
erskingardner: Yeah I don't think this is worth fixing now.

src/whitenoise/users/relay_sync.rs
```diff
@@ -339,32 +375,64 @@ impl User {
                 e
             })?;
 
-        self.apply_relay_event(whitenoise, relay_type, relay_event)
+        self.apply_relay_event_with_context(context, relay_type, relay_event)
             .await
     }
 
+    #[perf_instrument("relay_sync")]
+    async fn sync_relays_for_type_with_scope(
```
jgmontoya, line 383: is this method used?

src/whitenoise/push_notifications.rs
```diff
@@ -891,6 +1202,26 @@ impl Whitenoise {
         publish_push_group_message_with(&self.config, &self.relay_control, account, group_id, rumor)
             .await
     }
+
+    #[perf_instrument("push_notifications")]
+    pub(crate) async fn publish_notification_requests_after_delivery(
```
jgmontoya: Is this method used?

src/whitenoise/users/relay_sync.rs
erskingardner: Yes. 

The active call chain is:
relay_sync.rs (line 21) update_relay_lists(...)
-> relay_sync.rs (line 29) update_relay_lists_with_scope(...)
-> relay_sync.rs (line 87) update_nip65_relays_with_scope(...)
-> relay_sync.rs (line 383) sync_relays_for_type_with_scope(...)

PR #722, opened 2026-04-08
fix: bound MIP-05 token-list response task concurrency with semaphore

mubarakcoded:
Fixes #706.

`schedule_pending_token_response` previously spawned an unbounded `tokio::spawn` for every incoming MIP-05 token request, allowing memory and DB contention to grow without limit under a request flood.

## Changes

- Added `token_response_semaphore: Arc<Semaphore>` (capacity 10) to `Whitenoise`
- In `schedule_pending_token_response`, `try_acquire_owned()` is called after the dedup-map insert; tasks that cannot acquire a permit are dropped immediately and their dedup key is removed
- The `OwnedSemaphorePermit` is held for the lifetime of the spawned task and released on completion

## Summary by CodeRabbit

* **Chores**
  * Introduced a concurrency cap (10) for delayed token-list responses and ensure each scheduled response respects the limit for its entire lifetime.
  * Added atomic deduplication so repeated scheduling requests for the same token response are ignored.

* **Bug Fixes**
  * If concurrency is exhausted, scheduling is canceled (with a logged warning) and no stray pending entry is left behind.

* **Tests**
  * Added tests and test helpers validating deduplication and behavior when concurrency is exhausted.

### DIFF
--- a/src/whitenoise/event_processor/event_handlers/handle_mls_message.rs
+++ b/src/whitenoise/event_processor/event_handlers/handle_mls_message.rs
@@ -839,6 +839,8 @@ impl Whitenoise {
 
 #[cfg(test)]
 mod tests {
+    use std::time::Duration;
+
     use mdk_core::mip05::{
         ENCRYPTED_TOKEN_LEN, LeafTokenTag, TokenTag, build_token_list_response_rumor,
         build_token_removal_rumor, build_token_request_rumor,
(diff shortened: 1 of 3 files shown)


dannym-arx (dismissed):

src/whitenoise/event_processor/event_handlers/handle_mls_message.rs
```diff
@@ -2025,4 +2027,136 @@ mod tests {
             message_event.err()
         );
     }
+
+    #[tokio::test]
+    async fn test_duplicate_token_request_does_not_create_second_task() {
```
jgmontoya, line 2082: From my agent:

> ### Correctness: dedup test has a latent race with `delay_ms = 0`
> 
> **`src/whitenoise/event_processor/event_handlers/handle_mls_message.rs`:2042–2052**
> 
> ```rust
> whitenoise.schedule_token_response_for_test(
>     member_account.clone(),
>     fake_group_id.clone(),
>     fake_event_id,
> );
> 
> assert!(whitenoise.has_pending_token_response(
>     &member_account.pubkey,
>     &fake_group_id,
>     &fake_event_id,
> ));
> ```
> 
> `schedule_token_response_for_test` spawns a task with `delay_ms = 0`. The spawned task calls
> `dispatch_pending_token_response`, which removes the key from the map. Because Tokio's executor
> can run the spawned task on any available thread, and the fake group / account have no push
> tokens (so `respond_to_token_request` returns quickly), the key can be gone by the time the
> first `assert!` is reached. The test then passes vacuously: the second `has_pending_token_response`
> asserts `true` because the first removes the key (returning `false`), and the duplicate path is
> never exercised at all.
> 
> The test could be made deterministic by exhausting the semaphore before the first schedule call —
> matching the pattern already used in `test_token_request_dropped_when_semaphore_exhausted`. The
> spawned task would still be dropped immediately (no key to clear), but the dedup check would still
> be exercised on the second call via the semaphore-exhaustion path, which is close but not
> identical. Alternatively, using `tokio::task::yield_now` control or holding the semaphore to
> prevent the spawned task from running would work. Either way, the current test is not reliably
> testing what its name claims.

src/whitenoise/event_processor/event_handlers/handle_mls_message.rs
```diff
@@ -2025,4 +2027,136 @@ mod tests {
...
+        assert!(
+            whitenoise.has_pending_token_response(
+                &member_account.pubkey,
+                &fake_group_id,
+                &fake_event_id,
+            ),
+            "pending entry should still be present after duplicate request"
+        );
+
+        let _ = admin_account;
```
jgmontoya: what's this for?

jgmontoya (commented):
Question: Should the semaphore be a per-group thing instead of a global one?

mubarakcoded:
> Question: Should the semaphore be a per-group thing instead of a global one?

I kept it global intentionally because the concern in #706 is: total resource pressure (memory + DB connections), not per-group fairness. A per-group semaphore would still allow O(N) concurrent tasks across N groups. I can layer per-group fairness on top later if needed.

erskingardner (approved):

jgmontoya (approved):

PR #727, opened 2026-04-09
Re-export mdk-core types via whitenoise::mdk module

### LINKED ISSUE #561: Re-export some MDK core types
Currently the whitenoise flutter app needs to import both mdk-core and whitenoise-rs crates, this is awkward and can lead to version drift. We should re-export everything we need directly from whitenoise-rs

dannym-arx:
marmot

Adds a `whitenoise::mdk` module that re-exports the five `mdk-core` types the Flutter bridge layer uses: `GroupId`, `Group`, `GroupState`, `NostrGroupConfigData`, and `NostrGroupDataUpdate`.

The Flutter app currently depends on both `whitenoise` and `mdk-core` directly. When the two crates pin different revisions, the compiler sees them as distinct types, and builds break. This re-export module lets the Flutter app drop its direct `mdk-core` dependency and import everything through `whitenoise::mdk`, keeping versions locked to a single source of truth.

## What changed

- **`src/mdk.rs`** (new): Re-export module with the five types.
- **`src/lib.rs`**: Added `pub mod mdk;`.

## What the Flutter app does next (separate PR)

1. Remove `mdk-core` from `rust/Cargo.toml`.
2. Replace `use mdk_core::prelude::GroupId` with `use whitenoise::mdk::GroupId` (and the same for the other four types).

Closes #561

## Summary by CodeRabbit

* **New Features**
  * Public API now includes MLS protocol types for group management, configuration, and state tracking, enabling unified access to protocol functionality without requiring external dependencies.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -17,6 +17,7 @@ pub(crate) mod relay_control;
 // Crate-private because the macro expands to `crate::perf_span!(...)` which
 // only resolves inside this crate.
 pub(crate) use whitenoise_macros::perf_instrument;
+pub mod mdk;
 mod types;
 pub mod whitenoise;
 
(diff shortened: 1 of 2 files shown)


erskingardner (approved):
Oh thank god. 😂

PR #729, opened 2026-04-09
Reuse single MDK connection per push notification operation

mubarakcoded:
Fixes #707.

## Summary
- Refactored `publish_push_group_message_with` and `sync_local_group_push_token_cache` to accept `&MDK<MdkSqliteStorage>` instead of calling `Account::create_mdk` internally
- Each top-level operation (`share_push_token_to_joined_groups`, `remove_push_token_from_joined_groups`, etc.) now creates a single MDK instance and threads it through all helpers
- Removed the now-redundant `publish_push_group_message` and `group_push_token_tags_for_response` wrapper methods

## Test plan
- [x] `just precommit-quick` passes
- [x] Integration tests pass with Docker (`just precommit`)
- [ ] Verify push token share/remove flows work on device

## Summary by CodeRabbit

* **Refactor**
  * Push-notification flows now reuse a shared account context to reduce redundant setup and improve reliability.
  * Token publishing and local cache synchronization for group push tokens are streamlined for more consistent, efficient delivery.
  * Added performance instrumentation across push-notification paths to improve observability and tuning.

### DIFF
--- a/src/whitenoise/push_notifications.rs
+++ b/src/whitenoise/push_notifications.rs
@@ -185,25 +185,15 @@ impl PendingTokenResponseContext {
             &self.config.data_dir,
             &self.config.keyring_service_id,
         )?;
-        let token_tags = group_push_token_tags_for_response_with(
-            &account.pubkey,
-            group_id,
-            &self.database,
+        respond_to_token_request_with(
             &mdk,
-        )
-        .await?;
-
-        if token_tags.is_empty() {
-            return Ok(());
-        }
-
-        let rumor = build_token_list_response_rumor(
-            account.pubkey,
-            nostr_sdk::Timestamp::now(),
+            &self.database,
+            &self.relay_control,
+            account,
+            group_id,
             request_event_id,
-            token_tags,
-        )?;
-        publish_push_group_message_with(&mdk, &self.relay_control, account, group_id, rumor).await
+        )
+        .await
     }
 }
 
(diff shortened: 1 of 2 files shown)


dannym-arx (dismissed):

erskingardner (dismissed):

src/whitenoise/push_notifications.rs
```diff
@@ -1211,9 +1226,14 @@ impl Whitenoise {
...
-            .group_push_token_tags_for_response(account, group_id)
-            .await?;
+        let mdk = self.create_mdk_for_account(account.pubkey)?;
+        let token_tags = group_push_token_tags_for_response_with(
+            &account.pubkey,
+            group_id,
+            &self.database,
+            &mdk,
+        )
+        .await?;
```
jgmontoya: The body of this method and of `PendingTokenResponseContext::respond_to_token_request` are nearly identical. Could we extract it to a single method that receives `&mdk`?

src/whitenoise/push_notifications.rs
jgmontoya, line 1: BTW we're also getting a new instance for method `merge_token_list_response`, which is called from `handle_received_push_group_message` which in turn is called from `handle_push_application_message` which at its call point already could have an `mdk` instance.

Not sure if worth it to "drill down" the `mdk` instance all that many layers deep but just throwing it out there for consideration

erskingardner (approved):

PR #734, opened 2026-04-10
Delete ~150 tests that only tested the compiler

dannym-arx:
marmot

The marmots audited the burrow and found tests that were keeping the compiler company with nothing to say. They have been evicted.

## What was removed

Around 150 tests across 26 files, net ~2.7K lines. Each one matched a pattern with no defensive value:

- Derive tests: `format!("{:?}", x)` + substring checks, `let c = x.clone(); assert_eq!(x, c)`, and default-value assertions when `Default` is derived.
- Pure serde roundtrips: `to_string` then `from_str` then `assert_eq` for types with only derived `Serialize`/`Deserialize` and no custom wire format.
- Construct-and-echo tests that built a struct from literals and asserted those same literals came back out.
- Trivial wrapper tests covering one-line functions like `fn is_image(&self) -> bool { self.mime_type.starts_with("image/") }`.
- Getter echoes: `let x = T::with_config(c); assert_eq!(x.config(), c)`.
- Error-wrapping tests that exercised thiserror `#[from]` once and asserted the variant matched.
- Redundant tests already covered end-to-end through SQL.
- ~60 CLI protocol roundtrips in `src/cli/protocol.rs` that did `serde_json::to_string` then `from_str` then `matches!` with no literal wire assertion.

## What stayed

- DB CRUD tests that touch real SQL.
- Business-logic tests: timestamp guards, mute behavior, filtering, sorting.
- `thiserror` `#[error("...")]` Display tests that pin user-facing strings.
- Redaction-in-Debug tests for push tokens (a security property).
- Hand-written `Display`/`FromStr` with real mappings.
- CLI protocol tests that assert on a literal JSON string, exercise `#[serde(default)]` backward-compat, or cover the custom `MuteDuration` serde.
- `detect_media_type` magic-byte tests, exponential-backoff tests, MLS key-package semantic validation.

## Why

If a test breaks only when someone deliberately breaks it, it is not a seatbelt. Every deletion here either tested compiler-generated code, duplicated another test, or was symmetric with its own implementation (same constants in impl and test).

`just precommit-quick` green: fmt, clippy, docs, dead_code, tests.

## Tests

* **Tests**
  * Removed redundant unit and integration test cases across multiple modules to maintain code quality and testing efficiency
  * Added targeted test coverage for edge cases involving extreme integer ID values in database operations to ensure robust system behavior


src/cli/protocol.rs
```diff
@@ -435,30 +435,6 @@ mod tests {
...
-        let json = serde_json::to_string(&req).unwrap();
-        let parsed: Request = serde_json::from_str(&json).unwrap();
-        assert!(
-            matches!(parsed, Request::LoginWithCustomRelay { pubkey, relay_url }
-                if pubkey == "npub1xyz" && relay_url == "wss://relay.example.com")
-        );
-    }
-
     #[test]
     fn all_accounts_roundtrip() {
```
erskingardner: why did this one survive?

src/cli/protocol.rs
```diff
@@ -469,56 +445,6 @@ mod tests {
...
-            Request::CreateGroup { account, name, description, members }
-            if account == "npub1abc"
-                && name == "My Group"
-                && description.as_deref() == Some("A test group")
-                && members.len() == 2
-        ));
-    }
-
     #[test]
     fn create_group_minimal_roundtrip() {
```
erskingardner: this one too?

erskingardner (dismissed):
two small comments but I'm fine with this.

erskingardner:
CleanShot 2026-04-10 at 14 23 48

This about sums it up. 😂

src/cli/protocol.rs
dannym-arx: good question... lol fixing

erskingardner (approved):

mubarakcoded (approved):

PR #735, opened 2026-04-10
Architecture: Session and projection rearchitecture design doc

erskingardner:
## Summary

- Move architecture docs into `docs/` (relay-control-plane rearch doc was at repo root)
- Add the **session and projection rearchitecture** design doc for team review and discussion

This document proposes the next major architectural step after the relay control plane rearch. It covers:

- **AccountSession boundary** — account-scoped mutations go through a session handle where identity is baked in, not through a global facade that accepts pubkey arguments
- **Four state categories** — MDK authority, shared app state, account-scoped app state, and projections, each with declared ownership and scope rules
- **Singleton removal** — `Whitenoise` becomes a constructed value, not a process-wide global; shared services are independently constructible for testing
- **Per-account database split** — each account gets its own SQLite file; account deletion becomes `fs::remove_file`; account-scoped queries lose their `WHERE account_pubkey = ?` clauses entirely
- **Projection ownership** — each projection declares a builder (write side) and reader (read side) with explicit scope
- **Scoped relay handles** — sessions hold lightweight handles into shared relay infrastructure (group plane, ephemeral executor) that bake in identity and signer without duplicating connections
- **Services as views** — `session.messages()` returns a lightweight borrow-based view, avoiding Arc cycles and ownership complexity
- **Target workspace shape** — `whitenoise` (core), `whitenoise-cli` (CLI/daemon), `whitenoise-frb` (FRB wiring, extracted from Flutter repo)
- **Future: durable task runtime** — deterministic state machines for outbound operations that must survive crashes, enabled by the session architecture

The doc is intended as a high-level architecture overview for all engineers to digest before we move to a detailed implementation plan. Pseudocode examples in the appendix show initialization, session rehydration, relay handles, account-scoped operations, repositories, and testing.

## Discussion points

- Does the four-category state model (MDK / shared / account-scoped / projections) capture everything?
- Are the scoped relay handles the right abstraction for shared-infra-with-account-identity?
- Does the per-account DB split feel right, or is logical isolation (Path A) sufficient for now?
- Any concerns with the view pattern for account-scoped services?

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive architecture guidance for session and projection boundaries, ownership rules, storage/layout recommendations, migration sequencing, and phased refactor planning.
  * Added a detailed multi-phase implementation roadmap with validation steps, concurrency/scope rules, and a recommended borrow-based view pattern.
  * Added a testing strategy describing a three-tier test model and migration of scenarios across tiers.

* **Chores**
  * Documentation-only changes; no user-facing behavior changes.


docs/session-projection-rearchitecture.md
```diff
@@ -0,0 +1,838 @@
...
+// Shared reads go through the same facade
+let user = whitenoise.resolve_user(&pubkey).await?;
+```
+
+After:
+
+```rust
+// Account scope is baked into the session handle
+let session = whitenoise.session(account_pubkey)?;
+session.messages().send_to_group(&group_id, content).await?;
```
dannym-arx, line 202: I wonder if we could do 

session.groups().get(&group_id).send(content).await?

.get() in groups would return a singular group struct, and then operations happen on that. what do you think?

dannym-arx (dismissed):

docs/session-projection-rearchitecture.md
erskingardner: yeah - that's probably more sensible.

tbh, probably all message related stuff should go through groups?

dannym-arx: think so yeah

docs/session-projection-rearchitecture.md
```diff
@@ -0,0 +1,838 @@
...
+  → (writes message projection as above)
+  → chat_list_stream_manager.emit(account_pubkey, NewLastMessage)
+      → UI receives stream update
+      → UI calls ChatListService::get_chat_list() (the reader)
+      → ChatListService computes the list by joining source tables
+```
+
+The chat list "builder" is just the stream emission. The projection is computed fresh on each read from
+`accounts_groups` and `aggregated_messages` (account DB) plus `group_information` and `users` (shared DB). With the
+per-account DB split, this becomes an application-level join across two databases rather than a single SQL query — the
```
erskingardner, line 100: worth calling out here. this will result in less efficient queries that what you'd be able to do with complex joins. however, I think it will also result in much more readable/understandable code.

docs/session-projection-rearchitecture.md
mubarakcoded: yeah - I think this is one level deeper though, session.groups().get(&group_id)?.messages().send(content) rather than flattening everything onto the group struct directly. Our message API is too big to live on the group type without it getting huge. But agreed that the group should be the scope, not an argument.

docs/session-projection-rearchitecture.md
mubarakcoded: This is the right move, worth it.

docs/session-projection-rearchitecture.md
```diff
@@ -0,0 +1,838 @@
...
+**Each shared service is independently constructible.** For example,
+`UserDirectoryService::new(db: Arc<SharedDatabase>)`. Tests build only the services they need — no 25-field mock.
+
+**Per-account database.** Each `AccountSession` opens its own SQLite file. Account-scoped tables live there. Shared
+tables live in `shared.sqlite`. MDK keeps its existing per-account DB. Account deletion is trivial: drop the session,
+delete the file.
+
+**MDK cached on the session.** Today every call that needs MDK creates a fresh instance via
+`create_mdk_for_account()`, opening a new encrypted SQLite connection each time (~20+ call sites). The session holds a
+single `Arc<MDK<MdkSqliteStorage>>` for its lifetime. MDK is Send + Sync (its storage wraps rusqlite::Connection in
```
mubarakcoded, line 166: This is a concrete perf win on top of the architectural one ✌️

jgmontoya (commented):
I'm really in favor of moving away from the Whitenoise struct [god object](https://en.wikipedia.org/wiki/God_object) so that gets a thumbs up from me.

Other things that I like:
- Session handle
- Not having to call `create_mdk_for_account` everywhere so many times

On the other hand, I feel like the per-account db split is scary and might end up biting us. I believe there's actually a lot of complexity that's not necessarily obvious at first sight. Cross-db join complexity, loss of FK constraints, migrations burden (two different schemas for shared/account DBs, having to migrate each account db independently, etc..). I haven't really felt that having `WHERE`'s on our queries to be a huge problem.
It might not be a big deal, but it's something I'd defer.

Some questions
- How will event processor routing work? How do events reach the right `AccountSession`? 
- Who writes `group_information` rows? Given the "Shared Reads Rule" (sessions can read shared services but not write) it's not clear who writes them.

I'd like to understand better and see more about how relay deduplication and shared connections looks like on this new architecture.

The "Durable Task Runtime" I think is an orthogonal/independent thing. We could implement that with the current architecture. I'd take it out of the plan so we can focus on the actual architectural changes. I think the same about the "Target Workspace Shape". 

This got raised by my agent:

> View pattern (MessageOps<'_>) will not work in async Rust. Borrowed views cannot cross tokio::spawn boundaries, cannot be stored in structs, and are uncommon in production async codebases. The council unanimously recommends owned handles with Arc internals — either methods directly on Arc<AccountSession>, or typed domain handles (Arc<MessageHandle>).

dannym-arx:
> Cross-db join complexity

well the goal would be to not need cross db-joins haha

> How will event processor routing work? How do events reach the right AccountSession?

AccountManager would be responsible for that

> Who writes group_information rows? Given the "Shared Reads Rule" (sessions can read shared services but not write) it's not clear who writes them.

that's part of the shared state, so the session caller would trigger a function to the GroupInfoService

> View pattern (MessageOps<'_>) will not work in async Rust. Borrowed views cannot cross tokio::spawn boundaries, cannot be stored in structs, and are uncommon in production async codebases. The council unanimously recommends owned handles with Arc internals — either methods directly on Arc, or typed domain handles (Arc).

Hmmm that depends on the implementation, if every call to session.messages() returns a new instance it doesn't need to cross thread boundaries. the only thing that needs to be an arc is the session itself

jgmontoya:
> well the goal would be to not need cross db-joins haha

I hear the goal, but the doc's own ChatListService needs accounts_groups + aggregated_messages from account.sqlite AND group_information + users from shared.sqlite. How does the chat list get computed without accessing both databases? That's either an application-level join or denormalization, both have real costs.

> AccountManager would be responsible for that

👍

> that's part of the shared state, so the session caller would trigger a function to the GroupInfoService

👍

> Hmmm that depends on the implementation, if every call to session.messages() returns a new instance it doesn't need to cross thread boundaries. the only thing that needs to be an arc is the session itself

Right, but then what's the advantage over simply having methods on `AccountSession`?

docs/session-projection-rearchitecture.md
```diff
@@ -0,0 +1,838 @@
...
+├── account_<hex_pubkey_1>.sqlite    (account-scoped state)
+│   ├── account_settings
+│   ├── account_follows
+│   ├── accounts_groups
+│   ├── drafts
+│   ├── push_registrations
+│   ├── group_push_tokens
+│   ├── published_key_packages
+│   ├── published_events
+│   ├── aggregated_messages           (account-scoped message projection)
```
jgmontoya, line 296: Something I forgot to mention in my earlier comment: this makes it so the aggregated messages for the same group will be duplicated across all accounts that belong to that group

dannym-arx:
> I hear the goal, but the doc's own ChatListService needs accounts_groups + aggregated_messages from account.sqlite AND group_information + users from shared.sqlite. How does the chat list get computed without accessing both databases? That's either an application-level join or denormalization, both have real costs.

I think that the slight hit in performance of doing 2 queries on 2 dbs is perhaps worth it. 

> Right, but then what's the advantage over simply having methods on AccountSession?

Separation of concerns, tho I think we're gonna figure out what the best ergonomics are while building this. It may honestly be that it's actually better to have them on AccountSession

docs/session-projection-rearchitecture.md
dannym-arx: true, but this makes it easier to reason about, and it makes implementing things like signal's "delete for me" vs "delete for everyone" easier

erskingardner: yeah - the data storage here isn't really a big issue I don't think. most people have GBs of storage for whatsapp and have no idea.

erskingardner:
> > Right, but then what's the advantage over simply having methods on AccountSession?
> 
> Separation of concerns, tho I think we're gonna figure out what the best ergonomics are while building this. It may honestly be that it's actually better to have them on AccountSession

Yeah - SoC but also we should be able to test the logic in each of these *Ops type modules with the bare minimum of mocks and outside state. Open to the idea that we might decide to move things around a little here though.

erskingardner:
@dannym-arx @jgmontoya @mubarakcoded sounds like there are no major blockers here so maybe I'll move forward and add a few docs that are a little more implementation focused?

dannym-arx:
@erskingardner sounds good to me

mubarakcoded (dismissed):
LGTM

docs/session-projection-rearchitecture.md
```diff
@@ -0,0 +1,838 @@
...
+├── ephemeral: AccountEphemeralHandle             (scoped handle into shared infra)
+├── group_relay: AccountGroupHandle               (scoped handle into shared infra)
+└── cancellation: watch::Sender<bool>
+```
+
+### Key properties
+
+**No singleton.** `Whitenoise` is constructed at startup and passed by reference. `Whitenoise::get_instance()` is
+deleted. Entry points (FRB init, CLI daemon, test setup) each construct their own instance. Tests can run in parallel
+with isolated state.
```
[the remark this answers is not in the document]
dannym-arx: well the point of this is that it's a refactor, this doc is correct and takes precedence over your learnings

docs/session-projection-rearchitecture.md
```diff
@@ -0,0 +1,838 @@
...
+The current state is persisted to the account's database after every transition. On crash, the runtime scans the
+database for incomplete operations and resumes them from their last persisted state.
+
+```rust
+pub enum PublishState {
+    Pending { event: UnsignedEvent, target_relays: Vec<RelayUrl> },
+    Signing { event: UnsignedEvent },
+    Sending { event: Event, remaining: Vec<RelayUrl>, succeeded: Vec<RelayUrl> },
+    RetryWait { event: Event, remaining: Vec<RelayUrl>, attempt: u32, resume_at: Instant },
+    Complete { event: Event, accepted_by: Vec<RelayUrl> },
```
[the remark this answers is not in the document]
dannym-arx: not that it matters, but at that point let's just do u64; we don't need to worry about time travelers going back to the pre 1970 era.

but if there are time travelers reading this who disagree we'd like to know :rofl:

erskingardner:
Merging this now so that the agents can reference it for the coming PR wave.

mubarakcoded (dismissed):

PR #738, opened 2026-04-12
fix: add sleeps between follow mutations to avoid relay timestamp collisions

erskingardner:
## Changes

Add 2-second delays between follow/unfollow operations in the follow management integration test scenario.

## Why

Rapid consecutive follow/unfollow mutations can produce ContactList events with identical second-precision timestamps. When these events are sent to the relay in quick succession, the relay rejects newer events with "replaced: have newer event" errors, leaving the state inconsistent.

The delays ensure each ContactList event has a distinct timestamp, preventing relay rejection and maintaining correct state throughout the test scenario.

## Files Changed

- `src/integration_tests/scenarios/follow_management.rs`: Added `tokio::time::sleep` calls and necessary imports

## Summary by CodeRabbit

* **Tests**
  * Updated integration test timing to allow proper propagation of background processes in follow management scenarios.

**Note:** This release contains only internal test infrastructure improvements with no user-facing changes.

### DIFF
--- a/src/integration_tests/scenarios/follow_management.rs
+++ b/src/integration_tests/scenarios/follow_management.rs
@@ -1,10 +1,13 @@
+use std::time::Duration;
+
+use async_trait::async_trait;
+use nostr_sdk::Keys;
+
 use crate::integration_tests::{
     core::*,
     test_cases::{follow_management::*, shared::*},
 };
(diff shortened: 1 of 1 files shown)


dannym-arx (approved):
makes sense

mubarakcoded (approved):

PR #743, opened 2026-04-13
Add AccountSession and AccountManager scaffolding

dannym-arx:
marmot

This PR adds AccountSession and AccountManager, the first step toward replacing the global singleton with per-account session objects.

## What changed

**New module: `src/whitenoise/session/mod.rs`**

Two new types that give each account its own MDK instance and signer slot:

- `AccountSession` holds a pubkey, `Arc<MDK>`, and `RwLock<Option<Arc<dyn NostrSigner>>>`. External-signer accounts start with `None` in the signer slot until the platform signer is re-registered.
- `AccountManager` tracks active sessions and pending logins via `DashMap`. The `pending_logins` state moved here from `Whitenoise`.

**Lifecycle hooks across 6 existing files:**

- `activate_account` and `activate_account_without_publishing` insert a session after setup completes
- `logout` removes the session before tearing down subscriptions
- `register_external_signer` fills the session signer slot so operations through `AccountSession` can sign
- `restore_sessions` runs at startup before subscription setup, creating sessions for all persisted accounts
- `Whitenoise::session(pubkey)` provides the public lookup

**Moved state:**

`pending_logins` moved from a standalone `DashMap` on `Whitenoise` to `AccountManager.pending_logins`. All references in `login_multistep.rs`, `login_external_signer.rs`, and test code updated.

## What didn't change

- `create_mdk_for_account()` remains as a compatibility fallback
- No external API changes
- Existing "active session" checks via `background_task_cancellation` are untouched (switching to the session map comes next)

## Marmot note

Every marmot in the colony gets their own burrow entrance now. No more sharing a single tunnel for MLS key material.

## Summary by CodeRabbit

* **New Features**
  * Centralized per-account session management with automatic session restoration at startup

* **Improvements**
  * External signers now synchronize to active sessions for smoother signing
  * Multi-step login flow now tracks pending logins in centralized session store for more reliable completion
  * Background-task cancellation and concurrency controls tied to sessions for consistent behavior

* **Bug Fixes**
  * More reliable session cleanup during logout and startup/restores

### DIFF
--- a/src/whitenoise/accounts/login.rs
+++ b/src/whitenoise/accounts/login.rs
@@ -195,13 +195,10 @@ impl Whitenoise {
                 Vec::new()
             });
 
-        self.account_manager.remove_session(pubkey);
-
-        // Cancel any running background tasks (contact list user fetches, etc.)
-        // before tearing down subscriptions and relay connections.
-        if let Some((_, cancel_tx)) = self.background_task_cancellation.remove(pubkey) {
-            let _ = cancel_tx.send(true);
+        if let Some(session) = self.account_manager.get_session(pubkey) {
+            let _ = session.cancellation.send(true);
         }
+        self.account_manager.remove_session(pubkey);
 
         // Unsubscribe from account-specific subscriptions before logout
         self.relay_control
@@ -562,34 +559,33 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn test_create_identity_creates_cancellation_channel() {
+    async fn test_create_identity_creates_session() {
         let (whitenoise, _data_temp, _logs_temp) = create_mock_whitenoise().await;
 
         let account = whitenoise.create_identity().await.unwrap();
 
         assert!(
             whitenoise
-                .background_task_cancellation
-                .contains_key(&account.pubkey),
-            "activate_account should create a cancellation channel"
+                .account_manager
+                .get_session(&account.pubkey)
+                .is_some(),
+            "activate_account should create a session"
         );
     }
 
     #[tokio::test]
-    async fn test_logout_signals_and_removes_cancellation_channel() {
+    async fn test_logout_signals_cancellation_and_removes_session() {
         let (whitenoise, _data_temp, _logs_temp) = create_mock_whitenoise().await;
 
         let account = whitenoise.create_identity().await.unwrap();
 
-        // Subscribe to the cancellation channel before logout
         let cancel_rx = whitenoise
-            .background_task_cancellation
-            .get(&account.pubkey)
-            .expect("cancellation channel should exist after login")
-            .value()
+            .account_manager
+            .get_session(&account.pubkey)
+            .expect("session should exist after login")
+            .cancellation
             .subscribe();
 
-        // Initially not cancelled
         assert!(
             !*cancel_rx.borrow(),
             "should not be cancelled before logout"
@@ -597,18 +593,17 @@ mod tests {
 
         whitenoise.logout(&account.pubkey).await.unwrap();
 
-        // After logout, the channel should have been signalled
         assert!(
             *cancel_rx.borrow(),
             "logout should signal cancellation to running background tasks"
         );
 
-        // And the entry should be removed from the map
         assert!(
-            !whitenoise
-                .background_task_cancellation
-                .contains_key(&account.pubkey),
-            "logout should remove the cancellation channel entry"
+            whitenoise
+                .account_manager
+                .get_session(&account.pubkey)
+                .is_none(),
+            "logout should remove the session"
         );
     }
 
--- a/src/whitenoise/accounts/login.rs
+++ b/src/whitenoise/accounts/login.rs
@@ -195,6 +195,8 @@ impl Whitenoise {
                 Vec::new()
             });
 
+        self.account_manager.remove_session(pubkey);
+
         // Cancel any running background tasks (contact list user fetches, etc.)
         // before tearing down subscriptions and relay connections.
         if let Some((_, cancel_tx)) = self.background_task_cancellation.remove(pubkey) {
--- a/src/whitenoise/accounts/login.rs
+++ b/src/whitenoise/accounts/login.rs
@@ -195,11 +195,14 @@ impl Whitenoise {
                 Vec::new()
             });
 
+        // Signal cancellation first so in-flight handlers (e.g. contact-list
+        // guard) see the flag, but keep the session visible until subscription
+        // teardown completes — handlers that check get_session() during teardown
+        // need the entry to exist.
         if let Some(session) = self.account_manager.get_session(pubkey) {
-            let _ = session.cancellation.send(true);
+            session.cancel();
         }
 
-        // Unsubscribe from account-specific subscriptions before logout
         self.relay_control
             .deactivate_account_subscriptions(pubkey)
             .await;
@@ -584,8 +587,7 @@ mod tests {
             .account_manager
             .get_session(&account.pubkey)
             .expect("session should exist after login")
-            .cancellation
-            .subscribe();
+            .subscribe_cancellation();
 
         assert!(
             !*cancel_rx.borrow(),
--- a/src/whitenoise/accounts/login.rs
+++ b/src/whitenoise/accounts/login.rs
@@ -198,12 +198,13 @@ impl Whitenoise {
         if let Some(session) = self.account_manager.get_session(pubkey) {
             let _ = session.cancellation.send(true);
         }
-        self.account_manager.remove_session(pubkey);
 
         // Unsubscribe from account-specific subscriptions before logout
         self.relay_control
             .deactivate_account_subscriptions(pubkey)
             .await;
+
+        self.account_manager.remove_session(pubkey);
 
         if !ephemeral_warm_relays.is_empty()
             && let Err(error) = self
--- a/src/whitenoise/accounts/setup.rs
+++ b/src/whitenoise/accounts/setup.rs
@@ -1051,6 +1051,7 @@ mod tests {
     async fn reset_singleton_whitenoise_for_test(whitenoise: &Whitenoise) {
         whitenoise.background_task_cancellation.clear();
         whitenoise.external_signers.clear();
+        whitenoise.account_manager.clear_sessions();
         whitenoise.account_manager.pending_logins.clear();
         whitenoise.reset_nostr_client().await.unwrap();
         whitenoise.wipe_database().await.unwrap();
--- a/src/whitenoise/accounts/setup.rs
+++ b/src/whitenoise/accounts/setup.rs
@@ -1,4 +1,5 @@
 use std::collections::HashSet;
+use std::sync::Arc;
 
 use dashmap::mapref::entry::Entry;
 use mdk_core::prelude::group_types::GroupState;
@@ -11,6 +12,7 @@ use crate::relay_control::groups::GroupSubscriptionSpec;
 use crate::whitenoise::database::published_key_packages::PublishedKeyPackage;
 use crate::whitenoise::error::Result;
 use crate::whitenoise::relays::Relay;
+use crate::whitenoise::session::AccountSession;
 use crate::whitenoise::users::User;
 use crate::whitenoise::{Whitenoise, WhitenoiseError};
 
@@ -46,6 +48,33 @@ impl Whitenoise {
                 );
             }
         }
+    }
+
+    /// Create and insert an `AccountSession` into the `AccountManager`.
+    ///
+    /// Requires the global singleton for the `&'static Whitenoise` back-reference.
+    /// In unit tests (where the singleton is not set) this is a no-op.
+    fn insert_account_session(&self, account: &Account) -> Result<()> {
+        let wn = match Self::get_instance() {
+            Ok(wn) => wn,
+            Err(_) => {
+                tracing::debug!(
+                    target: "whitenoise::accounts",
+                    "Skipping session insertion (global singleton not available, likely in tests)"
+                );
+                return Ok(());
+            }
+        };
+
+        let mdk = self.create_mdk_for_account(account.pubkey)?;
+
+        // Resolve the signer: for local accounts, load from secrets store.
+        // For external accounts, check the external signers map.
+        let signer = self.get_signer_for_account(account).ok();
+
+        let session = Arc::new(AccountSession::new(account.pubkey, mdk, signer, wn));
+        self.account_manager.insert_session(session);
+        Ok(())
     }
 
     #[perf_instrument("accounts")]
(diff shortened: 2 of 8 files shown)


src/whitenoise/session/mod.rs
```diff
@@ -0,0 +1,146 @@
...
+///
+/// `signer` is `None` for restored external-signer accounts whose platform
+/// signer has not yet been re-registered. Operations requiring signing should
+/// return the existing external-signer-unavailable error until
+/// `register_external_signer()` fills the slot.
+pub struct AccountSession {
+    pub account_pubkey: PublicKey,
+    pub mdk: Arc<MDK<MdkSqliteStorage>>,
+    pub signer: RwLock<Option<Arc<dyn NostrSigner>>>,
+}
```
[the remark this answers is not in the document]
dannym-arx: this is part of the refactor. this is correct and expected

dannym-arx:
@claude oh claude, keen claude, with thoughtful view
please do a final full review
check every corner, old an new,
and guide this work till it rings true 

https://github.com/user-attachments/assets/1cae67c3-99ad-451d-995f-fe180494fc28

src/whitenoise/session/mod.rs
erskingardner, line 1: 🥹 it's... beautiful.

src/whitenoise/session/mod.rs
```diff
@@ -0,0 +1,168 @@
...
+use crate::whitenoise::Whitenoise;
+use crate::whitenoise::accounts::DiscoveredRelayLists;
+
+/// A per-account session holding the account's MDK instance and signer.
+///
+/// `signer` is `None` for restored external-signer accounts whose platform
+/// signer has not yet been re-registered. Operations requiring signing should
+/// return the existing external-signer-unavailable error until
+/// `register_external_signer()` fills the slot.
+pub struct AccountSession {
```
mubarakcoded, line 20: What about user_resolution_guards? Still sitting on Whitenoise as a DashMap.

src/whitenoise/accounts/setup.rs
```diff
@@ -93,6 +76,9 @@ impl Whitenoise {
             );
         }
         sub_result?;
+
+        self.insert_account_session(account)?;
```
mubarakcoded: Previously we set up the cancellation channel before subscriptions and had ensure_background_task_cancellation_channel as a fallback in setup_subscriptions. Both are gone and the session is now created after subscriptions. Was flipping that order intentional? Anything that needs the session during setup (like the contact list guard) won't find one.

src/whitenoise/session/mod.rs
dannym-arx: It's an interesting one because it's not per-account and serializes metadata fetches for a given pubkey globally, so multiple accounts resolving the same user don't hammer relays in parallel.

That means it doesn't naturally fit on AccountSession. Want me to extract a UserResolver struct on this PR or do that as part of another phase?

src/whitenoise/accounts/setup.rs
dannym-arx: fixing it

EDIT: addressed in 9442fd5b77d9f6a79b93178adc9312c61786853

src/whitenoise/session/mod.rs
mubarakcoded: Makes sense to leave it for now and pull it into the UserResolver at the later phase

src/whitenoise/session/mod.rs
```diff
@@ -0,0 +1,168 @@
...
+        *self.signer.write().await = Some(signer);
+    }
+}
+
+/// Manages active account sessions and pending logins.
+pub struct AccountManager {
+    sessions: DashMap<PublicKey, Arc<AccountSession>>,
+    /// Pubkeys with a login in progress. The value holds whichever relay lists
+    /// were already discovered so step 2a can publish defaults only for missing ones.
+    pub(crate) pending_logins: DashMap<PublicKey, DiscoveredRelayLists>,
```
erskingardner: Adds a little code but I agree with the agent on this one: 

> Call sites in `login_multistep.rs` and `login_external_signer.rs` reach directly into `account_manager.pending_logins.insert(...)`, `account_manager.pending_logins.get(...)`, etc. — there are at least 15 such direct field accesses across the diff. The refactor doc describes `AccountManager` as an evolving owner of session state; exposing the underlying `DashMap` means every caller is coupled to the storage representation. Consider adding named methods (`stash_pending_login`, `take_pending_login`, `cancel_pending_login`) and keeping the field private, the same way `sessions` is kept private.
> 
> This is directionally aligned with where the refactor is going anyway and costs nothing to do now while the callers are already being touched.

src/whitenoise/session/mod.rs
```diff
@@ -0,0 +1,168 @@
...
+use crate::whitenoise::accounts::DiscoveredRelayLists;
+
+/// A per-account session holding the account's MDK instance and signer.
+///
+/// `signer` is `None` for restored external-signer accounts whose platform
+/// signer has not yet been re-registered. Operations requiring signing should
+/// return the existing external-signer-unavailable error until
+/// `register_external_signer()` fills the slot.
+pub struct AccountSession {
+    pub account_pubkey: PublicKey,
```
erskingardner, line 21: I feel like we want as few public members here as possible: 

> `cancellation` being `pub` means any caller can send `true` or `false` — the session can be cancelled by anyone who holds an `Arc<AccountSession>`. `contact_list_guard` being `pub` means any caller can call `acquire_owned()` directly or even close the semaphore. These are implementation details of the session that should be encapsulated behind methods. The `mdk` field is likely fine as `pub` since MDK is the primary reason to hold a session reference. Consider:
> 
> - `cancellation`: make private, expose `subscribe_cancellation() -> watch::Receiver<bool>` and an internal `cancel()` method called only by `AccountManager::remove_session`.
> - `contact_list_guard`: make private, expose `acquire_contact_list_permit() -> Result<OwnedSemaphorePermit>` so that callers don't need to know about the `Arc<Semaphore>` wrapping.
> 
> This also makes it easier to change the internal representation (e.g. switch to `ArcSwap` for the signer, or add metrics on semaphore wait time) without touching callers.

src/whitenoise/accounts/setup.rs
```diff
@@ -11,41 +11,26 @@ use crate::relay_control::groups::GroupSubscriptionSpec;
...
-            Entry::Vacant(entry) => {
-                let (cancel_tx, _) = watch::channel(false);
-                entry.insert(cancel_tx);
-                tracing::debug!(
-                    target: "whitenoise::accounts",
-                    account_pubkey = %account_pubkey,
-                    "Created missing background-task cancellation channel during subscription setup",
-                );
-            }
+    fn insert_account_session(&self, account: &Account) -> Result<()> {
```
erskingardner, line 21: > **The `signer` field initialization in `insert_account_session` duplicates `restore_sessions`.** Both call `create_mdk_for_account` then optionally `get_signer_for_account`. The pattern is identical. A private `AccountSession::from_account(account, wn)` constructor (or a free function) would give both paths a single place to change when the logic evolves.

src/whitenoise/accounts/login.rs
```diff
@@ -195,17 +195,17 @@ impl Whitenoise {
                 Vec::new()
             });
 
-        // Cancel any running background tasks (contact list user fetches, etc.)
-        // before tearing down subscriptions and relay connections.
-        if let Some((_, cancel_tx)) = self.background_task_cancellation.remove(pubkey) {
-            let _ = cancel_tx.send(true);
+        if let Some(session) = self.account_manager.get_session(pubkey) {
+            let _ = session.cancellation.send(true);
```
erskingardner: > Between `send(true)` and `remove_session`, there is an `await` point (`deactivate_account_subscriptions`). Any code that checks `account_manager.get_session` in that window — such as `acquire_contact_list_guard` in `handle_contact_list.rs` — will find a session that has been cancelled but not yet removed. That is mostly fine, but it creates a subtle window where the session exists but the cancellation is already signalled. The old code had the same ordering (remove was atomic with send), so this is not a regression, but now that the session is the canonical presence check for "is this account active", the session should probably be removed before or right at the point of send, not after the await chain. Worth a comment at minimum to document the intentional ordering.

src/whitenoise/signer.rs
```diff
@@ -118,7 +124,18 @@ impl Whitenoise {
     ///
     /// Returns an error if no signer is available for the account.
     pub(crate) fn get_signer_for_account(&self, account: &Account) -> Result<Arc<dyn NostrSigner>> {
-        // First check for a registered external signer
+        if let Some(session) = self.account_manager.get_session(&account.pubkey)
+            && let Some(signer) = session.signer.try_read().ok().and_then(|g| g.clone())
```
erskingardner: this feels a bit nitpicky but I'll let you make the call: 

> `try_read()` returns `Err` when a write lock is held — specifically, when `set_signer` is concurrently running. When that happens, `.ok()` turns the error into `None`, the `if let` arm is not taken, and the code falls through to the legacy `external_signers` DashMap lookup. That's the old code path and it will return the same signer, so the functional impact is low for external-signer accounts. But for a local-key account that has a session signer and no entry in `external_signers`, the fallthrough hits the secrets-store lookup, which works but adds latency and a redundant syscall under exactly the contention that should never be observable in practice.
> 
> More fundamentally, using `try_read` to avoid an `await` boundary on a sync function means this function's behaviour is silently load-dependent. The right fix is one of:
> 
> 1. Make `get_signer_for_account` async and use `.read().await`.
> 2. Don't put the signer behind an `RwLock` — store it as an `ArcSwap<Option<Arc<dyn NostrSigner>>>` so read access is lock-free and always succeeds.
> 
> Option 2 is the cleaner long-term shape and aligns better with the zero-overhead read path the refactor doc describes.

src/whitenoise/session/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+    /// Return a snapshot of the pending-login stash for `pubkey`.
+    pub fn get_pending_login(&self, pubkey: &PublicKey) -> Option<DiscoveredRelayLists> {
+        self.pending_logins.get(pubkey).map(|r| r.clone())
+    }
+
+    /// Remove and return the pending-login stash for `pubkey`.
+    pub fn take_pending_login(
+        &self,
+        pubkey: &PublicKey,
+    ) -> Option<(PublicKey, DiscoveredRelayLists)> {
```
mubarakcoded: This fn returns Option<(PublicKey, DiscoveredRelayLists)> - that's just what DashMap::remove hands back. But get_pending_login on line 154 returns just the value. Callers have to destructure the tuple for no reason, worth unwrapping to Option<DiscoveredRelayLists> to keep them consistent.

src/whitenoise/session/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+            .map_err(|_| {
+                WhitenoiseError::ContactList(
+                    "Failed to acquire contact list processing permit".to_string(),
+                )
+            })
+    }
+}
+
+/// Manages active account sessions and pending logins.
+pub struct AccountManager {
```
mubarakcoded, line 95: No unit tests for this module, I think since this is the base everything else builds on, would be good to have a few focused tests here.

mubarakcoded (changes requested):

src/whitenoise/session/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+pub struct AccountSession {
+    pub account_pubkey: PublicKey,
+    pub mdk: Arc<MDK<MdkSqliteStorage>>,
+    signer: RwLock<Option<Arc<dyn NostrSigner>>>,
+    contact_list_guard: Arc<Semaphore>,
+    cancellation: watch::Sender<bool>,
+}
+
+impl AccountSession {
+    pub fn new(
```
erskingardner: Should this be `pub(crate)` instead?

src/whitenoise/session/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+        pubkey: &PublicKey,
+    ) -> Option<(PublicKey, DiscoveredRelayLists)> {
+        self.pending_logins.remove(pubkey)
+    }
+
+    /// Merge newly-discovered relay lists into the pending-login stash and
+    /// return a snapshot. The DashMap lock is released before returning so
+    /// the caller is free to do async work with the result.
+    ///
+    /// Returns `NoLoginInProgress` if no stash exists for `pubkey`.
```
erskingardner: comment is incorrect here. it returns an Option now

src/whitenoise/session/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+    fn default() -> Self {
+        Self {
+            sessions: DashMap::new(),
+            pending_logins: DashMap::new(),
+        }
+    }
+}
+
+impl AccountManager {
+    pub fn insert_session(&self, session: Arc<AccountSession>) {
```
erskingardner: I think the suggestion here is valid - we should have the minimum of public API surface and no [semi] duplication of how to do things. 

> `insert_account_session` (in `setup.rs`) correctly cancels the old session before calling
> `insert_session`. But `insert_session` is `pub` — any future caller can replace a session
> without triggering cancellation and will leave background tasks for the old session running
> indefinitely. The invariant "cancel before insert" belongs inside `insert_session`, not in
> its caller. Either:
> 
> - Move the cancel-if-present logic into `insert_session` (and rename it
>   `replace_session` or keep the name), or
> - Make `insert_session` `pub(crate)` and keep `insert_account_session` as the
>   sole production entry point.
> 
> The current split will likely cause a bug when the next caller of `insert_session` is
> added.

src/whitenoise/session/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+                target: "whitenoise::session",
+                "Removed session for {}",
+                pubkey.to_hex()
+            );
+        }
+        removed
+    }
+
+    /// Remove all active sessions.
+    pub fn clear_sessions(&self) {
```
erskingardner, line 142: does this also need to do the work of cleaning up background tasks and other stuff? I like that this only has local knowledge and responsibility - just want to make sure we're not leaving a footgun around.

src/whitenoise/session/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+        self.sessions.clear();
+    }
+
+    /// Record that a multi-step login is in progress for `pubkey`.
+    pub fn stash_pending_login(&self, pubkey: PublicKey, discovered: DiscoveredRelayLists) {
```
erskingardner: This is the only one of these methods that takes `PublicKey` instead of `&PublicKey` is that needed/intentional?

src/whitenoise/session/mod.rs
dannym-arx: no, not at this layer imo

erskingardner (approved):

mubarakcoded (approved):

PR #746, opened 2026-04-13
Rate-limit inbound MIP-05 token requests and removals

### LINKED ISSUE #704: No rate limiting on incoming MIP-05 token requests — any group member can trigger unbounded relay publishing
## Summary

When a MIP-05 token request arrives from a group member there is no per-sender rate limit. Any group member can send unlimited token request events, each with a unique `event.id`, causing the recipient to look up their push tokens, encrypt them, and publish an MLS-encrypted reply to group relays — with no cap on frequency.

## Impact

A malicious or compromised group member can continuously send token requests, forcing every other member to:

1. Write to the `group_push_tokens` database table on every request.
2. Spawn a new background task per request (see related unbounded-spawn issue).
3. Encrypt the local push token and publish an MLS message to group relays — consuming relay write quota and running MLS crypto (HKDF + ChaCha20) on every response.

In a group with many members, a single attacker triggering responses from all members amplifies the relay publishing cost across the whole group.

## Affected location

| File | Lines | Symbol |
|---|---|---|
| `src/whitenoise/push_notifications.rs` | 406–438 | `handle_received_push_group_message` — no rate check before processing |
| `src/whitenoise/push_notifications.rs` | 507–540 | `schedule_pending_token_response` — spawns task per request |

## Evidence

```rust
// push_notifications.rs:406 — no rate limiting before processing token request
Mip05GroupMessage::TokenRequest(request) => {
    let leaf_index = sender_leaf_index.ok_or_else(|| { ... })?;

    self.merge_token_request(          // DB write per request
        account,
        &message.mls_group_id,
        message.event.pubkey,
        leaf_index,
        request,
    ).await?;

    if let Some(request_event_id) = message.event.id {
        self.schedule_pending_token_response(  // spawns tokio task per request
            account.clone(),
            message.mls_group_id.clone(),
            request_event_id,
        );
    }
}
```

There is no check on how many requests a given `(group_id, sender_leaf_index)` pair has generated within any time window before executing these operations.

## Suggested fix

Add a per-sender rate limiter keyed on `(mls_group_id, sender_leaf_index)`. Token requests are rare events in normal operation (they happen when a user first registers for push notifications or joins a group), so a limit of 5–10 per hour per sender is generous:

```rust
// Before the merge_token_request call:
if !self.token_request_rate_limiter
    .check_and_increment(&(message.mls_group_id.clone(), leaf_index))
    .await
    .is_allowed()
{
    tracing::debug!(
        target: "whitenoise::push_notifications",
        "Rate limiting MIP-05 token request from leaf {}",
        leaf_index
    );
    return Ok(false);
}
```

mubarakcoded:
## Summary

Closes #704

- Adds a per-sender rate limiter (`DashMap<(PublicKey, GroupId, u32), Instant>`) that drops duplicate MIP-05 token requests and removals arriving within a 30-second cooldown window
- Uses DashMap's `entry()` API for atomic check-and-update under a single shard lock
- Gates both `TokenRequest` and `TokenRemoval` message types — prevents DB spam from either vector
- Key includes `account_pubkey` so multi-account devices don't bleed rate limits across accounts

## Test plan

- [x] `test_rate_limiter_allows_first_request` — first request passes
- [x] `test_rate_limiter_blocks_rapid_duplicate` — immediate repeat blocked
- [x] `test_rate_limiter_allows_different_senders` — different leaf index unaffected
- [x] `test_rate_limiter_allows_different_groups` — different group unaffected
- [x] `test_rate_limiter_allows_different_accounts` — different account unaffected
- [x] `test_rate_limiter_allows_after_cooldown` — request after 30s allowed
- [x] `test_rate_limiter_wires_into_message_handler` — end-to-end handler integration
- [x] `just check-fmt` passes
- [x] `just check-clippy` passes

## Summary by CodeRabbit

* **New Features**
  * Per-account/group rate limiting for token requests and removals with a 30‑second cooldown; requests and removals use separate cooldown buckets to avoid cross-suppression and duplicate handling.

* **Bug Fixes / Cleanup**
  * Cleared token-rate entries for an account on logout to prevent stale cooldown state.

* **Tests**
  * Added tests for allowance, rapid-duplicate rejection, parameter isolation, prevention of duplicate responses, and cooldown expiry.

### DIFF
--- a/src/whitenoise/mod.rs
+++ b/src/whitenoise/mod.rs
@@ -187,8 +187,10 @@ pub struct Whitenoise {
     /// See [`push_notifications::MAX_CONCURRENT_TOKEN_RESPONSE_TASKS`] for the cap value.
     token_response_semaphore: Arc<Semaphore>,
     /// Per-sender rate limiter for inbound MIP-05 token requests and removals.
-    /// Maps `(account_pubkey, GroupId, sender_leaf_index)` → last accepted `Instant`.
-    token_request_timestamps: DashMap<(PublicKey, GroupId, u32), std::time::Instant>,
+    /// Maps `(account_pubkey, GroupId, sender_leaf_index, kind)` → last accepted
+    /// `Instant`.  Requests and removals have independent cooldown buckets.
+    token_request_timestamps:
+        DashMap<(PublicKey, GroupId, u32, push_notifications::TokenRateKind), std::time::Instant>,
 }
 
 static GLOBAL_WHITENOISE: OnceCell<Whitenoise> = OnceCell::const_new();
--- a/src/whitenoise/mod.rs
+++ b/src/whitenoise/mod.rs
@@ -186,6 +186,9 @@ pub struct Whitenoise {
     /// Bounds the number of concurrently-active delayed MIP-05 token-list response tasks.
     /// See [`push_notifications::MAX_CONCURRENT_TOKEN_RESPONSE_TASKS`] for the cap value.
     token_response_semaphore: Arc<Semaphore>,
+    /// Per-sender rate limiter for inbound MIP-05 token requests and removals.
+    /// Maps `(account_pubkey, GroupId, sender_leaf_index)` → last accepted `Instant`.
+    token_request_timestamps: DashMap<(PublicKey, GroupId, u32), std::time::Instant>,
 }
 
 static GLOBAL_WHITENOISE: OnceCell<Whitenoise> = OnceCell::const_new();
@@ -224,6 +227,7 @@ impl std::fmt::Debug for Whitenoise {
             .field("scheduler_handles", &"<REDACTED>")
             .field("pending_push_token_responses", &"<REDACTED>")
             .field("token_response_semaphore", &"<REDACTED>")
+            .field("token_request_timestamps", &"<REDACTED>")
             .finish()
     }
 }
@@ -273,6 +277,7 @@ impl Whitenoise {
             token_response_semaphore: Arc::new(Semaphore::new(
                 push_notifications::MAX_CONCURRENT_TOKEN_RESPONSE_TASKS,
             )),
+            token_request_timestamps: DashMap::new(),
         }
     }
 
--- a/src/whitenoise/push_notifications.rs
+++ b/src/whitenoise/push_notifications.rs
@@ -37,11 +37,18 @@ use crate::{
     relay_control::{RelayControlPlane, ephemeral::EphemeralPlane},
 };
 
-/// Minimum interval between accepted MIP-05 token requests from the same
-/// `(group_id, sender_leaf_index)` pair.  Requests that arrive before the
-/// cooldown expires are silently dropped.  Normal usage is ≤5 requests per
-/// hour, so 30 seconds is generous while blocking sustained spam.
+/// Minimum interval between accepted MIP-05 token messages of the same kind
+/// from the same `(account, group_id, sender_leaf_index)` tuple.  Messages
+/// that arrive before the cooldown expires are silently dropped.  Normal usage
+/// is ≤5 requests per hour, so 30 seconds is generous while blocking spam.
 const TOKEN_REQUEST_COOLDOWN: Duration = Duration::from_secs(30);
+
+/// Discriminator so token requests and removals have independent cooldowns.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub(crate) enum TokenRateKind {
+    Request,
+    Removal,
+}
 
 /// Maximum number of groups to publish push token events to concurrently.
 /// Caps relay connection pressure while still parallelising the bulk of the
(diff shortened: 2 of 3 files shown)


src/whitenoise/push_notifications.rs
```diff
@@ -36,6 +36,19 @@ use crate::{
     relay_control::{RelayControlPlane, ephemeral::EphemeralPlane},
 };
 
+/// Minimum interval between accepted MIP-05 token messages of the same kind
+/// from the same `(account, group_id, sender_leaf_index)` tuple.  Messages
+/// that arrive before the cooldown expires are silently dropped.  Normal usage
+/// is ≤5 requests per hour, so 30 seconds is generous while blocking spam.
+const TOKEN_REQUEST_COOLDOWN: Duration = Duration::from_secs(30);
```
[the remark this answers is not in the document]
mubarakcoded: MIP-05 token requests are MLS application messages - only authenticated group members can send them. A malicious member pacing at 31s intervals produces ~120 lightweight DB writes/hour, which is negligible compared to the damage they could do through other MLS operations (commit bombs, message spam). The 30s cooldown blocks the actual threat: burst flooding from duplicate/replayed events. A token-bucket adds complexity for a threat model that doesn't apply here.

src/whitenoise/mod.rs
```diff
@@ -186,6 +186,11 @@ pub struct Whitenoise {
     /// Bounds the number of concurrently-active delayed MIP-05 token-list response tasks.
     /// See [`push_notifications::MAX_CONCURRENT_TOKEN_RESPONSE_TASKS`] for the cap value.
     token_response_semaphore: Arc<Semaphore>,
+    /// Per-sender rate limiter for inbound MIP-05 token requests and removals.
+    /// Maps `(account_pubkey, GroupId, sender_leaf_index, kind)` → last accepted
+    /// `Instant`.  Requests and removals have independent cooldown buckets.
+    token_request_timestamps:
```
erskingardner, line 192: I think this is actually worth addressing. It's not really relevant yet but as soon as we have a long-running desktop client it will be an issue and we won't remember to fix it then I'll bet. 

> `token_request_timestamps` is a `DashMap` with no eviction strategy. Every unique `(account, group, leaf_index, kind)` tuple inserts an entry that is never removed. Over time — as group membership changes, accounts leave and rejoin groups, or leaf indices are recycled after MLS commits — this map accumulates stale entries indefinitely. For a long-running process with many groups and member churn this is a slow memory leak.
> 
> Counterpoint: the PR description acknowledges "normal usage is ≤5 requests per hour", so growth is slow in practice. But it will never shrink.
> 
> Consider one of:
> 1. A periodic `retain` pass that removes entries older than `TOKEN_REQUEST_COOLDOWN` (cheap, can be hooked into the existing scheduler).
> 2. Clearing all entries for a given `account_pubkey` on logout (parallels how `background_task_cancellation` is cleaned up in `login.rs`).
> 
> Option 2 is the minimum — entries for logged-out accounts are definitely stale. There is already a pattern for this in `logout()` (`self.background_task_cancellation.remove(pubkey)`). The `token_request_timestamps` map should get the same treatment.

src/whitenoise/push_notifications.rs
```diff
@@ -854,6 +897,37 @@ impl Whitenoise {
...
+        account_pubkey: &PublicKey,
+        group_id: &GroupId,
+        leaf_index: u32,
+        kind: TokenRateKind,
+    ) -> bool {
+        let key = (*account_pubkey, group_id.clone(), leaf_index, kind);
+        let now = Instant::now();
+
+        match self.token_request_timestamps.entry(key) {
+            dashmap::mapref::entry::Entry::Occupied(mut entry) => {
```
erskingardner: Agent flagged this and I agree, it'd be cleaner.

> The fully-qualified `dashmap::mapref::entry::Entry::` prefix is necessary because `entry()` on `DashMap` returns `dashmap::mapref::entry::Entry`, which isn't in scope. But the two arms can be collapsed using the `or_insert_with` / `and_modify` combinators, which would make this much more readable:
> 
> ```rust
> fn check_token_request_rate(&self, ...) -> bool {
>     let key = (*account_pubkey, group_id.clone(), leaf_index, kind);
>     let now = Instant::now();
> 
>     let mut entry = self.token_request_timestamps.entry(key).or_insert(now);
>     if entry.value() == &now {
>         // Vacant — we just inserted, this is the first request.
>         return true;
>     }
>     if now.duration_since(*entry.value()) < TOKEN_REQUEST_COOLDOWN {
>         return false;
>     }
>     *entry = now;
>     true
> }

erskingardner: You can also do it this way: 

```rust
let mut allowed = false;
self.token_request_timestamps
    .entry(key)
    .and_modify(|last| {
        if now.duration_since(*last) >= TOKEN_REQUEST_COOLDOWN {
            *last = now;
            allowed = true;
        }
    })
    .or_insert_with(|| { allowed = true; now });
allowed
```

erskingardner (approved):

PR #747, opened 2026-04-13
Add background notification collection API for iOS silent push

erskingardner:
## Summary

- Adds `Whitenoise::ensure_initialized()` — an idempotent entry point that handles warm, suspended, and cold start scenarios without risking duplicate event processors or background tasks
- Adds `collect_notifications_after_push()` — subscribes to the notification broadcast, refreshes relay subscriptions to fetch missed events, and collects `NotificationUpdate`s within a bounded quiet-window + hard deadline
- Adds C ABI wrapper (`wn_collect_notifications_after_push` / `wn_string_free`) returning JSON for direct Swift interop without Flutter/FRB

This is the Rust-side implementation for iOS silent push notification support. The native iOS `BackgroundPushCoordinator` will call the C ABI function to collect notifications and schedule local `UNNotification`s.

## Design

The API handles three iOS app states with a single code path:
1. **Warm** (app active) — `ensure_initialized()` returns instantly, collection window likely yields no data (Flutter already handling)
2. **Suspended** (process frozen by iOS) — `ensure_initialized()` returns instantly, `ensure_all_subscriptions()` reconnects dead relays and fetches missed events
3. **Cold** (process terminated) — full Whitenoise initialization (~500-900ms), then normal collection

Cold-start benchmarks show initialization completes well within the iOS 30-second background execution budget.

## Test plan

- [x] Unit tests for JSON serialization of all result status variants (`new_data`, `no_data`, `failed`)
- [x] `just precommit-quick` passes
- [ ] Integration testing with iOS silent push on device (future: app repo work)

## Summary by CodeRabbit

* **New Features**
  * Background notification collection after push events with a configurable wait timeout.
  * Native JSON-based interface for other apps to trigger background collection and receive results.
  * Stable, public notification result and status types exposed for downstream use.
  * Idempotent initialization helper to ensure safe startup of the notification system.

* **Tests**
  * Unit tests validating JSON shapes, trigger mapping, draining behavior, and native-interface handling.

### DIFF
--- a/src/ffi.rs
+++ b/src/ffi.rs
@@ -176,13 +176,16 @@ fn result_to_json(result: BackgroundNotificationResult) -> String {
 /// or CString construction fails for any reason, the static fallback JSON is
 /// used instead.
 fn string_to_raw(json: String) -> *mut c_char {
-    // CString::new rejects interior NULs. First try the happy path.
-    if let Ok(cstr) = CString::new(json.clone()) {
-        return cstr.into_raw();
-    }
+    // CString::new rejects interior NULs. First try the happy path. On failure
+    // it hands back the original bytes via NulError::into_vec(), so we can
+    // recover ownership without needing to clone up front.
+    let bytes = match CString::new(json) {
+        Ok(cstr) => return cstr.into_raw(),
+        Err(nul_err) => nul_err.into_vec(),
+    };
 
     // Interior NUL present. Strip NUL bytes and retry.
-    let sanitized: String = json.chars().filter(|c| *c != '\0').collect();
+    let sanitized: Vec<u8> = bytes.into_iter().filter(|b| *b != 0).collect();
     if let Ok(cstr) = CString::new(sanitized) {
         return cstr.into_raw();
     }
--- a/src/ffi.rs
+++ b/src/ffi.rs
@@ -87,7 +87,16 @@ pub unsafe extern "C" fn wn_collect_notifications_after_push(
         );
         let max_wait = Duration::from_millis(u64::from(max_wait_ms));
 
-        let rt = match tokio::runtime::Runtime::new() {
+        // Use a current-thread runtime: this FFI call is a bounded, single
+        // async operation driven to completion via block_on. A multi-threaded
+        // runtime would spawn worker threads we don't need for a background
+        // iOS push handler. `enable_all()` keeps net + time + the blocking
+        // worker pool available for SQLite and any spawn_blocking callers
+        // down the call stack.
+        let rt = match tokio::runtime::Builder::new_current_thread()
+            .enable_all()
+            .build()
+        {
             Ok(rt) => rt,
             Err(e) => {
                 return result_to_json(BackgroundNotificationResult::failed(format!(
--- a/src/ffi.rs
+++ b/src/ffi.rs
@@ -1,19 +1,51 @@
 //! C ABI entrypoints for native iOS background push handling.
 //!
-//! These functions are callable from Swift via the static library without
-//! requiring Flutter or FRB. The interface is intentionally minimal: pass in
-//! directory paths and a timeout, get back a JSON string.
+//! Callable from Swift via the static library without requiring Flutter or FRB.
+//!
+//! # Design — why callback + length-prefixed bytes
+//!
+//! The FFI surface is designed for safety across the Rust/Swift boundary:
+//!
+//! * **Length-prefixed byte slices on input.** All string inputs are passed as
+//!   `(ptr, len)` pairs rather than NUL-terminated C strings. This avoids
+//!   unbounded scanning into foreign memory and keeps inputs binary-safe.
+//!
+//! * **Callback for output.** JSON results are delivered via a callback
+//!   invoked while Rust still owns the bytes. The caller can read (and copy)
+//!   during the callback; after the callback returns, the buffer is freed
+//!   deterministically by Rust. There is no `into_raw` / `from_raw` ownership
+//!   handoff across the boundary, so use-after-free and double-free are
+//!   structurally impossible.
 //!
 //! # Contract guarantees
 //!
-//! * [`wn_collect_notifications_after_push`] **never returns NULL** under any
-//!   circumstances — callers can unconditionally treat the result as a valid,
-//!   NUL-terminated UTF-8 C string.
-//! * It **never panics across the FFI boundary** — all internal panics are
-//!   caught and converted to a failure JSON response.
-//! * It **always returns valid JSON** conforming to the documented shape.
+//! * [`wn_collect_notifications_after_push`] **never panics across the FFI
+//!   boundary on outbound Rust→Swift unwinds** — all internal panics are
+//!   caught and surfaced as a failure JSON callback invocation.
+//! * It **always invokes the callback exactly once** before returning,
+//!   regardless of init/runtime/serialization failures.
+//! * The JSON payload delivered to the callback **always** conforms to:
+//!   ```json
+//!   {
+//!     "status": "new_data" | "no_data" | "failed",
+//!     "notifications": [ ... ],
+//!     "error": "..."   // only present when status is "failed"
+//!   }
+//!   ```
+//!
+//! # Caller responsibilities
+//!
+//! The caller must ensure the supplied [`WnJsonCallback`] does not panic or
+//! unwind. Panicking across an `extern "C"` boundary is undefined behavior in
+//! Rust and in practice will abort the process. On Darwin, Swift runtime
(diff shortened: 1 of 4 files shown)


erskingardner:
@coderabbitai review

src/ffi.rs
```diff
@@ -0,0 +1,307 @@
...
+///   NUL-terminated UTF-8 C strings. Null pointers and invalid UTF-8 are
+///   handled gracefully by returning a failure JSON response.
+/// * The returned pointer must be freed with [`wn_string_free`].
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn wn_collect_notifications_after_push(
+    data_dir: *const c_char,
+    logs_dir: *const c_char,
+    keyring_service_id: *const c_char,
+    max_wait_ms: u32,
+) -> *mut c_char {
```
dannym-arx: oh jeez... this might just be an impossibility of the api, but can we avoid c strings? or if we can't can we at least length prefix them?

In my experience, passing C strings, especially across boundaries, has a consistent track record of eventually producing bugs that turn into serious security vulnerabilities. I'd feel much more comfortable if we could sidestep that risk entirely or at least reduce by explicitly passing the length.

dannym-arx: to be fair I can't see anything here in this function that could break right now, but someone is gonna add something to this at some point in time, and break it

src/ffi.rs
```diff
@@ -0,0 +1,307 @@
...
+///
+/// * `ptr` must have been returned by a `wn_*` function that allocates via `CString`.
+/// * Must not be called more than once for the same pointer.
+/// * Passing a null pointer is a safe no-op.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn wn_string_free(ptr: *mut c_char) {
+    if !ptr.is_null() {
+        drop(unsafe { CString::from_raw(ptr) });
+    }
+}
```
dannym-arx: you can't do this. You can't guarantee that something outside the boundary isn't accessing that memory.

this only works inside rust, the memory is being accessed from outside, from essentially a black box (swift itself isn't the blackbox, ios is, to clarify). You could call free, but that could also result in use after free (which the user would experience as a bug, or worse as an os-level data leak)... 

at best this is UB

dannym-arx (changes requested):

src/whitenoise/mod.rs
```diff
@@ -634,6 +635,26 @@ impl Whitenoise {
...
+    /// This is the primary entry point for contexts where the caller does not know whether
+    /// Whitenoise has already been started — for example, iOS background push handlers that
+    /// may fire when the app process is still alive or after a full cold launch.
+    ///
+    /// Unlike calling [`initialize_whitenoise`] directly, this method is safe to call
+    /// repeatedly — it will never start duplicate event processors or background tasks.
+    ///
+    /// [`initialize_whitenoise`]: Self::initialize_whitenoise
+    pub async fn ensure_initialized(config: WhitenoiseConfig) -> Result<&'static Self> {
+        if GLOBAL_WHITENOISE.get().is_none() {
```
mubarakcoded, line 688: This check-then-act here isn't safe under concurrent callers - two background handlers could both see is_none(), and both call initialize_whitenoise. If initialize_whitenoise guards itself internally, this is safe, but the doc says "never start duplicate event processors" so that guarantee needs to be maintained here, not just assumed about the internals. Either let initialize_whitenoise own the idempotency entirely and drop the early check, or make this an atomic compare-and-init.

src/ffi.rs
erskingardner: yeah this part definitely worried me. let me take another crack at it.

src/ffi.rs
erskingardner: given that this is a single call and response - not interactive and is run in a one time call when a notification arrives does that change things? I'm guessing not...

src/ffi.rs
```diff
@@ -0,0 +1,307 @@
...
+
+/// Convert an owned `String` of JSON into a leaked C string pointer.
+///
+/// Guaranteed to never return NULL: if the input contains interior NUL bytes
+/// or CString construction fails for any reason, the static fallback JSON is
+/// used instead.
+fn string_to_raw(json: String) -> *mut c_char {
+    // CString::new rejects interior NULs. First try the happy path.
+    if let Ok(cstr) = CString::new(json.clone()) {
```
mubarakcoded: The clone here is unnecessary - CString::new gives back the bytes on failure via NulError::into_vec(), so you can recover ownership without cloning upfront. It's a minor, but it saves allocation on every call.

mubarakcoded (changes requested):

dannym-arx (dismissed):

src/ffi.rs
```diff
@@ -0,0 +1,500 @@
...
+            Ok(rt) => rt,
+            Err(e) => {
+                return result_to_json(BackgroundNotificationResult::failed(format!(
+                    "failed to create tokio runtime: {e}"
+                )));
+            }
+        };
+
+        let result = rt.block_on(collect_notifications_after_push(config, max_wait));
+        result_to_json(result)
```
jgmontoya: This got flagged by my agent:

> 
> ```rust
> let rt = match tokio::runtime::Builder::new_current_thread()
>     .enable_all()
>     .build()
> …
> let result = rt.block_on(collect_notifications_after_push(config, max_wait));
> ```
> 
> When `collect_notifications_after_push` triggers a _cold_ init, `Whitenoise::initialize_whitenoise` in `src/whitenoise/mod.rs:584,595,606` performs multiple `tokio::spawn` calls for:
> 
> - the event-processing loop (`src/whitenoise/event_processor/mod.rs:26`),
> - `start_scheduled_tasks` (`src/whitenoise/scheduled_tasks/mod.rs:82`),
> - `discovery_sync_worker::run` (`src/whitenoise/mod.rs:606`).
> 
> Every one of those tasks is pinned to the runtime active when `tokio::spawn` is called — i.e., _this FFI call's_ current-thread runtime. When `wn_collect_notifications_after_push` returns, `rt` drops and all those tasks are aborted mid-execution. The singleton's `GLOBAL_WHITENOISE` cell is still populated, and on the next FFI invocation `ensure_initialized` returns early via the fast path (`src/whitenoise/mod.rs:682-684`), so none of those tasks get re-spawned. From that point on, the "live" singleton has no event processor: `ensure_all_subscriptions` feeds events into `event_sender`, nothing drains them, and the broadcast channel never emits a single `NotificationUpdate`. Drain waits out `DEFAULT_QUIET_WINDOW` and returns `no_data` silently.
> 
> This is operationally invisible on a "single push per process" iOS model, and in practice Apple does often terminate the extension between silent pushes — which is presumably why this passed eyeballing. But:
> 
> 1. The invariant "iOS always terminates between silent pushes" is undocumented and not enforced by the code; a future caller (including tests) that invokes the FFI twice in one process will see silent failure.
> 2. Unit tests like `null_ptr_with_zero_len_is_empty_string_not_failure` (`src/ffi.rs:456`) run `wn_collect_notifications_after_push` end-to-end and are already sharing `GLOBAL_WHITENOISE` across tests — which means the order tests run in can mask or cause failures.
> 
> **Suggested fix.** Park a process-lifetime runtime in a `OnceLock<tokio::runtime::Runtime>` owned by the `ffi` module (or a separate module) and `block_on` on _its_ `Handle`. A multi-threaded runtime is not required for this workload, but it must outlive the FFI call. Alternatively: document and enforce "process terminates after each call" (e.g., by refusing a second call).

src/whitenoise/background_notifications.rs
```diff
@@ -0,0 +1,459 @@
...
+                e
+            );
+            BackgroundNotificationResult::failed(e.to_string())
+        }
+    }
+}
+
+async fn collect_notifications_inner(
+    config: WhitenoiseConfig,
+    max_wait: Duration,
```
jgmontoya, line 179: The `max_wait` is only being enforced by `drain_notifications` but if `Whitenoise::ensure_initialized` and `whitenoise.subscribe_to_notifications` could take some time particularly the cold-initialization.

Perhaps we should wrap the `collect_notifications_inner` call on a `tokio::time::timeout(max_wait, …)` although tbh I'm not sure if there's much of a difference in practice given we're already leaving 5s margin for initialization

mubarakcoded (dismissed):

src/whitenoise/background_notifications.rs
erskingardner: yeah - this is taking that startup time into account already. I guess it isn't super semantically clear but "wait" time to me was how long I wanted it to wait once it was actually getting events. I'm going to leave this as is for now.

dannym-arx (approved):

mubarakcoded (approved):

jgmontoya (approved):

PR #753, opened 2026-04-15
Add scoped relay handles to AccountSession (Phase 3)

dannym-arx:
marmot

## What

Phase 3 of the session-projection rearchitecture (see `docs/session-projection-implementation-plan.md`). Adds scoped relay handles to `AccountSession` so each session carries identity-bound access to the shared ephemeral and group relay planes.

## Why

Today, every account-scoped relay operation passes `account_pubkey` and `signer` as loose arguments through `RelayControlPlane`. The compiler cannot enforce that the right identity is used for the right account. Scoped handles bake the identity in at construction time: callers use `session.ephemeral.publish_gift_wrap(...)` with no pubkey or signer argument, and the handle routes through shared connections internally.

This is the prerequisite for later phases that migrate callers off the global `relay_control` facade and onto session-owned operations.

## How

**New file: `src/whitenoise/session/relay_handles.rs`**

Two handle types, both constructed inside `AccountSession::new()`:

- `AccountEphemeralHandle` wraps `EphemeralPlane` + a shared signer slot (`Arc<RwLock<Option<Arc<dyn NostrSigner>>>>`). Provides signing helpers for gift wraps, metadata, relay lists, key packages, deletions, and follow lists. Provides fetch helpers and relay warming. The signer slot is shared with the session, so `register_external_signer()` updates propagate to the handle without reconstruction.

- `AccountGroupHandle` wraps `Arc<RelayControlPlane>` + account pubkey. Provides subscription sync, health checks, and group count queries.

**Migrated callers (proof of concept):**

- `subscriptions.rs`: `is_account_subscriptions_operational` now routes through the session's group handle instead of calling `relay_control` directly.
- `setup.rs`: account-scoped ephemeral relay warming routes through the session's ephemeral handle when a session exists.

**Other changes:**

- `SharedSigner` type alias in `session/mod.rs` for the `Arc<RwLock<Option<Arc<dyn NostrSigner>>>>` pattern.
- `WhitenoiseError::SignerUnavailable` variant for when a handle needs a signer but none is registered.
- `AccountSession::from_account` pulls `EphemeralPlane` and `Arc<RelayControlPlane>` from `Whitenoise` to build handles during construction.
- Test helpers extracted to `session::test_helpers` with async `test_session()` that builds minimal relay infrastructure.

## API consolidation on the handle

The handle API is tighter than the underlying `EphemeralPlane` surface it wraps:

| EphemeralPlane (two methods) | Handle (one method) |
|---|---|
| `publish_follow_list_with_signer` + `publish_follow_list_with_signer_at` | `publish_follow_list(..., created_at: Option<Timestamp>)` |
| `publish_event_deletion_with_signer` + `publish_batch_event_deletion_with_signer` | `publish_event_deletion(event_ids: &[EventId], ...)` |
| `publish_event_to` + `publish_event_with_quorum` | `publish_event(..., quorum_threshold: Option<usize>)` |

## Validation

- `just check-fmt` / `just check-clippy` / `just check-docs` / `just check-dead-code-allows`: all pass
- New tests (4 in `relay_handles`, 15 existing session tests updated): all pass
- Handle methods not yet called from production use `#[expect(dead_code)]` (passes the `allow(dead_code)` lint check and will warn when later phases activate them)

## Summary by CodeRabbit

* **Improvements**
  * Clearer error messages when attempting to use external signers that have not been registered
  * Refactored session and relay management architecture for improved system stability and reliability
  * Optimized relay connection warming and subscription health monitoring through enhanced internal implementation
  * Better organization of relay operations within account sessions for improved performance

### DIFF
--- a/src/whitenoise/accounts/setup.rs
+++ b/src/whitenoise/accounts/setup.rs
@@ -963,10 +963,19 @@ impl Whitenoise {
 
         // Warm both scopes concurrently — they use separate relay sessions
         // so there's no contention, and this halves the worst-case timeout.
+        let session = self.session(&account.pubkey);
         let (anon_result, account_result) = tokio::join!(
             self.relay_control.warm_ephemeral_relays(&warm_relays),
-            self.relay_control
-                .warm_ephemeral_relays_for_account(account.pubkey, &warm_relays),
+            async {
+                match &session {
+                    Some(s) => s.ephemeral.warm_relays(&warm_relays).await,
+                    None => {
+                        self.relay_control
+                            .warm_ephemeral_relays_for_account(account.pubkey, &warm_relays)
+                            .await
+                    }
+                }
+            },
         );
 
         if let Err(error) = anon_result {
--- a/src/whitenoise/session/relay_handles.rs
+++ b/src/whitenoise/session/relay_handles.rs
@@ -1,0 +1,359 @@
+use std::sync::Arc;
+
+use mdk_core::prelude::GroupId;
+use nostr_sdk::prelude::*;
+use nostr_sdk::{PublicKey, RelayUrl};
+
+use super::SharedSigner;
+use crate::RelayType;
+use crate::nostr_manager::{NostrManagerError, Result as NostrResult};
+use crate::relay_control::RelayControlPlane;
+use crate::relay_control::ephemeral::{EphemeralPlane, MessagePublishResult};
+use crate::relay_control::groups::GroupSubscriptionSpec;
+use crate::whitenoise::database::Database;
+use crate::whitenoise::error::WhitenoiseError;
+use crate::whitenoise::message_streaming::MessageStreamManager;
+
+/// Scoped handle into the shared ephemeral relay plane.
+///
+/// Carries this account's identity and signer reference. Delegates to shared
+/// warm relay connections internally. The API never exposes the pubkey or signer
+/// — they are baked in at construction time.
+pub(crate) struct AccountEphemeralHandle {
+    account_pubkey: PublicKey,
+    ephemeral: EphemeralPlane,
+    signer: SharedSigner,
+}
+
+#[expect(dead_code, reason = "signing/fetch helpers used by later phases")]
+impl AccountEphemeralHandle {
+    pub(crate) fn new(
+        account_pubkey: PublicKey,
+        ephemeral: EphemeralPlane,
+        signer: SharedSigner,
+    ) -> Self {
+        Self {
+            account_pubkey,
+            ephemeral,
+            signer,
+        }
+    }
+
+    /// Read the current signer, returning an error if none is set.
+    async fn require_signer(&self) -> NostrResult<Arc<dyn NostrSigner>> {
+        self.signer.read().await.clone().ok_or_else(|| {
+            NostrManagerError::WhitenoiseInstance(
+                WhitenoiseError::SignerUnavailable(self.account_pubkey).to_string(),
+            )
+        })
+    }
+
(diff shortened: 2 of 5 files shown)


dannym-arx:
@claude pretty sure this is fine, but do a quick review
@coderabbitai you too plz :rabbit:

erskingardner (approved):

mubarakcoded (approved):

src/whitenoise/session/relay_handles.rs
```diff
@@ -0,0 +1,357 @@
...
+
+    /// Read the current signer, returning an error if none is set.
+    async fn require_signer(&self) -> NostrResult<Arc<dyn NostrSigner>> {
+        self.signer.read().await.clone().ok_or_else(|| {
+            NostrManagerError::WhitenoiseInstance(
+                WhitenoiseError::SignerUnavailable(self.account_pubkey).to_string(),
+            )
+        })
+    }
+
```
jgmontoya, line 49: From my agent:

> Two problems compounded in four lines:
> 
> 1. **Type erasure.** We just added a typed `WhitenoiseError::SignerUnavailable(PublicKey)` variant in this same branch (`src/whitenoise/error.rs:86`) — and then immediately throw the type away by converting it to a `String` and wrapping it in `NostrManagerError::WhitenoiseInstance(String)`. Any future caller that wants to match on "signer missing, surface a reconnect prompt" vs. "relay error, retry" has to grep error messages. That is exactly the pattern a typed `thiserror` enum exists to prevent. Adding `SignerUnavailable(PublicKey)` only to stringify it defeats the change.
> 2. **Layering inversion.** `AccountEphemeralHandle` lives in `src/whitenoise/session/` — the whitenoise layer. `NostrManagerError` is the error type of a lower layer (`src/nostr_manager/`). The handle is choosing to surface a *whitenoise-level* failure (account has no signer) through a *nostr-manager-level* error variant. This is upside‑down.
> 
> Both problems have the same root cause: the public methods of `AccountEphemeralHandle` return `NostrResult<T>` to mirror `EphemeralPlane`. Once you commit to `NostrResult` at this layer, every non-plane failure mode has to be smuggled through `NostrManagerError`.
> 
> **Suggested fix.** Return `whitenoise::Result<T>` from the handle's public API and use `?` to lift the plane's `NostrResult` (there is already `WhitenoiseError::NostrManager(#[from] NostrManagerError)` at `src/whitenoise/error.rs:132`). Then `require_signer` can just return `Err(WhitenoiseError::SignerUnavailable(self.account_pubkey))` directly. Call sites also win: they already use `WhitenoiseError` end-to-end, and today they would all need `.map_err(WhitenoiseError::from)` to consume the handle.
> 
> If keeping `NostrResult` is somehow important (I don't see why — the handle is pure pass-through), then at minimum add a typed `NostrManagerError::SignerUnavailable(PublicKey)` variant so the pubkey isn't embedded in a string.

src/whitenoise/accounts/setup.rs
```diff
@@ -963,10 +963,19 @@ impl Whitenoise {
...
+                match &session {
+                    Some(s) => s.ephemeral.warm_relays(&warm_relays).await,
+                    None => {
+                        self.relay_control
+                            .warm_ephemeral_relays_for_account(account.pubkey, &warm_relays)
+                            .await
+                    }
+                }
+            },
         );
```
jgmontoya, line 979: When do we expect there to be no session / the session to be `None`?

dannym-arx: - During multi-step login
- During app startup, before restore_sessions() completes
- External signer not yet connected

PR #754, opened 2026-04-15
Add AccountRepositories scaffold (Phase 4)

mubarakcoded:
## What changed

Implements Phase 4 of the session/projection refactor plan.

**New module: `src/whitenoise/database/account/`**

Three per-account repositories that wrap existing DB functions with the account pubkey baked in at construction — callers never thread a pubkey argument through method calls:

- `DraftsRepo` — wraps `Draft::save`, `Draft::find`, `Draft::delete`
- `AccountSettingsRepo` — wraps `AccountSettings::find_or_create_for_pubkey`, `notifications_enabled_for_pubkey`, `update_notifications_enabled`
- `AccountFollowsRepo` — wraps `Account::follows`, `is_following_user`, `follow_user`, `unfollow_user`

`AccountRepositories` bundles all three and is wired onto `AccountSession` as `pub repos: AccountRepositories`.

## What didn't change

- No existing DB functions modified
- No Phase 5 view types — repos only
- No external API changes

## Tests

Focused unit tests for each repo covering save/find/delete, default values, round-trips, and cross-account isolation (two accounts sharing a group cannot see each other's drafts/follows).

## Marmot note

Every marmot now has their own filing cabinet. No more rummaging through the colony's shared pile.

## Release Notes

* **Refactor**
  * Enhanced internal architecture for managing account-scoped features including message drafts, follow relationships, and notification settings to improve code organization and maintainability.

### DIFF
--- a/src/whitenoise/database/account/drafts.rs
+++ b/src/whitenoise/database/account/drafts.rs
@@ -66,45 +66,12 @@ impl DraftsRepo {
 #[cfg(test)]
 mod tests {
     use mdk_core::prelude::GroupId;
-    use nostr_sdk::{Keys, PublicKey};
+    use nostr_sdk::Keys;
 
     use super::DraftsRepo;
-    use crate::whitenoise::database::Database;
-    use crate::whitenoise::test_utils::create_mock_whitenoise;
-
-    async fn insert_test_account(database: &Database, pubkey: &PublicKey) {
-        let user_pubkey = pubkey.to_hex();
-        sqlx::query("INSERT INTO users (pubkey, metadata) VALUES (?, '{}')")
-            .bind(&user_pubkey)
-            .execute(&database.pool)
-            .await
-            .expect("insert user");
-        let (user_id,): (i64,) = sqlx::query_as("SELECT id FROM users WHERE pubkey = ?")
-            .bind(&user_pubkey)
-            .fetch_one(&database.pool)
-            .await
-            .expect("get user id");
-        sqlx::query("INSERT INTO accounts (pubkey, user_id, last_synced_at) VALUES (?, ?, NULL)")
-            .bind(&user_pubkey)
-            .bind(user_id)
-            .execute(&database.pool)
-            .await
-            .expect("insert account");
-    }
-
-    async fn insert_test_group(database: &Database, group_id: &GroupId) {
-        let now = chrono::Utc::now().timestamp_millis();
-        sqlx::query(
-            "INSERT INTO group_information (mls_group_id, group_type, created_at, updated_at)
-             VALUES (?, 'group', ?, ?)",
-        )
-        .bind(group_id.as_slice())
-        .bind(now)
-        .bind(now)
-        .execute(&database.pool)
-        .await
-        .expect("insert group_information");
-    }
+    use crate::whitenoise::test_utils::{
+        create_mock_whitenoise, insert_test_account, insert_test_group,
+    };
 
     #[tokio::test]
     async fn save_creates_draft_scoped_to_account() {
--- a/src/whitenoise/database/account/drafts.rs
+++ b/src/whitenoise/database/account/drafts.rs
@@ -142,7 +142,9 @@ mod tests {
     async fn find_returns_none_when_no_draft_exists() {
         let (wn, _d, _l) = create_mock_whitenoise().await;
         let keys = Keys::generate();
+        insert_test_account(&wn.database, &keys.public_key()).await;
         let group_id = GroupId::from_slice(b"test-group-id-00");
+        insert_test_group(&wn.database, &group_id).await;
 
         let repo = DraftsRepo::new(keys.public_key(), wn.database.clone());
         let found = repo.find(&group_id).await.unwrap();
@@ -168,7 +170,9 @@ mod tests {
     async fn delete_nonexistent_is_noop() {
         let (wn, _d, _l) = create_mock_whitenoise().await;
         let keys = Keys::generate();
+        insert_test_account(&wn.database, &keys.public_key()).await;
         let group_id = GroupId::from_slice(b"test-group-id-00");
+        insert_test_group(&wn.database, &group_id).await;
 
         let repo = DraftsRepo::new(keys.public_key(), wn.database.clone());
         assert!(repo.delete(&group_id).await.is_ok());
--- a/src/whitenoise/database/account/drafts.rs
+++ b/src/whitenoise/database/account/drafts.rs
@@ -1,0 +1,197 @@
+//! Per-account repository for message drafts.
+//!
+//! Wraps the existing [`Draft`] DB functions so that callers do not need to
+//! thread an `account_pubkey` argument through every call — the pubkey is
+//! baked in at construction time.
+
+use std::sync::Arc;
+
+use mdk_core::prelude::GroupId;
+use nostr_sdk::{EventId, PublicKey};
+
+use crate::whitenoise::database::Database;
+use crate::whitenoise::drafts::Draft;
+use crate::whitenoise::error::Result;
+use crate::whitenoise::media_files::MediaFile;
+
+/// Repository for message drafts scoped to a single account.
+#[derive(Clone, Debug)]
+pub struct DraftsRepo {
+    account_pubkey: PublicKey,
+    db: Arc<Database>,
+}
+
+impl DraftsRepo {
+    /// Construct a new [`DraftsRepo`] for `account_pubkey`.
+    pub(crate) fn new(account_pubkey: PublicKey, db: Arc<Database>) -> Self {
+        Self { account_pubkey, db }
+    }
+
+    /// Upsert the draft for `group_id`.
+    ///
+    /// Delegates to `Draft::save` with the baked-in account pubkey.
+    pub async fn save(
+        &self,
+        mls_group_id: &GroupId,
+        content: &str,
+        reply_to_id: Option<&EventId>,
+        media_attachments: &[MediaFile],
+    ) -> Result<Draft> {
+        Draft::save(
+            &self.account_pubkey,
+            mls_group_id,
+            content,
+            reply_to_id,
+            media_attachments,
+            &self.db,
+        )
+        .await
+    }
+
+    /// Return the draft for `group_id`, or `None` if absent.
+    ///
+    /// Delegates to `Draft::find` with the baked-in account pubkey.
+    pub async fn find(&self, mls_group_id: &GroupId) -> Result<Option<Draft>> {
+        Draft::find(&self.account_pubkey, mls_group_id, &self.db).await
+    }
+
+    /// Delete the draft for `group_id`. No-op if no draft exists.
+    ///
+    /// Delegates to `Draft::delete` with the baked-in account pubkey.
+    pub async fn delete(&self, mls_group_id: &GroupId) -> Result<()> {
+        Draft::delete(&self.account_pubkey, mls_group_id, &self.db).await
+    }
+}
+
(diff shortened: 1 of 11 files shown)


mubarakcoded:
@claude do a full review please, @coderabbitai you too

src/whitenoise/database/account/drafts.rs
```diff
@@ -0,0 +1,197 @@
...
+
+    #[tokio::test]
+    async fn find_returns_none_when_no_draft_exists() {
+        let (wn, _d, _l) = create_mock_whitenoise().await;
+        let keys = Keys::generate();
+        let group_id = GroupId::from_slice(b"test-group-id-00");
+
+        let repo = DraftsRepo::new(keys.public_key(), wn.database.clone());
+        let found = repo.find(&group_id).await.unwrap();
+        assert!(found.is_none());
```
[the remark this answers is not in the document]
mubarakcoded: fixed

src/whitenoise/database/account/drafts.rs
```diff
@@ -0,0 +1,201 @@
...
+pub struct DraftsRepo {
+    account_pubkey: PublicKey,
+    db: Arc<Database>,
+}
+
+impl DraftsRepo {
+    /// Construct a new [`DraftsRepo`] for `account_pubkey`.
+    pub(crate) fn new(account_pubkey: PublicKey, db: Arc<Database>) -> Self {
+        Self { account_pubkey, db }
+    }
```
dannym-arx, line 28: not sure if this is something we should do, open question for you, @mubarakcoded as well as @erskingardner and @jgmontoya 

what do you think about having a DraftsRepo, which only has one method, with_group, or for_group, (or perhaps even having a group id here in the new), and then having the group always implied in all the functions?

dannym-arx (approved):
this is great!

the API is gonna be so much cleaner with this refactor I can see it forming slowly with every phase 😮

just one minor question

src/whitenoise/database/account/drafts.rs
mubarakcoded: Good call. I'd hold off for now because settings and follows don't have a sub-scope, and here the group is just a lookup key, not a context you stay in between calls. But it'd really fit is Phase 5 where DraftOps is already planned as a view on the session, and that's exactly where a for_group(&group_id) accessor would feel natural. Worth revisiting when we get there.

dannym-arx: makes sense to me!

src/whitenoise/database/account/follows.rs
```diff
@@ -0,0 +1,245 @@
...
+use crate::whitenoise::error::{Result, WhitenoiseError};
+use crate::whitenoise::users::User;
+
+/// Repository for follow relationships scoped to a single account.
+#[derive(Clone, Debug)]
+pub struct AccountFollowsRepo {
+    account_pubkey: PublicKey,
+    db: Arc<Database>,
+    /// Cached [`Account`] row. Populated on first use; shared across clones.
+    account_cache: Arc<OnceLock<Account>>,
```
erskingardner: one thing that my agent flagged. I _think_ this is by design in that these repos are not long-lived objects that hang around. But I wanted to flag that and make sure we're all thinking the same way about? All the *Repo objects are stood up to perform some action then they are dropped? 

> `Account` is cached on first load and never refreshed. The underlying `accounts` row likely has mutable fields (`updated_at`, `last_synced_at`, etc.) that can change during the lifetime of the repo. Any call that follows a mutation in a long-lived `AccountSession` will silently return the original snapshot. Even if `Account` itself is mostly immutable today, the cache imposes a hidden constraint on future changes to the `accounts` table. The comment claims the cache avoids repeated DB round-trips, but `Account::find_by_pubkey` is an indexed single-row lookup — the cost is negligible. The caching adds complexity and a subtle correctness risk for marginal gain.
> 
> Suggested fix: remove the cache and the `Arc<OnceLock<Account>>` field. Load the account inline at each call site, or extract the integer `account_id` once on construction (since that FK is immutable).

src/whitenoise/database/account/settings.rs
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    /// Return the settings for this account, creating a default row if none
+    /// exists.
+    ///
+    /// Delegates to `AccountSettings::find_or_create_for_pubkey`.
+    pub async fn find_or_create(&self) -> Result<AccountSettings> {
+        AccountSettings::find_or_create_for_pubkey(&self.account_pubkey, &self.db)
+            .await
+            .map_err(|e| WhitenoiseError::Database(DatabaseError::Sqlx(e)))
```
erskingardner: I don't think you need these error mappings here. Shouldn't a `?` on the end just map them up through the chain to `WhitenoiseError`?

src/whitenoise/database/account/settings.rs
```diff
@@ -0,0 +1,160 @@
...
+
+    /// Return whether notifications are enabled for this account.
+    ///
+    /// Returns `true` (the default) when no settings row exists.
+    ///
+    /// Delegates to `AccountSettings::notifications_enabled_for_pubkey`.
+    pub async fn notifications_enabled(&self) -> Result<bool> {
+        AccountSettings::notifications_enabled_for_pubkey(&self.account_pubkey, &self.db)
+            .await
+            .map_err(|e| WhitenoiseError::Database(DatabaseError::Sqlx(e)))
```
erskingardner: same here.

src/whitenoise/database/account/settings.rs
```diff
@@ -0,0 +1,160 @@
...
+    /// **Note:** this only updates the database row. It does **not** perform
+    /// push-token reconciliation (sharing or removing the local push token
+    /// from joined groups). Callers that need the full side-effectful behaviour
+    /// should use `Whitenoise::update_notifications_enabled` instead, or handle
+    /// push-token sync separately. Phase 5 callers must account for this when
+    /// migrating away from the `Whitenoise` facade.
+    pub async fn update_notifications_enabled(&self, enabled: bool) -> Result<AccountSettings> {
+        AccountSettings::update_notifications_enabled(&self.account_pubkey, enabled, &self.db)
+            .await
+            .map_err(|e| WhitenoiseError::Database(DatabaseError::Sqlx(e)))
```
erskingardner: same here?

src/whitenoise/database/account/drafts.rs
```diff
@@ -0,0 +1,201 @@
...
+#[cfg(test)]
+mod tests {
+    use mdk_core::prelude::GroupId;
+    use nostr_sdk::{Keys, PublicKey};
+
+    use super::DraftsRepo;
+    use crate::whitenoise::database::Database;
+    use crate::whitenoise::test_utils::create_mock_whitenoise;
+
+    async fn insert_test_account(database: &Database, pubkey: &PublicKey) {
```
erskingardner: This is duplicated in a bunch of places. we should centralize it

src/whitenoise/database/account/drafts.rs
```diff
@@ -0,0 +1,201 @@
...
+            .expect("get user id");
+        sqlx::query("INSERT INTO accounts (pubkey, user_id, last_synced_at) VALUES (?, ?, NULL)")
+            .bind(&user_pubkey)
+            .bind(user_id)
+            .execute(&database.pool)
+            .await
+            .expect("insert account");
+    }
+
+    async fn insert_test_group(database: &Database, group_id: &GroupId) {
```
erskingardner: This one too.

erskingardner (changes requested):
Some minor things. mostly LGTM

src/whitenoise/mod.rs
```diff
@@ -779,6 +779,56 @@ pub mod test_utils {
     use nostr_sdk::{EventBuilder, EventId, Keys, PublicKey, RelayUrl, UnsignedEvent};
     use tempfile::TempDir;
 
+    // ── Shared raw-SQL fixture helpers ────────────────────────────────────────
+
+    /// Insert a minimal user + account row for `pubkey`.
+    ///
+    /// Used by repository unit tests that need a real account row in the DB
+    /// without going through the full account-creation flow.
+    pub async fn insert_test_account(
```
erskingardner, line 788: `database/drafts.rs` and `database/account_settings.rs` also have usages of their own version of this method - you might as well remove the duplication from there as well. 

`database/drafts.rs` also has a copy of `insert_test_group` let's remove that too.

src/whitenoise/database/account/follows.rs
```diff
@@ -0,0 +1,271 @@
...
+            .fetch_optional(&db.pool)
+            .await
+            .map_err(DatabaseError::Sqlx)?;
+        row.map(|(id,)| id).ok_or(WhitenoiseError::AccountNotFound)
+    }
+
+    /// Return all users followed by this account.
+    pub async fn all(&self) -> Result<Vec<User>> {
+        let users = sqlx::query_as::<_, User>(
+            "SELECT u.id, u.pubkey, u.metadata, u.created_at, u.metadata_known_at, u.updated_at
```
erskingardner, line 54: We're duplicating the sql that is already in the accounts database module? 

Is this sort of an early test of moving that SQL out to these repository files?

src/whitenoise/session/mod.rs
```diff
@@ -453,21 +461,44 @@ pub(crate) mod test_helpers {
...
+            .bind(&user_pubkey)
+            .execute(&db.pool)
+            .await
+            .expect("insert test user");
+        let (user_id,): (i64,) = sqlx::query_as("SELECT id FROM users WHERE pubkey = ?")
+            .bind(&user_pubkey)
+            .fetch_one(&db.pool)
+            .await
+            .expect("get test user id");
+        sqlx::query("INSERT INTO accounts (pubkey, user_id, last_synced_at) VALUES (?, ?, NULL)")
```
erskingardner, line 477: Couldn't this also use `insert_test_account`?

src/whitenoise/database/account/follows.rs
```diff
@@ -0,0 +1,271 @@
...
+    /// Returns `false` (rather than an error) when `target_pubkey` has no user
+    /// record in the database, matching the existing behaviour in
+    /// `Whitenoise::is_following_user`.
+    pub async fn is_following(&self, target_pubkey: &PublicKey) -> Result<bool> {
+        let user = match User::find_by_pubkey(target_pubkey, &self.db).await {
+            Ok(user) => user,
+            Err(WhitenoiseError::UserNotFound) => return Ok(false),
+            Err(e) => return Err(e),
+        };
+        let result = sqlx::query(
```
erskingardner, line 77: You're doing two roundtrips to the DB here. First fetches the full `User` row to get `user.id`, then issues a `COUNT(*)` query. Since only `user.id` is needed, a single `EXISTS (SELECT 1 FROM users u JOIN account_follows af ON af.user_id = u.id WHERE u.pubkey = ? AND af.account_id = ?)` would avoid the extra round-trip and the `User` allocation.

erskingardner (changes requested):

src/whitenoise/mod.rs
mubarakcoded: It has been addressed in d4f91e7

src/whitenoise/database/account/follows.rs
mubarakcoded: Yes, intentional - it's the start of migrating SQL ownership to the repo layer. Both coexist during the migration, Account::follows will be removed once all callers have moved here. I have added a doc comment to all() to make the intent explicit

erskingardner (approved):

PR #755, opened 2026-04-15
Move inbox plane ownership into AccountSession (Phase 3b)

dannym-arx:
marmot

## What

This marmot carried its mailbox home. AccountInboxPlane ownership moves from RelayControlPlane (a shared HashMap) into AccountSession, where it belongs.

Ephemeral and group planes are shared infrastructure accessed through scoped handles. The inbox plane is different: each account gets its own dedicated relay session, its own telemetry persistor, its own lifecycle. Storing it in a global HashMap keyed by pubkey was a holdover from before sessions existed.

## Why

The session/projection rearchitecture (Phase 3b) calls for each piece of state to have a declared owner. The inbox plane's owner is the account session, period. Parking it in RelayControlPlane forced every activate/deactivate call to route through the control plane and look up the right entry. Now the session creates, activates, and tears down its own inbox directly.

## How

**AccountSession gains inbox ownership:**
- New `AccountInboxState` struct wraps the plane + telemetry task handle
- `inbox: RwLock<Option<AccountInboxState>>` field, `None` until first activation
- `activation_lock: Mutex<()>` serializes concurrent activate/deactivate calls
- `activate_subscriptions()` orchestrates group update, inbox creation, telemetry spawn, and rollback on failure
- `deactivate_subscriptions()` tears down inbox + group + ephemeral scopes
- `has_inbox_subscription()` and `inbox_snapshot()` for health checks and diagnostics

**RelayControlPlane becomes a factory and shared-infra host:**
- `AccountInboxEntry`, `account_inbox_planes` HashMap, and `account_inbox_activation_locks` removed
- `activate_account_subscriptions()` and `deactivate_account_subscriptions()` removed
- New factory methods: `create_account_inbox_plane()`, `spawn_account_inbox_telemetry()`
- New delegation helpers: `group_plane_account_state()`, `remove_account_from_group_plane()`, `has_group_subscription()`, `remove_account_ephemeral_scope()`
- `snapshot()` accepts pre-collected inbox snapshots from the caller

**AccountGroupHandle extended for rollback support:**
- `save_state()`, `remove()`, `remove_ephemeral_scope()`
- `has_active_subscription()` checks only group health (inbox health checked via session)

**Callers updated:**
- `setup_account_subscriptions` and `refresh_account_subscriptions` route through the session
- `logout` deactivates subscriptions via the session before removing it
- `is_account_subscriptions_operational` checks both `session.has_inbox_subscription()` and `session.group_handle.has_active_subscription()`
- `get_relay_control_state` collects inbox snapshots from `AccountManager`
- `delete_all_data` deactivates session inboxes before shared-infra shutdown
- All tests updated to use session-based activation

## Validation

- `just check` (fmt, clippy, docs, dead_code): all pass
- 68 refactor-specific tests: all pass
- Full unit test suite: passes (known flaky key-package-timeout tests excluded)

### DIFF
--- a/src/whitenoise/session/mod.rs
+++ b/src/whitenoise/session/mod.rs
@@ -5,10 +5,16 @@ use std::sync::Arc;
 use dashmap::DashMap;
 use mdk_core::prelude::MDK;
 use mdk_sqlite_storage::MdkSqliteStorage;
-use nostr_sdk::PublicKey;
 use nostr_sdk::prelude::NostrSigner;
-use tokio::sync::{OwnedSemaphorePermit, RwLock, Semaphore, watch};
+use nostr_sdk::{PublicKey, RelayUrl};
+use tokio::sync::{Mutex, OwnedSemaphorePermit, RwLock, Semaphore, watch};
+use tokio::task::JoinHandle;
 
+use crate::nostr_manager::Result as NostrResult;
+use crate::relay_control::RelayControlPlane;
+use crate::relay_control::account_inbox::AccountInboxPlane;
+use crate::relay_control::groups::GroupSubscriptionSpec;
+use crate::types::AccountInboxPlaneStateSnapshot;
 use crate::whitenoise::Whitenoise;
 use crate::whitenoise::accounts::{Account, DiscoveredRelayLists};
 use crate::whitenoise::error::{Result, WhitenoiseError};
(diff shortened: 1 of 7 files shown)


src/whitenoise/session/mod.rs
```diff
@@ -117,6 +146,106 @@ impl AccountSession {
...
+        if let Some(state) = state {
+            state.deactivate().await;
+        }
+
+        self.group_handle.remove().await;
+        self.group_handle.remove_ephemeral_scope().await;
+    }
+
+    /// Deactivate only the inbox plane (used during shutdown).
+    pub(crate) async fn deactivate_inbox(&self) {
```
erskingardner, line 227: Does this method need to take the lock as well?

erskingardner (approved):
Minor nit - but otherwise looks good.

src/whitenoise/session/mod.rs
dannym-arx: we'll figure it out if it does once we start using it in the later phases, but maybe!

PR #757, opened 2026-04-16
Add dual-kind SelfRemove key package lifecycle support

erskingardner:
## Summary
- Publish MLS key packages as both canonical `30443` and legacy `443` events from the same generated package.
- Fetch both key package kinds, prefer valid `30443`, and fall back to compatible `443` only when needed.
- Require SelfRemove proposal support before group creation or member add, with a distinct update-needed error for incompatible packages.
- Track published key package `kind` and `d_tag`, and treat dual-published rows with the same `hash_ref` as one lifecycle unit.
- Remove the legacy missing-encoding cleanup path and deduplicate consumed key material deletion by `hash_ref`.

## Testing
- `just check-clippy`
- `just precommit-quick`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Publish and track paired key-package variants (canonical + legacy) with a lookup that prefers the newest valid canonical package and reports incompatibility or absence.

* **Bug Fixes**
  * Deduplicate cleanup and deletion by key-package group so twin records are handled together and redundant work is avoided.
  * Legacy-only deletion now preserves canonical packages.

* **Improvements**
  * Stricter key-package validation and clearer incompatibility/missing-metadata statuses.
  * Database migration adds columns and indexes to speed key-package lookups and maintenance.

### DIFF
--- a/src/integration_tests/test_cases/scheduler/key_package_maintenance.rs
+++ b/src/integration_tests/test_cases/scheduler/key_package_maintenance.rs
@@ -65,9 +65,9 @@ impl TestCase for KeyPackageMaintenanceTestCase {
         );
 
         // Delete any existing key packages to start with a clean slate. This
-        // test needs a fully empty relay state, while the app-facing "delete
-        // all" helper intentionally removes only legacy kind:443 copies.
-        let deleted = delete_all_relay_key_packages(context, &account, true).await?;
+        // test needs a fully empty relay state, while the developer-facing
+        // legacy cleanup intentionally removes only kind:443 copies.
+        let deleted = delete_all_relay_key_packages_for_test_setup(context, &account, true).await?;
         tracing::info!("✓ Deleted {} existing key package(s)", deleted);
 
         let before_delete = wait_for_key_packages(
@@ -101,7 +101,7 @@ impl TestCase for KeyPackageMaintenanceTestCase {
         );
 
         // Delete current key packages and publish an expired one.
-        delete_all_relay_key_packages(context, &account, true).await?;
+        delete_all_relay_key_packages_for_test_setup(context, &account, true).await?;
         wait_for_key_packages(
             context,
             &account,
@@ -209,7 +209,7 @@ where
     .await
 }
 
-async fn delete_all_relay_key_packages(
+async fn delete_all_relay_key_packages_for_test_setup(
     context: &ScenarioContext,
     account: &crate::Account,
     delete_mls_stored_keys: bool,
--- a/src/integration_tests/test_cases/scheduler/key_package_maintenance.rs
+++ b/src/integration_tests/test_cases/scheduler/key_package_maintenance.rs
@@ -8,6 +8,8 @@ use crate::whitenoise::scheduled_tasks::{KeyPackageMaintenance, Task};
 use async_trait::async_trait;
 use nostr_sdk::prelude::*;
 
+const MAX_RELAY_KEY_PACKAGE_DELETE_ROUNDS: u32 = 10;
+
 /// Verifies the key package maintenance task handles both cases:
 /// 1. Publishes key packages when none exist
 /// 2. Rotates expired key packages (deletes old, publishes new)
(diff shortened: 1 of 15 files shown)


src/whitenoise/key_packages.rs
```diff
@@ -394,6 +377,67 @@ impl Whitenoise {
...
+                    target: "whitenoise::key_packages",
+                    "Published canonical kind:30443 key package for account {} but failed \
+                     to publish legacy kind:443 twin: {}",
+                    account.pubkey.to_hex(),
+                    e,
+                );
+            }
+        }
+
+        Ok(())
```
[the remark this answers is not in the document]
erskingardner: this is ok - we're only supporting legacy key packages for another 2 weeks so this is an acceptable trade off for a low probability failure in publishing

src/whitenoise/key_packages.rs
```diff
@@ -520,13 +568,16 @@ impl Whitenoise {
...
                     let mdk = self.create_mdk_for_account(account.pubkey)?;
                     mdk.delete_key_package_from_storage_by_hash_ref(&pkg.key_package_hash_ref)?;
-                    if let Err(e) =
-                        PublishedKeyPackage::mark_key_material_deleted(pkg.id, &self.database).await
+                    if let Err(e) = PublishedKeyPackage::mark_key_material_deleted_by_hash_ref(
+                        &account.pubkey,
+                        &pkg.key_package_hash_ref,
+                        &self.database,
+                    )
+                    .await
```
jgmontoya, line 592: This removes the keypackage twins on the local state, but later (line 619) on only the delete for the handed event_id is sent to the relays. We should make sure that we're sending the delete for the other as well.

src/whitenoise/key_packages.rs
```diff
@@ -672,14 +725,14 @@ impl Whitenoise {
     pub async fn delete_all_key_packages_for_account(
         &self,
         account: &Account,
-        delete_mls_stored_keys: bool,
+        _delete_mls_stored_keys: bool,
     ) -> Result<usize> {
```
jgmontoya: We should probably rename this method from `delete_all_...` to `delete_legacy_...` to avoid confusion.

Also, we should drop the `_delete_mls_stored_keys` if we're not using it

src/whitenoise/groups.rs
```diff
@@ -106,12 +107,29 @@ impl Whitenoise {
...
-            mdk_core::Error::KeyPackage("Does not exist".to_owned()),
-        ))?;
+        let event = match lookup {
+            KeyPackageLookup::Found(event) => event,
+            KeyPackageLookup::Incompatible { error } => {
+                if matches!(error, WhitenoiseError::MissingMlsProposals { .. }) {
+                    return Err(WhitenoiseError::KeyPackageMissingSelfRemove {
+                        member_pubkey: *pk,
+                    });
+                }
```
jgmontoya, line 128: This is correct today, because the only "additional" proposal is `SelfRemove`, however for future-proofing we should inspect and see what exactly we're missing before returning the specific `WhitenoiseError::KeyPackageMissingSelfRemove` error

src/whitenoise/database/published_key_packages.rs
jgmontoya, line 1: I think the `mark_key_material_deleted` method is now not used anymore, we should delete it.

src/whitenoise/key_packages.rs
```diff
@@ -394,6 +375,67 @@ impl Whitenoise {
...
+            }
+            Err(e) => {
+                tracing::warn!(
+                    target: "whitenoise::key_packages",
+                    "Published canonical kind:30443 key package for account {} but failed \
+                     to publish legacy kind:443 twin: {}",
+                    account.pubkey.to_hex(),
+                    e,
+                );
+            }
```
jgmontoya, line 436: Should we perhaps retry publishing the legacy keypackage?

src/integration_tests/test_cases/scheduler/key_package_maintenance.rs
```diff
@@ -210,6 +209,50 @@ where
...
+                "Deleted 0 key package(s) despite {} remaining after {} relay cleanup round(s)",
+                key_package_count,
+                round + 1,
+            );
+            break;
+        }
+    }
+
+    Ok(total_deleted)
+}
```
jgmontoya: I think that we needing this as an integration test helper is kind of a red flag

jgmontoya (commented):
In many places we're still using the constant `Kind::MlsKeyPackage` which now in some places is `MLS_KEY_PACKAGE_KIND_LEGACY`. I think this can be a source of confusion.

dannym-arx (approved):

src/integration_tests/test_cases/scheduler/key_package_maintenance.rs
erskingardner: updated it a bit but I'd like to understand more why - we just need to start with a clean slate?

src/whitenoise/key_packages.rs
erskingardner: we're only keeping them around another 2 weeks. going to leave this.

mubarakcoded (approved):
LGTM

PR #758, opened 2026-04-16
Add SQLCipher encryption for Whitenoise database

erskingardner:
## Summary
- Open the Whitenoise SQLite database through SQLCipher during app initialization.
- Add plaintext database detection and one-time migration to an encrypted database using `sqlcipher_export`.
- Preserve user data during migration with WAL checkpointing, temp encrypted DB validation, backup/restore handling, and cleanup.
- Add tests for encrypted DB creation, reopen, plaintext migration, and missing key handling.
- Scope benchmark DB key restoration to a benchmark-only Whitenoise key id.

## Testing
- `just precommit-quick`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * End-to-end on-disk DB encryption support: encrypted DB init, guarded plaintext→encrypted migration, and per-run database key selection (including test/benchmark variants).
  * Config builder option to select a specific database key id for a run.

* **Bug Fixes / Behaviour**
  * More robust migration/recovery with sidecar validation, stale-lock detection, atomic swap/rollback, and multi-key sidecar restoration.
  * Subscription setup now awaits before key-package setup to avoid races.

* **Tests**
  * Added tests for encrypted DB creation, migration, reopen persistence, and missing-key failure scenarios.

### DIFF
--- a/src/whitenoise/database/encryption.rs
+++ b/src/whitenoise/database/encryption.rs
@@ -127,7 +127,17 @@ async fn migrate_plaintext_database(
         return Err(err.into());
     }
 
-    validate_encrypted_database(db_path, config).await?;
+    if let Err(err) = validate_encrypted_database(db_path, config).await {
+        fs::rename(&backup_path, db_path).map_err(|rollback_err| {
+            DatabaseError::EncryptionMigration(format!(
+                "Encrypted database validation failed ({err}); rollback from {} to {} failed: {rollback_err}",
+                backup_path.display(),
+                db_path.display(),
+            ))
+        })?;
+        return Err(err);
+    }
+
     remove_file_if_exists(&backup_path)?;
 
     Ok(())
--- a/src/whitenoise/database/encryption.rs
+++ b/src/whitenoise/database/encryption.rs
@@ -90,7 +90,7 @@ async fn assert_sqlcipher_available() -> Result<(), DatabaseError> {
 
     if version.is_empty() {
         return Err(DatabaseError::EncryptionMigration(
-            "SQLite was built without SQLCipher support".to_string(),
+            "SQLCipher is not available (cipher_version is empty)".to_string(),
         ));
     }
 
@@ -128,6 +128,14 @@ async fn migrate_plaintext_database(
     }
 
     if let Err(err) = validate_encrypted_database(db_path, config).await {
+        if let Err(rollback_err) = remove_file_if_exists(db_path) {
+            return Err(DatabaseError::EncryptionMigration(format!(
+                "Encrypted database validation failed ({err}); rollback from {} to {} failed while removing bad encrypted database: {rollback_err}",
+                backup_path.display(),
+                db_path.display(),
+            )));
+        }
+
         fs::rename(&backup_path, db_path).map_err(|rollback_err| {
             DatabaseError::EncryptionMigration(format!(
                 "Encrypted database validation failed ({err}); rollback from {} to {} failed: {rollback_err}",
@@ -147,16 +155,13 @@ async fn recover_interrupted_migration(
     db_path: &Path,
     config: &EncryptionConfig,
 ) -> Result<(), DatabaseError> {
-    if db_path.exists() {
-        let temp_path = sidecar_path(db_path, ".encrypted.tmp");
-        if database_file_state(db_path)? == DatabaseFileState::Plaintext {
-            remove_file_if_exists(&temp_path)?;
-        }
-        return Ok(());
-    }
-
     let temp_path = sidecar_path(db_path, ".encrypted.tmp");
     let backup_path = sidecar_path(db_path, ".plaintext.backup");
+
+    if db_path.exists() {
+        remove_file_if_exists(&temp_path)?;
+        return Ok(());
+    }
 
     if temp_path.exists() {
         match validate_encrypted_database(&temp_path, config).await {
--- a/src/whitenoise/database/encryption.rs
+++ b/src/whitenoise/database/encryption.rs
@@ -1,0 +1,395 @@
+use std::{
+    fs::{self, File, OpenOptions},
+    io::{ErrorKind, Read},
+    path::{Path, PathBuf},
+    time::{Duration, SystemTime},
+};
+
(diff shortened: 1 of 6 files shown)


src/whitenoise/database/mod.rs
```diff
@@ -84,40 +93,55 @@ pub struct Database {
...
+        db_path: PathBuf,
+        keyring_service_id: &str,
+        key_id: &str,
+    ) -> Result<Self, DatabaseError> {
         if let Some(parent) = db_path.parent() {
             std::fs::create_dir_all(parent)?;
         }
 
-        let db_url = format!("sqlite://{}", db_path.display());
+        let encryption_config =
```
jgmontoya, line 125: I think this `std::fs::create_dir_all(parent)?;` is redundant because the `open` method does it, no?

src/whitenoise/database/encryption.rs
```diff
@@ -0,0 +1,421 @@
...
+        let temp_path = sidecar_path(db_path, ".encrypted.tmp");
+        if database_file_state(db_path)? == DatabaseFileState::Plaintext {
+            remove_file_if_exists(&temp_path)?;
+        }
+        return Ok(());
+    }
+
+    let temp_path = sidecar_path(db_path, ".encrypted.tmp");
+    let backup_path = sidecar_path(db_path, ".plaintext.backup");
+
```
jgmontoya: I think we could do a more general check:

```rs
let temp_path = sidecar_path(db_path, ".encrypted.tmp");
let backup_path = sidecar_path(db_path, ".plaintext.backup");

if db_path.exists() {
    remove_file_if_exists(&temp_path)?;
    return Ok(());
}
```

Basically if the main db is intact we treat any surviving `.encrypted.tmp` as a leftover from a crashed migration

src/whitenoise/database/mod.rs
```diff
@@ -413,9 +460,34 @@ where
...
+
+    const SQLITE_HEADER: &[u8; 16] = b"SQLite format 3\0";
+
+    fn unique_service_id() -> String {
+        static COUNTER: AtomicU64 = AtomicU64::new(0);
+        let id = COUNTER.fetch_add(1, Ordering::SeqCst);
+        format!("com.whitenoise.database-test.{id}")
+    }
+
+    fn read_db_header(db_path: &PathBuf) -> [u8; 16] {
```
jgmontoya: From my agent:

> Per the project style, test helpers should prefer `&Path` over `&PathBuf`. All callers already pass `&db_path` where `db_path: PathBuf`, so `&Path` would work directly and matches the rest of the codebase.

src/whitenoise/database/encryption.rs
```diff
@@ -0,0 +1,421 @@
...
+async fn assert_sqlcipher_available() -> Result<(), DatabaseError> {
+    let mut conn = SqliteConnectOptions::new()
+        .in_memory(true)
+        .connect()
+        .await?;
+    let version = sqlcipher_version(&mut conn).await?;
+
+    if version.is_empty() {
+        return Err(DatabaseError::EncryptionMigration(
+            "SQLite was built without SQLCipher support".to_string(),
```
jgmontoya: From my agent:

> An empty `cipher_version` can also mean the cipher pragma is not recognised at the connection level (not just at build time). The message "SQLite was built without SQLCipher support" is technically accurate for most cases, but "SQLCipher is not available (cipher_version is empty)" would be less surprising if someone encounters this with a partially-linked build.

src/whitenoise/database/mod.rs
```diff
@@ -174,9 +200,30 @@ impl Database {
             })
             .connect_with(connect_options)
             .await?;
+
+        if encryption_config.is_some() {
+            // Confirm the pool can acquire a working connection end-to-end. Key correctness and
+            // schema integrity are already verified in prepare_sqlcipher_database before we reach here.
+            let mut conn = pool.acquire().await?;
+            sqlx::query("SELECT 1").fetch_one(&mut *conn).await?;
+        }
```
jgmontoya: From my agent:

> `validate_encrypted_database` (called from `prepare_sqlcipher_database` before the pool is opened) already opens a separate connection, runs `cipher_version`, queries `sqlite_master`, and runs `integrity_check`. The smoke-test `SELECT 1` here adds little safety net beyond those checks. The comment acknowledges this ("Key correctness and schema integrity are already verified"). Consider removing this block, since it costs a round-trip on every startup and the comment essentially admits it does not do anything novel.

src/whitenoise/database/mod.rs
jgmontoya, line 1: Last one from my agent:

> The test suite covers: new encrypted DB, reopen, plaintext migration, and missing key. It does not cover the `recover_interrupted_migration` path where `db_path` is absent and only `.plaintext.backup` exists (the code at `encryption.rs:183–185`). This is the most dangerous failure mode for a real user (power loss after rename but before the final rename). Adding a test that creates the backup file manually and calls `Database::new_encrypted` would provide confidence in this critical recovery path.

jgmontoya (commented):
Some comments!

jgmontoya (dismissed):

src/whitenoise/database/encryption.rs
```diff
@@ -0,0 +1,426 @@
...
+    remove_file_if_exists(&backup_path)?;
+    fs::rename(db_path, &backup_path)?;
+
+    if let Err(err) = fs::rename(&temp_path, db_path) {
+        let _ = fs::rename(&backup_path, db_path);
+        return Err(err.into());
+    }
+
+    if let Err(err) = validate_encrypted_database(db_path, config).await {
+        if let Err(rollback_err) = remove_file_if_exists(db_path) {
```
mubarakcoded, line 131: On rollback here. we deletes db_path first, then renames the backup into place. If the removal succeeds but the rename fails, both files are gone. On Unix, rename() atomically replaces the destination, I think we can skip the remove and just rename the backup directly over db_path right? It seems more safer and simpler.

mubarakcoded (approved):

dannym-arx (approved):
LGTM

PR #762, opened 2026-04-18
fix: video support

pepina-dev:
Sending videos across devices wasn't working in this flutter PR: https://github.com/marmot-protocol/whitenoise/pull/574. It only worked if both accounts were in same device (where media attachment was already saved on db). After making this vibecoded change, it works. 

The explanation is that for videos dimensions and blurhash is not processed so then the file metadata was None, but the file name was actually needed, so that condition was problematic. The quick fix is to make that optional. Next step would be to process video dimensions, blurhash  in mdk.... if that's the correct place to do that

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensure original filenames are consistently saved with uploaded media so filename metadata always persists and round-trips through storage.

* **Tests**
  * Added an end-to-end test that uploads a minimal MP4 and verifies MIME type, media type, persisted original filename, and database round-trip.

### DIFF
--- a/src/whitenoise/groups.rs
+++ b/src/whitenoise/groups.rs
@@ -865,6 +865,7 @@ impl Whitenoise {
 mod tests {
     use super::*;
     use crate::whitenoise::Whitenoise;
+    use crate::whitenoise::database::media_files::MediaFile;
     use crate::whitenoise::test_utils::*;
     use mdk_core::media_processing::MediaProcessingOptions;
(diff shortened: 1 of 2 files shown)


dannym-arx (approved):
Looks good to me, tho I think the long term solution is to include dimensions, blurhash and thumbhash in the video uploads, just for consistency.

mubarakcoded (approved):
LGTM!

PR #774, opened 2026-04-22
refactor: extract CLI into standalone whitenoise-cli workspace crate (Phase 19)

mubarakcoded:
## What this does

Extracts the CLI layer out of the core `whitenoise` library and into a new dedicated workspace crate: `crates/whitenoise-cli`.

Previously, `src/cli/`, `src/bin/wn.rs`, and `src/bin/wnd.rs` all lived inside the core library. This meant the core crate pulled in CLI-specific dependencies (`clap`, `dirs`, `rpassword`) for every consumer — including the Flutter FFI layer, which doesn't use them. This PR fixes that separation.

## Changes

**New crate: `crates/whitenoise-cli`**
- Contains all CLI source code previously under `src/cli/`
- Contains `wn` and `wnd` binaries previously under `src/bin/`
- Contains CLI integration tests previously under `tests/cli_e2e.rs` and `tests/cli_relay_control_e2e.rs`
- Depends on `whitenoise = { path = "../..", features = ["cli"] }` — the core library as a normal dependency

**Core library (`whitenoise`)**
- Removes `dirs` and `rpassword` from workspace dependencies
- Retains `cli = []` as an empty feature flag (still used in `src/whitenoise/mod.rs` for macOS keyring store selection)
- Makes `Relay::defaults()` `pub` — was `pub(crate)`, which was only fine when CLI lived inside the same crate
- Adds `UserSearchParams` and `ChatListItem` to the top-level `lib.rs` re-exports so consumers don't need to reference deep internal module paths

## Why the integration tests moved

`tests/cli_e2e.rs` and `tests/cli_relay_control_e2e.rs` use `env!("CARGO_BIN_EXE_wn")` to locate the compiled binary. That macro only resolves for bins declared in the same Cargo package, so the tests must live in `crates/whitenoise-cli` alongside the bins.

## What didn't change

- The `clap` dependency stays in the root `Cargo.toml` — it's still used by the `integration-tests` feature in core
- `indicatif` stays in core — used by `benchmark-tests`
- The integration test and benchmark bins (`integration_test.rs`, `benchmark_test.rs`, etc.) stay in core — they reference each other and use `env!("CARGO_BIN_EXE_wn")` which requires co-location

## Pre-existing failure

`test_archive_chat_emits_to_both_stream_channels` fails on the `arch-refactor` base branch and is unrelated to this PR.


mubarakcoded:
@claude, review this PR phase 19, make sure to reference session-projection-implementation-plan.md and session-projection-rearchitecture.md, @coderabbitai you too

mubarakcoded:
@claude, go for second round of review of this PR phase 19, make sure to reference session-projection-implementation-plan.md and session-projection-rearchitecture.md, @coderabbitai you too

crates/whitenoise-cli/src/bin/wn.rs
dannym-arx, line 1: We can probably remove the outer mod and turn all `whitenoise_cli::cli::blah` into `whitenoise_cli::blah`

mubarakcoded: Yes, make sense

dannym-arx (approved):

PR #781, opened 2026-04-27
feat(groups): soft-validate legacy key packages so mixed-version group creation succeeds

jgmontoya:
---

<a href="https://stagereview.app/marmot-protocol/whitenoise-rs/pull/781">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * More nuanced key-package compatibility and per-member capability projection affecting group creation and member invites
  * New explicit error for rejected invitees when an invitee’s key-package lacks required capabilities
  * Account activation can optionally skip immediate key-package publication (for tests/fixtures)

* **Tests**
  * Added integration scenarios for mixed-version groups and strict-add rejection
  * New test helpers and unit coverage for legacy-capability publishing and capability mapping

* **Chores**
  * Key-package maintenance and validation behavior updated to follow the new compatibility rules


src/whitenoise/key_packages.rs
```diff
@@ -56,11 +72,20 @@ pub(crate) fn has_encoding_tag(event: &Event) -> bool {
...
+/// (`filter_compatible_key_packages`) to decide whether to retain or rotate
+/// our own previously-published KPs. We never want to soft-validate our own
+/// output, so the strict check stays in place for that path.
+///
+/// For consumer-side paths (fetching a peer's KP, group creation, group
+/// membership) use [`validate_marmot_key_package_baseline`] instead — that
+/// validator drops the SelfRemove proposal/extension checks so legacy peers
+/// can be invited without preflight rejection (capability LCD is then
+/// computed by MDK).
+pub(crate) fn validate_marmot_key_package_strict(
```
erskingardner, line 88: > validate_marmot_key_package_strict and validate_marmot_key_package_baseline duplicate the kind, d-tag, encoding, base64, and ciphersuite checks. A small common helper would keep the strict/baseline policy split while preventing future drift in the hard requirements.

erskingardner (commented):
One minor comment inline and also this (which is outside the diff but might be worth looking into): 

> setup_key_package is part of the self-publish lifecycle: on login for an existing account, it decides whether this device already has a usable relay-published key package or should publish a fresh one. After this PR, fetch_user_key_package uses the relaxed baseline validator in src/relay_control/ephemeral.rs:823, so a locally tracked event that omits the SelfRemove proposal now returns KeyPackageLookup::Found. The subsequent is_own_key_package check only verifies that the event was tracked locally and key material is live; it does not re-apply validate_marmot_key_package_strict.
> 
> That means an account upgrading from a pre-SelfRemove app can log in, find its own old KP, and skip publishing the modern replacement. The scheduler may eventually fix this because maintenance still uses the strict validator, but login/open is the moment we need to refresh the account's advertised capabilities. This also contradicts the new strict-validator contract in key_packages.rs, which says self-publish lifecycle paths should not soft-validate our own output.
> 
> Use a strict check before setting needs_publish = false, or add a separate "fetch own live strict key package" path for account setup. A focused regression test should cover: tracked live KP missing SelfRemove -> setup_key_package(..., is_new_account = false, ...) publishes a new KP instead of skipping.

erskingardner (approved):

PR #784, opened 2026-04-29
Fix keyring store migration paths

erskingardner:
## Summary
- Update the keyring stack to `keyring-core` v1 and platform-native v1 stores, keeping host-side store initialization for direct `mdk-sqlite-storage` use.
- Replace Linux keyutils with D-Bus Secret Service, inject a stable non-default `whitenoise` target for Linux entries, and refresh Linux keyring-unavailable guidance.
- Add fallible keyring initialization so platform store creation errors propagate as `WhitenoiseError::KeyringUnavailable` instead of panicking.
- Add Android legacy-to-primary keyring migration support so existing v0.5 by-service credentials can be read and moved into the v1 named-store layout before encrypted DBs open.

## Why
The MDK/keyring upgrade changed platform credential behavior. On Android, `android-native-keyring-store` v1 defaults to the new named-store layout, while existing released builds used the v0.5 by-service layout. Existing installs can therefore have encrypted MLS databases on disk but appear to be missing `mdk.db.key.<pubkey>` unless the old layout is consulted. Linux keyutils also is not durable across normal restarts, so Secret Service is the better Linux backend.

## Impact
- Android upgrades can recover known legacy credentials such as `whitenoise.db.key.v1`, account private-key entries, and per-account `mdk.db.key.<pubkey>` entries through the default keyring path.
- Linux uses a durable Secret Service collection target consistently for create/read/delete operations, including MDK's plain `Entry::new(...)` path through the wrapped default store.
- Startup can now surface keyring backend failures as typed White Noise errors.

## Testing
- `just precommit-quick`
- `git diff --check`
- `cargo tree -e features --target aarch64-linux-android -i android-native-keyring-store@1.0.0`
- `cargo check --manifest-path /Users/jeff/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/android-native-keyring-store-1.0.0/Cargo.toml --target aarch64-linux-android --features legacy`
- Verified platform dependency graphs for Android, macOS, iOS, Windows, and Linux keyring stores.

## Notes
A full app-level `cargo check --target aarch64-linux-android --lib` still requires the Android SQLCipher/OpenSSL header environment; locally it stopped in `libsqlite3-sys` before reaching whitenoise code.

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-rs/pull/784">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **Chores**
  * Upgraded keyring libraries to v1, switched Linux storage to D‑Bus Secret Service, and added once-cell support for initialization.
* **New Features**
  * Safer, fallible keyring initialization that preserves an existing default and allows retrying on failure.
  * Scoped credential handling and automatic migration of legacy credentials into the primary store on access.
* **Bug Fixes**
  * Clearer, user-facing guidance when credential storage is unavailable; errors now surface reasons instead of panicking.

### DIFF
--- a/src/whitenoise/error.rs
+++ b/src/whitenoise/error.rs
@@ -260,3 +260,28 @@ impl From<Box<dyn std::error::Error + Send + Sync>> for WhitenoiseError {
         Self::Internal(err.to_string())
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn keyring_unavailable_display_includes_reason() {
+        let err = WhitenoiseError::KeyringUnavailable("Secret Service locked".to_string());
+
+        assert_eq!(
+            err.to_string(),
+            "Keyring storage unavailable: Secret Service locked"
+        );
+    }
+
+    #[test]
+    fn boxed_error_converts_to_internal_error() {
+        let boxed: Box<dyn std::error::Error + Send + Sync> =
+            Box::new(std::io::Error::other("boxed failure"));
+
+        let err = WhitenoiseError::from(boxed);
+
+        assert!(matches!(err, WhitenoiseError::Internal(msg) if msg == "boxed failure"));
+    }
+}
--- a/src/whitenoise/error.rs
+++ b/src/whitenoise/error.rs
@@ -29,6 +29,9 @@ pub enum WhitenoiseError {
     #[error("Configuration error: {0}")]
     Configuration(String),
 
+    #[error("Keyring storage unavailable: {0}")]
+    KeyringUnavailable(String),
+
     #[error("Contact list error: {0}")]
     ContactList(String),
 
--- a/src/whitenoise/error.rs
+++ b/src/whitenoise/error.rs
@@ -266,16 +266,6 @@ mod tests {
     use super::*;
 
     #[test]
-    fn keyring_unavailable_display_includes_reason() {
-        let err = WhitenoiseError::KeyringUnavailable("Secret Service locked".to_string());
-
-        assert_eq!(
-            err.to_string(),
-            "Keyring storage unavailable: Secret Service locked"
-        );
-    }
-
-    #[test]
     fn boxed_error_converts_to_internal_error() {
         let boxed: Box<dyn std::error::Error + Send + Sync> =
             Box::new(std::io::Error::other("boxed failure"));
--- a/src/whitenoise/error.rs
+++ b/src/whitenoise/error.rs
@@ -29,9 +29,6 @@ pub enum WhitenoiseError {
     #[error("Configuration error: {0}")]
     Configuration(String),
 
-    #[error("Keyring storage unavailable: {0}")]
-    KeyringUnavailable(String),
-
     #[error("Contact list error: {0}")]
     ContactList(String),
 
--- a/src/whitenoise/keyring_store.rs
+++ b/src/whitenoise/keyring_store.rs
@@ -237,6 +237,7 @@ mod tests {
     #[derive(Default)]
     struct RecordingStore {
         build_modifiers: Mutex<Vec<HashMap<String, String>>>,
+        credentials: Mutex<HashMap<(String, String), Arc<RecordingCredential>>>,
         search_specs: Mutex<Vec<HashMap<String, String>>>,
     }
 
@@ -272,15 +273,26 @@ mod tests {
 
         fn build(
             &self,
-            _service: &str,
-            _user: &str,
+            service: &str,
+            user: &str,
             modifiers: Option<&HashMap<&str, &str>>,
         ) -> Result<Entry> {
             self.build_modifiers
                 .lock()
                 .unwrap()
                 .push(owned_map(modifiers));
-            Ok(Entry::new_with_credential(Arc::new(TestCredential)))
+            let mut credentials = self.credentials.lock().unwrap();
+            let credential = credentials
+                .entry((service.to_string(), user.to_string()))
+                .or_insert_with(|| {
+                    Arc::new(RecordingCredential {
+                        service: service.to_string(),
+                        user: user.to_string(),
+                        secret: Mutex::new(None),
+                    })
+                })
+                .clone();
+            Ok(Entry::new_with_credential(credential))
         }
 
         fn search(&self, spec: &HashMap<&str, &str>) -> Result<Vec<Entry>> {
@@ -289,6 +301,43 @@ mod tests {
                 .unwrap()
                 .push(owned_map(Some(spec)));
             Ok(Vec::new())
+        }
+
+        fn as_any(&self) -> &dyn Any {
+            self
+        }
+    }
+
+    #[derive(Debug)]
+    struct RecordingCredential {
+        service: String,
+        user: String,
+        secret: Mutex<Option<Vec<u8>>>,
+    }
+
+    impl CredentialApi for RecordingCredential {
+        fn set_secret(&self, secret: &[u8]) -> Result<()> {
+            *self.secret.lock().unwrap() = Some(secret.to_vec());
+            Ok(())
+        }
+
+        fn get_secret(&self) -> Result<Vec<u8>> {
+            self.secret.lock().unwrap().clone().ok_or(Error::NoEntry)
+        }
+
+        fn delete_credential(&self) -> Result<()> {
+            match self.secret.lock().unwrap().take() {
+                Some(_) => Ok(()),
+                None => Err(Error::NoEntry),
+            }
+        }
+
+        fn get_credential(&self) -> Result<Option<Arc<Credential>>> {
+            Ok(None)
+        }
+
+        fn get_specifiers(&self) -> Option<(String, String)> {
+            Some((self.service.clone(), self.user.clone()))
         }
 
         fn as_any(&self) -> &dyn Any {
@@ -491,6 +540,30 @@ mod tests {
     fn target_is_non_default_for_wsl() {
         assert_eq!(LINUX_SECRET_SERVICE_TARGET, "whitenoise");
         assert_ne!(LINUX_SECRET_SERVICE_TARGET, "default");
+    }
+
+    #[test]
+    fn targeted_primary_store_migrates_legacy_secret() {
+        let primary_inner = Arc::new(RecordingStore::default());
+        let primary: Arc<CredentialStore> =
+            TargetedCredentialStore::new(primary_inner.clone(), LINUX_SECRET_SERVICE_TARGET);
+        let legacy = keyring_core::mock::Store::new().unwrap();
+        let legacy_entry = legacy
+            .build("com.whitenoise.app", "mdk.db.key.pubkey", None)
+            .unwrap();
+        legacy_entry.set_secret(b"legacy-mdk-key").unwrap();
+        let store = LegacyMigrationCredentialStore::new(primary, legacy);
+
+        let entry = store
+            .build("com.whitenoise.app", "mdk.db.key.pubkey", None)
+            .unwrap();
+
+        assert_eq!(entry.get_secret().unwrap(), b"legacy-mdk-key");
+        assert_eq!(
+            primary_inner.last_build_modifier("target"),
+            Some(LINUX_SECRET_SERVICE_TARGET.to_string())
+        );
+        assert!(matches!(legacy_entry.get_secret(), Err(Error::NoEntry)));
     }
 
     #[test]
--- a/src/whitenoise/keyring_store.rs
+++ b/src/whitenoise/keyring_store.rs
@@ -375,6 +375,13 @@ mod tests {
         }
 
         fn get_secret(&self) -> Result<Vec<u8>> {
+            if self.secret.is_empty() {
+                return Err(Error::Invalid(
+                    "legacy get".to_string(),
+                    "failed".to_string(),
+                ));
+            }
+
             Ok(self.secret.clone())
         }
 
@@ -420,6 +427,23 @@ mod tests {
     }
 
     #[test]
+    fn targeted_store_delegates_metadata_to_inner_store() {
+        let inner = Arc::new(RecordingStore::default());
+        assert!(format!("{inner:?}").contains("RecordingStore"));
+        let store: Arc<CredentialStore> =
+            TargetedCredentialStore::new(inner, LINUX_SECRET_SERVICE_TARGET);
+
+        assert_eq!(store.vendor(), "test-store");
+        assert_eq!(store.id(), "test-store");
+        assert!(matches!(
+            store.persistence(),
+            CredentialPersistence::UntilDelete
+        ));
+        assert!(store.as_any().is::<TargetedCredentialStore>());
+        assert!(format!("{store:?}").contains("TargetedCredentialStore"));
+    }
+
+    #[test]
     fn build_preserves_explicit_target() {
         let inner = Arc::new(RecordingStore::default());
         let store = TargetedCredentialStore::new(inner.clone(), LINUX_SECRET_SERVICE_TARGET);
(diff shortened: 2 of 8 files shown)


src/whitenoise/mod.rs
```diff
@@ -319,10 +319,15 @@ impl Whitenoise {
...
     ///
     /// This function is safe to call multiple times; only the first call has
-    /// an effect.
+    /// an effect. If another crate or host application has already configured
+    /// a `keyring-core` default store, that store is preserved.
     fn initialize_keyring_store() {
         static KEYRING_STORE_INIT: OnceLock<()> = OnceLock::new();
         KEYRING_STORE_INIT.get_or_init(|| {
+            if keyring_core::get_default_store().is_some() {
+                return;
```
dannym-arx: should we add a tracing log here?

dannym-arx (commented):
image

src/whitenoise/secrets_store.rs
```diff
@@ -147,11 +147,12 @@ fn format_storage_access_error(inner: &dyn std::error::Error) -> String {
...
-             occurs on headless systems, in SSH sessions, or in containers where no \
-             session keyring is active. Try running `keyctl session` before starting \
-             the daemon, or ensure your init system provides a session keyring. \
+            "Platform keyring is not available. On Linux, White Noise uses the D-Bus \
+             Secret Service to store secret keys. This error typically occurs on \
+             headless systems, in SSH sessions, in containers, or when no Secret \
+             Service implementation such as GNOME Keyring or KWallet is running and \
+             unlocked. Start and unlock a Secret Service provider before starting \
+             White Noise. \
              (Original error: {inner})"
```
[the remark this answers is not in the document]
dannym-arx: WSL is not a concern; if a user is running windows they will probably run a windows native bin

src/whitenoise/mod.rs
```diff
@@ -383,8 +388,8 @@ impl Whitenoise {
                 }
                 #[cfg(target_os = "linux")]
                 {
-                    let store = linux_keyutils_keyring_store::Store::new()
-                        .expect("Failed to create Linux keyutils credential store");
+                    let store = zbus_secret_service_keyring_store::Store::new()
+                        .expect("Failed to create Linux Secret Service credential store");
                     keyring_core::set_default_store(store);
```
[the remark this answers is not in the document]
dannym-arx: for headless this is explicitly acknowledged in secrets_store.rs

src/whitenoise/mod.rs
```diff
@@ -1,12 +1,15 @@
 use std::path::{Path, PathBuf};
-use std::sync::{Arc, OnceLock};
+use std::sync::Arc;
+#[cfg(test)]
+use std::sync::OnceLock;
```
jgmontoya: Why not move this to the test mod?

src/whitenoise/mod.rs
```diff
@@ -35,6 +38,15 @@ pub mod group_information;
...
 mod init_timing;
 pub mod key_packages;
+#[cfg(any(
+    test,
+    all(
+        any(target_os = "linux", target_os = "android"),
+        not(feature = "integration-tests"),
+        not(feature = "benchmark-tests")
+    )
+))]
```
jgmontoya, line 55: Why do we need this?

src/whitenoise/mod.rs
```diff
@@ -319,96 +363,146 @@ impl Whitenoise {
...
+            {
+                let store = Self::create_keyring_store(
+                    zbus_secret_service_keyring_store::Store::new(),
+                    "Linux Secret Service",
+                )?;
+                keyring_core::set_default_store(keyring_store::TargetedCredentialStore::new(
+                    store,
+                    keyring_store::LINUX_SECRET_SERVICE_TARGET,
+                ));
+            }
```
jgmontoya, line 453: I think this will be breaking for linux users. I think this is probably fine for now but in the future, as things like the CLI see more usage, we should probably be a bit more conservative with this kind of change

src/whitenoise/keyring_store.rs
```diff
@@ -0,0 +1,754 @@
...
+        self.primary.set_secret(&secret)?;
+        Self::delete_entry(&self.legacy)?;
+        Ok(secret)
+    }
+}
+
+impl CredentialApi for LegacyMigrationCredential {
+    fn set_secret(&self, secret: &[u8]) -> Result<()> {
+        self.primary.set_secret(secret)
+    }
```
jgmontoya, line 178: Question: should we delete the legacy entry here if it exists?

Something like:

```rs
fn set_secret(&self, secret: &[u8]) -> Result<()> {
    self.primary.set_secret(secret)?;
    Self::delete_entry(&self.legacy)?;
    Ok(())
}
```

src/whitenoise/keyring_store.rs
```diff
@@ -0,0 +1,754 @@
...
+        if let Some(err) = first_error {
+            return Err(err);
+        }
+
+        if deleted { Ok(()) } else { Err(Error::NoEntry) }
+    }
+
+    fn get_credential(&self) -> Result<Option<Arc<Credential>>> {
+        self.get_secret().map(|_| None)
+    }
```
jgmontoya, line 213: From my agent:

> The contract of `CredentialApi::get_credential` per `keyring-core` is metadata: return the underlying credential (or `None` to mean "I am the credential"). The wrapping `Entry::get_credential` already handles `Ok(None)` by returning a new Entry over `self.inner.clone()` — see `keyring-core-1.0.0/src/lib.rs:378-388`. So a caller doing `entry.get_credential().get_secret()` would call our `get_secret` *anyway* on the wrapped Entry, and the migration would run there.
> 
> Calling `get_secret()` here just to discard the value:
> 
> - Reads from the legacy store every time `get_credential` is called (potentially blocking I/O on Secret Service / Android Keystore).
> - Performs the legacy-delete side effect, surfacing the issue from #3 even on metadata calls.
> - Hides what's happening — a reader expects `get_credential` to be cheap and pure.
> 
> Suggested:
> 
> ```rust
> fn get_credential(&self) -> Result<Option<Arc<Credential>>> {
>     Ok(None)
> }
> ```
> 
> The migration still runs on `get_secret`. If anyone is currently relying on the side effect, that should be made explicit.

src/whitenoise/error.rs
```diff
@@ -29,6 +29,9 @@ pub enum WhitenoiseError {
     #[error("Configuration error: {0}")]
     Configuration(String),
 
+    #[error("Keyring storage unavailable: {0}")]
+    KeyringUnavailable(String),
```
jgmontoya: We had a `KeyringUnavailable` in `SecretsStoreError` (`src/whitenoise/secrets_store.rs`). I think we should probably only keep one of the two

src/whitenoise/mod.rs
```diff
@@ -319,96 +363,146 @@ impl Whitenoise {
...
+            #[cfg(target_os = "linux")]
+            {
+                let store = Self::create_keyring_store(
+                    zbus_secret_service_keyring_store::Store::new(),
+                    "Linux Secret Service",
+                )?;
+                keyring_core::set_default_store(keyring_store::TargetedCredentialStore::new(
+                    store,
+                    keyring_store::LINUX_SECRET_SERVICE_TARGET,
+                ));
```
jgmontoya: This got flagged by my agent. We could try to fix it in MDK:

> **KDE Wallet binary-data limitation — verified, real, and Linux-only.** From the same upstream docs: "you cannot use this module to store arbitrary binary secrets unless you first encode them as UTF-8 strings". I checked MDK at `../MDK`:
> 
> - `crates/mdk-sqlite-storage/src/encryption.rs:83-89` generates the DB key as `let mut key = [0u8; 32]; getrandom::fill(&mut key)?;` — 32 raw random bytes, almost never valid UTF-8.
> - `crates/mdk-sqlite-storage/src/keyring.rs:118` stores it via `entry.set_secret(config.key())` where `config.key()` returns `&[u8; 32]` (`encryption.rs:93`).
> - `crates/mdk-sqlite-storage/src/keyring.rs:152-160` reads it back with `entry.get_secret()` and validates `EncryptionConfig::from_slice(&secret)`, which expects exactly 32 raw bytes.
> 
> So on a KDE-based Linux desktop where KWallet is the Secret Service backend, MDK's `set_secret` of 32 random bytes will be rejected or corrupted by KWallet (which insists on UTF-8). The Whitenoise *Nostr private key* path is fine — `secrets_store.rs:54-62` uses `set_password` with `keys.secret_key().to_secret_hex()` (hex string, UTF-8 safe). But MDK's DB key path will fail.
> 
> KDE Plasma is the default desktop on Kubuntu, openSUSE KDE, KDE Neon, Manjaro KDE, Garuda KDE, etc. — not a niche audience. Concrete consequences:
> 
> - First-launch on KWallet-only systems: `mdk-sqlite-storage` either fails to create the keyring entry, or stores garbage and then the encrypted DB it creates is unreadable next time.
> - This is not something this branch can fix — it's an MDK behavior + zbus-secret-service-keyring-store limitation. But it should be tracked.
> 
> Suggested actions, ordered:
> 
> 1. File an upstream issue against `mdk-sqlite-storage` to encode the key (e.g. base64) before `set_secret` and decode after `get_secret`. Cheap, backward-incompatible (existing macOS/Windows/Android entries would need a migration), but solves all binary-secret-rejecting backends in one move.
> 2. Until that lands, document the KWallet limitation in Whitenoise's Linux installation notes and recommend GNOME Keyring as the Secret Service implementation.
> 3. Consider detecting KWallet at startup (read `org.freedesktop.secrets` introspection) and refusing to initialize MDK with a clear error rather than silently corrupting state.

src/whitenoise/mod.rs
```diff
@@ -1371,10 +1465,72 @@ pub mod test_utils {
...
+    }
+
+    #[test]
+    fn keyring_store_init_preserves_existing_default_store() {
+        let init = KeyringStoreInit::new();
+
+        let result = init.initialize_with(|| true, || panic!("initializer ran for existing store"));
+
+        assert!(result.is_ok());
+    }
```
jgmontoya, line 1628: I'm not sure I understand what this test is actually testing

src/whitenoise/mod.rs
```diff
@@ -319,96 +363,146 @@ impl Whitenoise {
...
+            #[cfg(target_os = "android")]
+            {
+                let primary = Self::create_keyring_store(
+                    android_native_keyring_store::Store::new(),
+                    "Android",
+                )?;
+                let legacy = Self::create_keyring_store(
+                    android_native_keyring_store::LegacyStore::from_ndk_context(),
+                    "legacy Android",
+                )?;
```
jgmontoya: From my agent:

> If the NDK context isn't initialized (hostile shutdown, race during process start), this fails *before* the primary store is even tried, and the user can't access their primary credentials either. Consider:
> 
> - Making the legacy half optional: try to create it, log warning on failure, and fall back to a primary-only store. Costs you the migration on that one failed startup but doesn't strand a user out of their already-migrated data.
> - Or document in the doc comment that this is intentionally fatal because we don't trust startups where NDK context is missing.
> 
> This is a Flutter-driven app, so the NDK context bring-up is presumably reliable, but the fail-closed behavior is worth a deliberate note.

jgmontoya (commented):
I think this will break headless CLI on linux :| I think we should have a way to keep kernel keyutils for that case

src/whitenoise/keyring_store.rs
mubarakcoded: Yes, and it's worth writing a unit test via primary, then assert legacy was deleted

src/whitenoise/error.rs
```diff
@@ -257,3 +260,28 @@ impl From<Box<dyn std::error::Error + Send + Sync>> for WhitenoiseError {
...
+    #[test]
+    fn boxed_error_converts_to_internal_error() {
+        let boxed: Box<dyn std::error::Error + Send + Sync> =
+            Box::new(std::io::Error::other("boxed failure"));
+
+        let err = WhitenoiseError::from(boxed);
+
+        assert!(matches!(err, WhitenoiseError::Internal(msg) if msg == "boxed failure"));
+    }
+}
```
dannym-arx, line 274: we don't need this test, it only tests that ThisError is working correctly imo

src/whitenoise/mod.rs
dannym-arx: why do you think this will break? most desktop users have dbus & libsecret, it's required by a lot of applications basically everyone uses, including web browsers. 

it only breaks for non-desktop users who of course usually don't have dbus... but that's a simple install on any distro

src/whitenoise/mod.rs
```diff
@@ -319,96 +363,146 @@ impl Whitenoise {
...
-                        "No keyring-core credential store available for this target OS. \
-                         Add a platform-specific store crate to Cargo.toml and handle it \
-                         in initialize_keyring_store()."
-                    );
-                }
+                compile_error!(
+                    "No keyring-core credential store available for this target OS. \
+                     Add a platform-specific store crate to Cargo.toml and handle it \
+                     in initialize_keyring_store()."
+                );
```
dannym-arx, line 493: BSDs can also use dbus & libsecret, just like linux.

this will needlessly break on freebsd/openbsd, when the same linux-specific tools will work

dannym-arx (changes requested):
mostly fine but left a few comments

src/whitenoise/mod.rs
jgmontoya: because there is no migration path like there is for android

src/whitenoise/mod.rs
erskingardner: created this issue: https://github.com/marmot-protocol/mdk/issues/268

src/whitenoise/mod.rs
erskingardner: @jgmontoya I've now added a migration path and we fail hard on lack of a compatible keyring store. I also added a paragraph to the CLI readme about the requirement for a keyring tool

src/whitenoise/mod.rs
erskingardner: I've updated this to hard fail when there isn't a keyring tool installed or available and added details to the CLI readme to explain.

dannym-arx: yeah but the target_os macros still won't catch those

dannym-arx (commented):
looks good, just one minor comment

src/whitenoise/mod.rs
erskingardner: This is needed because keyring_store is only compiled where Whitenoise installs a custom keyring-core store: tests plus Linux/Android production builds. Integration and benchmark builds use the mock store path, so excluding the module there prevents Linux all-features cfg/dead-code breakage.

src/whitenoise/mod.rs
erskingardner: Addressed now. Linux uses Secret Service as the durable primary store and wraps it with keyutils legacy migration for existing users. Headless/runtime requirements are documented in src/cli/README.md.

src/whitenoise/keyring_store.rs
erskingardner: Addressed. LegacyMigrationCredential::get_credential now returns Ok(None), so it stays metadata-only. Migration/deletion side effects only happen when get_secret or set_secret is called.

src/whitenoise/mod.rs
erskingardner: Clarified in 0824373d: the test is now named keyring_store_init_preserves_existing_default_store and has a comment explaining the host-app default-store case.

src/whitenoise/mod.rs
erskingardner: Agreed this is worth supporting, but I am keeping PR #784 scoped to Linux because the current change also carries Linux keyutils migration semantics. I opened #793 to track explicit BSD Secret Service support.

dannym-arx (approved):

PR #785, opened 2026-04-30
refactor(db): unify global+local migration timeline with fresh-account bootstrap

dannym-arx:
marmot

Unifies global and per-account migrations into one ordered timeline so the marmot colony's databases stay in lockstep no matter which burrow opens first.

### What changed

Globals and locals used to live in two separate version sequences, each starting at 1. That made it impossible to land a migration sequence like "global X, then local A, then global Y" where A reads schema added by X and removed by Y. The runner had no notion of "apply X first, then A on every account, then Y"; it only knew "all globals, then all locals."

This PR collapses both into a single `Migrator` that:

- Stores every migration in one version-sorted timeline. Uniqueness is asserted across both kinds.
- Walks the timeline once on every `run`. Globals dispatch to the shared DB; locals dispatch to the supplied account DB (if any).
- Adds `run_all` for lockstep walks across the shared DB and every account DB at startup. Use this once per-user DB files become physically separate (Phase 18b+).

### New: account-fresh bootstrap

`LocalMigration::for_new_accounts_only()` lets a migration declare itself as fresh-account-only. The runner snapshots `_rust_migrations` at the start of the walk and skips such migrations on any account that already shows local-timeline progress. The skip is silent, with no row written.

Freshness is gauged against local versions only. Globals stamped on the same file (which happens today because Phase 18a's per-user file is still opened by the generic `Database` constructor) do not count against freshness. Otherwise every account would look "non-fresh" the moment it opened.

The first such migration is `m0012_bootstrap`, which loads `fresh_account_schema.sql` for new per-user files and stamps any prior local versions whose effect that snapshot already captures. Today the stamp list is empty; the rebaseline that goes with Phase 18b extraction work will populate it.

### Wiring

- `Database::new` and `Database::migrate_up` now call `MIGRATOR.run(&pool, None)` (shared-only mode).
- `AccountDatabase::run_account_migrations(&shared)` runs the timeline against the per-user pool, including the bootstrap on a fresh file.
- The eleven existing globals moved out of the `global/` submodule into a flat layout next to the new local. Their version numbers are unchanged.

### Tests

The bootstrap suite covers: fresh-account stamping, idempotency, rebaseline-style prior-version stamping via a probe migration, the silent skip on a pre-existing account, and a regression test proving globals stamped on the same file do not count toward freshness. Existing global runner coverage is preserved against the unified `Migrator`.

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-rs/pull/785">
  <picture>
    Open in Stage
  </picture>
</a>


mubarakcoded (approved):
This looks good Danny, nice work on the unified timeline and the fresh-account bootstrap

PR #787, opened 2026-04-30
Phase 18b: media_files → media_blobs + media_references split

mubarakcoded:
## Summary

Splits the monolithic `media_files` table into two tables as part of Phase 18b:

- **`media_blobs`** — shared, content-addressed blob cache deduplicated by `encrypted_file_hash`
- **`media_references`** — account-scoped references linking blobs to MLS groups

### Migration design

Two global migrations handle the full split:

**m0013 (`media_blob_reference_split`):**
1. Creates `media_blobs` with a partial index on `blossom_url`
2. Creates `media_references` with indexes on `account_pubkey`, `(mls_group_id, encrypted_file_hash)`, and `created_at`
3. Backfills `media_blobs` from `media_files` (deduplicating by hash, picking first non-empty `file_path` and non-NULL `blossom_url`)
4. Backfills `media_references` from `media_files` (preserving all per-account rows)

**m0014 (`drop_media_files`):**
5. Drops the legacy `media_files` table — safe because m0013 atomically backfills both new tables in a single global transaction, so no data is lost

### Why `media_references` stays on shared (not a local migration)

`media_references` is logically account-scoped, but `AccountDatabase` is not wired into production init yet (Phase 18a scaffold only). `Database::new` calls `MIGRATOR.run(&pool, None)` — shared-only mode that skips all local migrations. All production queries in `media_files.rs` do `INNER JOIN media_references r ... media_blobs b` on `database.pool` (shared). If `media_references` were created by a local migration, it would never exist on the shared pool and every media query would fail with "no such table."

The table stays on shared until Phase 18c+ wires `AccountDatabase` into production init, at which point a local migration extracts it and a follow-up global drops it from shared.

### Why the FK constraint was removed

The `FOREIGN KEY (encrypted_file_hash) REFERENCES media_blobs(encrypted_file_hash)` was removed because once `media_references` moves to the account DB (Phase 18c+), the FK would be cross-database and unenforced by SQLite anyway. Removing it now avoids a schema change during extraction.

## Changes

- **m0013**: Creates `media_blobs` and `media_references` on shared, backfills both from `media_files`
- **m0014**: Drops legacy `media_files` table
- **fresh_account_schema.sql**: Updated comment to reflect Phase 18b status
- **All query methods** in `media_files.rs`: Rewritten to JOIN `media_references` + `media_blobs`
- **`save()`**: Wrapped in transaction; blob upsert uses `COALESCE` to promote `blossom_url` and `file_path` on subsequent writes
- **`find_by_hash()`**: Added `ORDER BY r.created_at ASC, r.id ASC` for deterministic `LIMIT 1`
- **`update_file_path()`**: Wrapped in transaction; doc comment documents shared-blob cross-account mutation caveat

## Test plan

- [x] `just precommit-quick` passes (fmt, docs, clippy, dead_code, tests)
- [x] Only pre-existing login test failures (3 tests in `accounts::login::tests` — failing on base `arch-refactor` branch)
- [x] m0013 test: blob deduplication (2 blobs from 3 rows), reference preservation (3 refs), first-writer-wins for `file_path`, non-NULL `blossom_url` selection, `media_files` still exists after m0013
- [x] m0014 test: `media_files` dropped, `media_blobs` and `media_references` survive
- [x] Shared blob survival test: deleting one account's references leaves the blob intact for other accounts

### DIFF
--- a/docs/session-projection-implementation-plan.md
+++ b/docs/session-projection-implementation-plan.md
@@ -512,7 +512,7 @@ boundary.
 
 ### Phase 18b: Media blob/reference split (~800 LOC)
 
-- [x] Completed in PR #xxx
+- [x] Completed in PR #787
 
 <details>
 <summary>Details</summary>
--- a/docs/session-projection-implementation-plan.md
+++ b/docs/session-projection-implementation-plan.md
@@ -512,7 +512,7 @@ boundary.
 
 ### Phase 18b: Media blob/reference split (~800 LOC)
 
-- [ ] Not started
+- [x] Completed in PR #xxx
 
 <details>
 <summary>Details</summary>
--- a/src/whitenoise/database/media_files.rs
+++ b/src/whitenoise/database/media_files.rs
@@ -198,6 +198,7 @@ impl MediaFile {
              INNER JOIN media_blobs b
                 ON b.encrypted_file_hash = r.encrypted_file_hash
              WHERE r.encrypted_file_hash = ?
+             ORDER BY r.created_at ASC, r.id ASC
              LIMIT 1",
         )
         .bind(&encrypted_file_hash_hex)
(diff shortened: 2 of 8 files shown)


mubarakcoded:
@claude, review this PR, make sure to reference session-projection-implementation-plan.md and session-projection-rearchitecture.md @coderabbitai you too

mubarakcoded:
@claude @coderabbitai review again, actionable items fixed

mubarakcoded:
@claude @coderabbitai review again, all actionable items fixed.

dannym-arx (approved):

PR #791, opened 2026-05-03
feat(groups): add required proposal upgrades for legacy groups

jgmontoya:
---

<a href="https://stagereview.app/marmot-protocol/whitenoise-rs/pull/791">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added group capability upgrade status checking to query proposal upgrade readiness
  * Added functionality to upgrade and manage required proposals within groups
  * Exposed new types for managing required proposals and capability upgrade status

* **Bug Fixes**
  * Improved storage initialization with mutex-based synchronization to handle concurrent access

* **Chores**
  * Updated upstream dependencies

### DIFF
--- a/src/whitenoise/accounts/mod.rs
+++ b/src/whitenoise/accounts/mod.rs
@@ -11,6 +11,7 @@ use serde::{Deserialize, Serialize};
 use std::fmt;
 use std::path::Path;
 use std::str::FromStr;
+use std::sync::{LazyLock, Mutex};
 use thiserror::Error;
 
 use crate::RelayType;
@@ -24,6 +25,8 @@ use crate::whitenoise::secrets_store::SecretsStoreError;
 use crate::whitenoise::user_streaming::{UserUpdate, UserUpdateTrigger};
 use crate::whitenoise::users::User;
 use crate::whitenoise::{Whitenoise, WhitenoiseError};
+
+static MDK_STORAGE_INIT_LOCK: LazyLock<Mutex<()>> = LazyLock::new(|| Mutex::new(()));
 
 /// The type of account authentication.
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Default, Serialize, Deserialize)]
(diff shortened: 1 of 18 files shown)


src/whitenoise/groups/required_proposals.rs
```diff
@@ -46,6 +48,28 @@ pub enum RequiredProposal {
     Unknown,
 }
 
+/// Upgrade readiness for the proposal types WhiteNoise exposes.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize)]
```
erskingardner: Should this also derive `Deserialize`? Same for `RequiredProposalUpgradeStatus` below.

src/whitenoise/groups.rs
```diff
@@ -638,6 +642,77 @@ impl Whitenoise {
...
+            .group_capability_upgrade_status(group_id)
+            .map_err(|e| match e {
+                mdk_core::Error::GroupNotFound => WhitenoiseError::GroupNotFound,
+                other => WhitenoiseError::from(other),
+            })?;
+        Ok(project_group_capability_upgrade_status(status))
+    }
+
+    #[perf_instrument("groups")]
+    pub async fn upgrade_group_required_proposals(
```
erskingardner, line 691: We should add a doc comment since this is a public method

src/whitenoise/accounts/mod.rs
```diff
@@ -25,6 +26,8 @@ use crate::whitenoise::user_streaming::{UserUpdate, UserUpdateTrigger};
 use crate::whitenoise::users::User;
 use crate::whitenoise::{Whitenoise, WhitenoiseError};
 
+static MDK_STORAGE_INIT_LOCK: LazyLock<Mutex<()>> = LazyLock::new(|| Mutex::new(()));
```
erskingardner, line 29: This only guards against same process races - which is all we really need to deal with right now I think. But something I've started into with the push notifications thing is a separate process/isolate for doing the fetch for new messages in a more streamlined way. 

I would say that we don't necessarily need any changes here now but wanted to flag.

erskingardner (commented):
A few comments.

src/whitenoise/groups.rs
```diff
@@ -638,6 +642,77 @@ impl Whitenoise {
...
+            ));
+        }
+
+        let proposal_types = proposals_to_add
+            .iter()
+            .copied()
+            .map(ProposalType::try_from)
+            .collect::<Result<BTreeSet<_>>>()?;
+
+        self.ensure_account_is_group_admin(account, group_id)
```
erskingardner: minor nit: shouldn't this validation happen just above the `proposal_types` conversion just above?

src/whitenoise/groups/required_proposals.rs
jgmontoya: done

src/whitenoise/groups.rs
jgmontoya: done

src/whitenoise/accounts/mod.rs
jgmontoya: 👍 agreed it's sufficient for the current single-process model. We can revisit whenever we have another process

src/whitenoise/groups.rs
jgmontoya: Good catch, moved!

erskingardner (approved):

PR #799, opened 2026-05-05
[codex] Fix stale database keyring rotation

### LINKED ISSUE #798: Rotate app database key when creating a fresh encrypted database
## Problem

`whitenoise.sqlite` can reuse a stale keyring entry when the app database file no longer exists.

`Database::new_encrypted()` calls `prepare_sqlcipher_database(...)`. When the database state is `MissingOrEmpty` or `Plaintext`, that path currently calls `get_or_create_key(...)`, which delegates to MDK keyring helpers. If the app data directory was removed but the platform secret store kept or restored `whitenoise.db.key.v1`, a fresh app database can be created with the old key.

This is the app-level version of the MDK issue for per-account MLS databases: https://github.com/marmot-protocol/mdk/issues/274

It can also explain the observed startup failure after downgrade testing: a newer build using MDK #270 stores a 62-byte encoded keyring payload, then an older build reads that payload as if it were a raw 32-byte database key and fails before account/MLS storage is opened.

## Affected paths

- `Database::new_encrypted(...)`
- `prepare_sqlcipher_database(...)`
- key id `whitenoise.db.key.v1`
- app wipe/delete flows that remove database files but leave keyring entries behind

## Fix

For a fresh app database lifecycle, rotate the app DB key instead of reusing it:

- if `whitenoise.sqlite` is missing or empty and there are no migration recovery sidecars, delete `whitenoise.db.key.v1` and generate a new key
- if migrating a plaintext database and there are no migration recovery sidecars, use a fresh key for the encryption migration
- if `.encrypted.tmp` or `.plaintext.backup` exists, preserve the existing key and run recovery so interrupted migrations remain recoverable
- keep existing encrypted database behavior unchanged: require the existing key and fail if it is missing

Also update deletion paths:

- app-level wipe should delete `whitenoise.db.key.v1`
- account deletion/app wipe should delete per-account MDK keys (`mdk.db.key.<pubkey>`) when removing MLS database directories

## Tests

Add coverage for:

- stale `whitenoise.db.key.v1` exists but `whitenoise.sqlite` is absent: initialization creates a new key
- existing encrypted `whitenoise.sqlite` still reopens with the existing key
- existing encrypted DB missing its key still fails
- interrupted encryption migration sidecars still recover with the original key
- wipe/delete code removes app and per-account database keyring entries

erskingardner:
## Summary

- Rotate stale Whitenoise app database keyring entries when creating a fresh encrypted database or encrypting a plaintext database.
- Preserve existing keyring entries during interrupted migration recovery sidecar handling.
- Delete app and per-account MDK database keyring entries during app wipe, and delete per-account MDK storage/keyring entries on logout.
- Pin MDK to merged master commit `634085a919ffac725323f7085291d945979af507`, which includes marmot-protocol/mdk#275 for fresh per-account MLS database key rotation.

## Root Cause

If a database file was removed while its keyring entry remained, fresh database creation could reuse stale SQLCipher key material. That was wrong for a fresh app install/wipe and could also affect account-scoped MDK storage when MLS database files were removed independently.

## Validation

- `cargo metadata --locked --no-deps --format-version 1`
- `just precommit-quick`

Fixes #798.

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-rs/pull/799">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Logout removes per-account storage and clears associated encryption keys.
  * Global data deletion now closes databases, deletes underlying files, and removes app and orphaned per-account keys.
  * Safer encryption/migration recovery with serialized key rotation and improved file cleanup to avoid stale-key issues.

* **Chores**
  * Aligned core/storage dependency versions for consistency.

* **Tests**
  * Added/updated tests verifying keyring cleanup and migration recovery.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2691,7 +2691,7 @@ dependencies = [
 [[package]]
 name = "mdk-core"
 version = "0.8.0"
-source = "git+https://github.com/marmot-protocol/mdk?rev=575ae29d25d58494135058d9e46affec7174e79a#575ae29d25d58494135058d9e46affec7174e79a"
+source = "git+https://github.com/marmot-protocol/mdk?rev=634085a919ffac725323f7085291d945979af507#634085a919ffac725323f7085291d945979af507"
 dependencies = [
  "base64",
  "blurhash",
@@ -2719,12 +2719,12 @@ dependencies = [
 [[package]]
 name = "mdk-macros"
 version = "0.8.0"
-source = "git+https://github.com/marmot-protocol/mdk?rev=575ae29d25d58494135058d9e46affec7174e79a#575ae29d25d58494135058d9e46affec7174e79a"
+source = "git+https://github.com/marmot-protocol/mdk?rev=634085a919ffac725323f7085291d945979af507#634085a919ffac725323f7085291d945979af507"
 
 [[package]]
 name = "mdk-sqlite-storage"
 version = "0.8.0"
-source = "git+https://github.com/marmot-protocol/mdk?rev=575ae29d25d58494135058d9e46affec7174e79a#575ae29d25d58494135058d9e46affec7174e79a"
+source = "git+https://github.com/marmot-protocol/mdk?rev=634085a919ffac725323f7085291d945979af507#634085a919ffac725323f7085291d945979af507"
 dependencies = [
  "base64",
  "getrandom 0.4.2",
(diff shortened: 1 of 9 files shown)


jgmontoya (approved):
LGTM!

PR #804, opened 2026-05-06
Make chat-list unread badges ignore blocked authors

erskingardner:
## Summary
- Made chat-list unread counts block-aware by filtering out messages authored by blocked users in the Rust projection layer.
- Kept raw unread-count helpers intact for storage, history, and read-marker behavior.
- Updated block/unblock chat-list refreshes so streaming payloads recalculate visible unread counts.
- Added tests for blocked-only, mixed blocked/non-blocked, unblock restoration, and streaming update behavior.

## Testing
- `cargo test --lib test_get_chat_list_unread_count --features cli`
- `cargo test --lib user_block_changed_update_recalculates_group_unread_count --features cli`
- `just precommit-quick`
- `just coverage`

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-rs/pull/804">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://stagereview.app/assets/gh-open-in-stage-dark.svg">
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Unread message counts now respect visibility (blocked/muted authors) and are recalculated across all visible chats when block/unblock changes occur.
  * Block/unblock actions now trigger a single, correct refresh for affected chats.

* **Tests**
  * Expanded coverage for unread-count behavior, block/unblock emission, and related flows.

* **New Features**
  * Improved CLI group response formatting for clearer group details.

PR #810, opened 2026-05-07
feat: move mute_list to per-account database (m0042 + m0043)

mubarakcoded:
## Summary

Moves the `mute_list` table from the shared SQLite database into each account's per-account DB file, then drops the now-empty shared table. This is the last bit of per-account state that hadn't migrated yet, and it's blocking a clean re-merge of master.

## Why now

Master's PR #804 (`feat: ignore blocked authors in unread badges`) added a `count_visible_unread_for_group` query that joins `aggregated_messages` (per-account post-Phase 18d) with `mute_list` (shared). That cross-DB subquery can't run on arch-refactor — it fails with `no such table: mute_list` against the per-account pool.

Rather than bridge with `ATTACH DATABASE`, this moves the table to where it logically belongs. After this lands, the next master merge brings master's signature changes through cleanly because both tables now live in the same DB.

## What changed

**Migrations**

- `m0042_move_mute_list` (LocalMigration, v42) creates per-account `mute_list` and copies rows from shared filtered to the owning account. Per-row `rows_affected()` accounting emits `tracing::warn!` on any silently-skipped constraint violations so corrupt source rows leave a diagnostic trail.
- `m0043_drop_shared_mute_list` (GlobalMigration, v43) drops the shared table and its account-scoped index. Runs after every account has completed v42, in the same boot.
- `account_pubkey` column dropped — redundant once the table is per-account. UNIQUE collapses from `(account_pubkey, muted_pubkey)` to `muted_pubkey`.
- Schema baked into `fresh_account_schema.sql` so future bootstrap rebaselines pick it up.

**DAO (`database/mute_list.rs`)**

- `MuteListEntry` no longer carries `account_pubkey`.
- All methods take `&AccountDatabase` instead of `&Database` and drop the `account_pubkey` parameter.
- `find_by_account` → `find_all`; `find_by_account_and_target` → `find_by_muted_pubkey` (matches column-name suffix convention used across other DAOs).
- `exists()` uses `SELECT EXISTS(SELECT 1 ...)` for short-circuit boolean semantics.

**Session (`session/mute_list.rs`)**

- `db()` returns `&AccountDatabase`.
- `account_relay_urls()` reaches into `&shared.database` directly for `Account` / `User` / `Relay` reads (those tables stay shared).
- `unblock_user` rollback path: explicit `match` with `tracing::debug!` when the entry vanishes between exists-check and read, instead of a silent `unwrap_or(true)`.

**Docs**

- `AccountDatabase` doc comment now lists `mute_list` alongside the other per-account tables.

## Migration safety

- Existing accounts: rows copy from `shared.mute_list WHERE account_pubkey = ?` into the per-account file before v43 drops shared.
- Fresh accounts: bootstrap creates the table from `fresh_account_schema.sql`; m0042's `CREATE TABLE IF NOT EXISTS` is a no-op; data copy hits empty shared; m0043 drops the empty shared table.
- Idempotency: each migration runs in its own transaction; `INSERT OR IGNORE` makes re-runs safe; `DROP TABLE IF EXISTS` is idempotent.
- Corrupt source rows (CHECK violations on hex format) are silently skipped rather than aborting startup — a `tracing::warn!` records the count, and `migration_silently_skips_invalid_pubkey_rows` test pins this contract.
- Ordering across kinds: the migrator processes a unified version timeline, so all per-account v42 instances complete before the v43 global drop runs.

## Downstream impact

`MuteListEntry` is publicly re-exported via `lib.rs:72`. The `marmot-protocol/whitenoise` Flutter app's bridge layer at `whitenoise/rust/src/api/mute_list.rs` reads `e.account_pubkey.to_hex()` (line 20) plus a unit test that constructs the struct directly. The released app pins whitenoise-rs by git rev so it isn't retroactively broken — but the next rev bump on the Flutter side requires:

- Drop `account_pubkey: String` field from the bridge `MuteListEntry` struct
- Drop `account_pubkey: e.account_pubkey.to_hex()` from the `From` impl
- Update the `WhitenoiseEntry { ... }` literals in the two unit tests
- Run `flutter_rust_bridge_codegen generate` + `dart run build_runner build`

No Dart consumers of `accountPubkey` exist outside the auto-generated bridge code, so the Flutter app side is ~8 lines of Rust + a regen.

## Follow-ups (not in this PR)

- Companion PR on `marmot-protocol/whitenoise` that lands together with the eventual arch-refactor → master merge (drops `account_pubkey` from the bridge struct, regen).

## Test plan

- [x] `just precommit-quick` passes — fmt, docs, clippy, dead_code, all unit tests
- [x] m0042 tests: account-scoped copy, empty source, missing shared table, silent skip on invalid hex
- [x] m0043 tests: drops table when present, idempotent when absent
- [x] DAO tests: insert/exists/delete/find_all/find_by_muted_pubkey/sync_from_event with the new schema
- [x] Session tests: block/unblock/sync_and_emit go through the per-account DB

### DIFF
--- a/src/whitenoise/database/rust_migrations/m0042_move_mute_list.rs
+++ b/src/whitenoise/database/rust_migrations/m0042_move_mute_list.rs
@@ -1,0 +1,300 @@
+use async_trait::async_trait;
+use sqlx::{SqliteConnection, SqlitePool};
+
+use crate::whitenoise::database::DatabaseError;
+use crate::whitenoise::database::rust_migrations::LocalMigration;
+
+/// Temporary carrier for rows read from the shared `mute_list` table during
+/// migration. Account scoping happens in the WHERE clause, so the
+/// `account_pubkey` column isn't materialised here.
+#[derive(sqlx::FromRow)]
+struct SharedMuteListRow {
+    muted_pubkey: String,
+    is_private: i64,
+    created_at: i64,
+}
+
+pub struct Migration;
+
+#[async_trait]
+impl LocalMigration for Migration {
+    fn version(&self) -> u32 {
+        42
+    }
+
+    fn description(&self) -> &'static str {
+        "Move mute_list into per-account database"
+    }
+
+    async fn run_local(
+        &self,
+        tx: &mut SqliteConnection,
+        global_db: &SqlitePool,
+        account_pubkey: &str,
+    ) -> Result<(), DatabaseError> {
+        // Per-account schema drops the redundant `account_pubkey` column —
+        // ownership is implicit in the file. UNIQUE collapses to muted_pubkey
+        // alone since each account file holds only its own entries.
+        sqlx::query(
+            "CREATE TABLE IF NOT EXISTS mute_list (
+                id            INTEGER PRIMARY KEY AUTOINCREMENT,
+                muted_pubkey  TEXT NOT NULL UNIQUE
+                    CHECK (length(muted_pubkey) = 64
+                           AND muted_pubkey GLOB '[0-9a-fA-F]*'),
+                is_private    INTEGER NOT NULL DEFAULT 1
+                    CHECK (is_private IN (0, 1)),
+                created_at    INTEGER NOT NULL
+            )",
+        )
+        .execute(&mut *tx)
+        .await?;
+
+        let shared_table_exists: bool = sqlx::query_scalar(
+            "SELECT EXISTS(SELECT 1 FROM sqlite_master \
+             WHERE type='table' AND name='mute_list')",
+        )
+        .fetch_one(global_db)
+        .await?;
+
+        if !shared_table_exists {
+            tracing::warn!(
+                target: "whitenoise::database::rust_migrations::m0042",
+                account = account_pubkey,
+                "shared.mute_list is missing; skipping local copy. \
+                 Expected only on fresh installs."
+            );
+            return Ok(());
+        }
+
+        let rows: Vec<SharedMuteListRow> = sqlx::query_as(
+            "SELECT muted_pubkey, is_private, created_at
+             FROM mute_list
+             WHERE account_pubkey = ?",
+        )
+        .bind(account_pubkey)
+        .fetch_all(global_db)
+        .await?;
+
+        let copied = rows.len() as u64;
+        for r in &rows {
+            sqlx::query(
+                "INSERT OR IGNORE INTO mute_list (muted_pubkey, is_private, created_at)
+                 VALUES (?, ?, ?)",
+            )
+            .bind(&r.muted_pubkey)
+            .bind(r.is_private)
+            .bind(r.created_at)
+            .execute(&mut *tx)
+            .await?;
+        }
+
+        if copied > 0 {
+            tracing::info!(
+                target: "whitenoise::database::rust_migrations::m0042",
+                account = account_pubkey,
+                "Copied {copied} mute_list row(s) to per-account DB"
+            );
+        }
+
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use sqlx::SqlitePool;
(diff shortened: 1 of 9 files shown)


mubarakcoded:
@claude review this PR, @coderabbitai and you too

src/whitenoise/database/rust_migrations/m0042_move_mute_list.rs
```diff
@@ -0,0 +1,318 @@
...
+                "INSERT OR IGNORE INTO mute_list (muted_pubkey, is_private, created_at)
+                 VALUES (?, ?, ?)",
+            )
+            .bind(&r.muted_pubkey)
+            .bind(r.is_private)
+            .bind(r.created_at)
+            .execute(&mut *tx)
+            .await?;
+
+            if result.rows_affected() == 0 {
```
erskingardner, line 91: > `INSERT OR IGNORE` returns `rows_affected() == 0` for *both* a CHECK-constraint
> violation (malformed pubkey) *and* a UNIQUE conflict (duplicate row). In the
> context of a fresh per-account table that starts empty, a UNIQUE conflict can
> only happen if the migration runs twice (idempotency path). Today the warning
> message says "constraint violation — likely malformed muted\_pubkey", which
> would be misleading if the actual cause is a UNIQUE collision from a re-run.
> 
> The comment in the code notes this ambiguity ("CHECK / UNIQUE constraint
> failures get swallowed"), but the warning text presented to operators doesn't
> reflect it. Either:
> 
> - Split the insert query into a prior `SELECT EXISTS(...)` check so the two
>   cases can be distinguished, or
> - Update the warning message to name both possible causes so an operator
>   reading the log is not misled into thinking their data is corrupt when it
>   is actually a harmless re-run.
> 
> The second option is the pragmatic fix here.

src/whitenoise/session/mute_list.rs
```diff
@@ -110,25 +106,33 @@ impl<'a> MuteListOps<'a> {
...
-        // entry exactly as it was.
-        let is_private =
-            MuteListEntry::find_by_account_and_target(account_pubkey, target_pubkey, self.db())
-                .await?
-                .map(|e| e.is_private)
-                .unwrap_or(true);
+        // entry exactly as it was. If the entry vanished between the
+        // exists-check above and now (concurrent sync from another flow),
+        // default to private — the safer choice for a rollback.
+        let is_private = match MuteListEntry::find_by_muted_pubkey(target_pubkey, self.db()).await?
```
erskingardner, line 120: > The `None` branch defaults `is_private = true` for the rollback and logs a
> debug-level message. The PR description describes this as "the safer choice
> for a rollback." That is correct in isolation, but if the entry genuinely
> vanished (concurrent sync wiped it), the subsequent `delete` will silently
> succeed against a non-existent row, and then `publish_mute_list` will publish
> a list that no longer includes the target — which is arguably the right
> outcome. However, the `is_private = true` default is only meaningful for the
> rollback `insert`, and a rollback `insert` on a row that was already removed
> by a concurrent sync would *re-add* that row.
> 
> The rollback path (`MuteListEntry::insert(target_pubkey, is_private, ...)`)
> uses `INSERT OR IGNORE`, so the concrete risk is: concurrent sync deletes the
> row → `find_by_muted_pubkey` returns `None` → `delete` is a no-op → publish
> fails → rollback `insert OR IGNORE` re-inserts the row (since it no longer
> exists). The user ends up still blocked when they should not be.
> 
> This is an edge case, but the `None` branch silently makes the wrong
> decision. A `debug!` log alone is insufficient — at minimum a `warn!` should
> fire because the caller needs to know a logical inconsistency occurred. More
> robustly: if `find_by_muted_pubkey` returns `None`, skip the delete and the
> publish (the user is already not in the local list), and return `Ok(())`.

dannym-arx (approved):

PR #822, opened 2026-05-12
Merge master into arch-refactor (post-#818)

mubarakcoded:
## What this PR does

Brings `arch-refactor` (currently at `ee30f7a` — Danny's #818 deprecated-facade removal) onto a branch that has the full master tip integrated, and fixes the CLI dispatch layer that #818 left broken.

## Commits

1. **`d8a59fe` — Merge `arch-refactor` into the post-mute-list integration branch**

   Resolves 4 small conflicts, all caused by #818's facade removal colliding with HEAD's older call sites:

   - `src/integration_tests/test_cases/group_membership/leave_group.rs` — doc comment refreshed to `AccountSession::groups().leave()`
   - `src/integration_tests/test_cases/scheduler/key_package_maintenance.rs` — three facade calls migrated to `session.key_packages().fetch_all()` / `delete_batch(...)`
   - `src/whitenoise/accounts_groups.rs` — eight test-side facade calls migrated to `session.membership().for_group(...).{get_or_create, mark_as_left, mark_as_removed}()`
   - `src/whitenoise/chat_list.rs` — drops the four deprecated facades (`get_chat_list`, `get_archived_chat_list`, `get_chat_list_item`, `build_chat_list_item`); emit path already uses `session.chat_list().build_item()`

   Master tip `59aa42c` (group state streaming + signer recovery + 22 others) was already integrated on this branch — nothing new from master is being pulled in here, just arch-refactor's post-#812 commits (`#816`, `#817`, BLOB fix, `#818`).

2. **`4dbeab1` — `fix: migrate CLI dispatch from deprecated facades to session API`**

   PR #818 removed the deprecated `Whitenoise::*` facades from the library, but missed `crates/whitenoise-cli/src/cli/dispatch.rs`, which still called 34 of them. Each call has been translated to `wn.require_session(&account.pubkey)?.area().method(...)` form. All 22 stale `#[allow(deprecated)]` attributes also removed.

   Two visibility workarounds (worth flagging for follow-up):
   - `Whitenoise::require_session` is `pub(crate)` — added a small `require_session(wn, pubkey)` helper at the top of `dispatch.rs` that wraps the public `Whitenoise::session(...)` accessor.
   - `ChatList::build_item` is `pub(crate)` — `get_chat_list_item` now scans `active()` + `archived()` for the matching `mls_group_id` instead. Less efficient but stays within the public API.

3. **`a1a903d` — `fix: restore add_members SelfRemove pre-check + hash_ref-scoped backdate`**

[...]

PR #786, opened 2026-04-30
Phase 18c: split per-account tables into per-account DBs

dannym-arx:
marmot

Phase 18c moves six account-scoped tables out of `shared.sqlite` into per-account SQLite files. Each marmot now keeps its own filing cabinet: deletion becomes `fs::remove_file`, joins lose their `WHERE account_pubkey = ?` boilerplate, and per-account schemas drop the redundant `account_pubkey` column where the file is the scope.

## What moved

| Table | Notes |
|---|---|
| `account_settings` | Singleton row per file (`id = 1`) |
| `drafts` | `mls_group_id` is the unique key |
| `published_key_packages` | `event_id` is the unique key |
| `published_events` | Pre-split FK reschema at v13 (`account_id` to `account_pubkey`) |
| `processed_events` | Row-scope split: account-scoped rows (`account_pubkey IS NOT NULL`) move, global rows (`account_pubkey IS NULL`) stay in shared |
| `account_follows` | Collapsed to `(pubkey, created_at, updated_at)`. No FK to users; followed pubkeys stored as plain TEXT |

## How ordering works

Builds on the unified migration timeline from #785: globals and locals share one version space, so the local-copy (vN) to global-drop (vN+k) ordering is automatic. The migration list:

- v13 to v14 (global): pre-split FK reschemas
- v15 to v20 (local): physical copies into per-account DBs
- v21 to v25 (global): `DROP TABLE` on the moved tables in shared
- v26 (global): `DELETE FROM processed_events WHERE account_pubkey IS NOT NULL`

`AccountSession::new` calls `account_db.run_account_migrations(&shared.pool)`, which walks both pools in lockstep. After `restore_sessions` finishes at boot, `Whitenoise::new` re-runs `MIGRATOR.run(&shared, None)` so deferred drops only commit once every on-disk account has copied. Accounts that never log in keep their data safely.

## EventTracker rerouting

`global_published_event(event_id)` used to be a single `WHERE` on shared. Post-split there is no shared index, so `WhitenoiseEventTracker` now holds `Weak<Whitenoise>` (built via `Arc::new_cyclic`) and iterates the active sessions, checking each per-account DB. The number of logged-in accounts is small, so the iteration cost stays trivial. `EphemeralPlane` delegates published-event tracking through the `EventTracker` trait so the session lookup lives in one place.

## API changes worth flagging

[...]

PR #830, opened 2026-05-13
Recover some coverage on account DB wrappers and membership ops

mubarakcoded:
## Summary

The arch-refactor merge dropped CI coverage 94.95% → 89.02%. This PR adds 8 targeted unit tests covering paths the refactor introduced without tests — three new per-account-DB repositories and the lightweight column-flip behavior of `MembershipOpsForGroup::archive` / `unarchive`.

## What's tested

`database/account/push_registrations.rs` (2 tests)
- `upsert_binds_repo_account_pubkey_to_returned_registration` — the wrapper's only added behavior is binding `account_pubkey`; this locks it in.
- `separate_accounts_have_independent_per_account_dbs` — the per-account-DB architectural invariant (post-#786 split); two repos over distinct account DBs must not see each other's rows. Inner `PushRegistration` tests share one pool and can't express this.

`database/account/group_push_tokens.rs` (1 test)
- `upsert_binds_repo_account_pubkey_to_returned_token` — pubkey-binding contract on the wrapper.

`database/account/published_key_packages.rs` (1 test)
- `mark_key_material_deleted_by_id_flips_one_row_only` — contrapositive regression guard for PR #822: the by-id variant must NOT cascade to dual-kind twins. A future refactor that "fixes consistency" by accidentally hash_ref-scoping the by-id variant trips this test. No inner-test analogue.

`accounts_groups.rs` (4 tests)
- `test_archive_sets_archived_at`, `test_archive_is_idempotent`, `test_unarchive_clears_archived_at`, `test_unarchive_on_unarchived_chat_is_no_op` — lightweight column-flip tests for `MembershipOpsForGroup::archive`/`unarchive` via the `get_or_create(None)` path. Distinct from the existing `chat_list.rs` archive tests, which spin up a real MLS group via `create_group(...)` and assert stream emit; these target the wrapper's mutation contract without the MLS setup cost.

## What's deliberately not in this PR

- **Wrapper pure-delegation roundtrip tests.** Pure pass-through `Ok(Inner::fn(...).await?)` calls add coverage but no signal — the inner SQL has its own tests. Per the user rule "tests that really increase coverage, not tests that are not testing anything," these were dropped after review.
- **Hash_ref-scoped twin tests** for `mark_consumed` / `mark_key_material_deleted_by_hash_ref`. Already covered by the inner tests `test_mark_consumed_updates_timestamp_for_twins` and `test_mark_key_material_deleted_by_hash_ref_marks_twins` in `database/published_key_packages.rs`.
- **Archive/unarchive emit-frame tests.** Already covered at the heavyweight level by `test_archive_chat_emits_to_both_stream_channels` in `chat_list.rs`.
- **Schema-shadowing fix.** The account-DB tests currently hand-roll a `DROP TABLE + CREATE TABLE` because `AccountDatabase::new` uses `open_without_migrations` (running the full timeline here would require wiring a shared pool). Filed for a separate follow-up — a `setup_account_db_with_migrations()` helper that the whole `account/*.rs` suite can migrate to.

## Coverage math caveat

Coverage probably lands around 90–92% local lib — clears the post-merge CI baseline of 89.02% but doesn't recover the full pre-merge 94.95%. The remaining gap is structural: arch-refactor genuinely deleted code paths that had coverage, and shallow tests aren't the right fix.

## Verification

- `just check-fmt` ✓
- `just check-clippy` ✓
- `cargo test --lib whitenoise::database::account` — 55 pass
- `cargo test --lib whitenoise::accounts_groups::tests` — all pass including the 4 new archive/unarchive

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive integration scenario and test cases for mute-list (block/unblock/verify) and expanded unit tests across account groups, push registrations, published key packages, group push tokens, and mute-list edge cases.
* **Chores**
  * Updated coverage reporting exclusions to ignore migration and CLI paths in CI and local build tooling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/marmot-protocol/whitenoise-rs/pull/830)


erskingardner (commented):
@mubarakcoded sorry to dump these here directly but since they span lots of different lines it doesn't really make sense to try and put them against a single location: 

> ### Schema drift: hand-rolled DDL duplicates production schema without a single source of truth
> 
> **`src/whitenoise/database/account/push_registrations.rs`** (test `setup()`, lines ~824–854)
> **`src/whitenoise/database/account/group_push_tokens.rs`** (test `setup()`, lines ~668–692)
> **`src/whitenoise/database/account/published_key_packages.rs`** (test `setup()`, lines ~742–762)
> 
> The tests hand-roll `CREATE TABLE` DDL in each `setup()` function. This is the established pattern in the codebase (drafts, follows, mute_list all do the same), and the PR description explicitly calls it out as a known limitation to address in a follow-up. That's fine.
> 
> The concern is that this pattern is spreading: the three new `setup()` blocks are the 4th, 5th, and 6th instances in the `database/account/` subtree. Each one is a silent schema-drift landmine — if a column is added or a constraint is changed in `fresh_account_schema.sql`, the test DDL will silently diverge and the tests will pass against a schema that no longer exists in production.
> 
> The group_push_tokens case is already slightly inconsistent with `fresh_account_schema.sql`: the schema file includes three indexes (`idx_group_push_tokens_group`, `idx_group_push_tokens_group_member`, `idx_group_push_tokens_server`), none of which appear in the test DDL. For the current tests this is harmless, but it's another sign that the gap will grow over time.
> 
> The follow-up `setup_account_db_with_migrations()` helper mentioned in the PR description would eliminate all of this. Since this helper doesn't exist yet, consider at minimum adding a `// Schema must stay in sync with fresh_account_schema.sql` comment in each `setup()` to make the maintenance obligation visible to the next person who edits either side.
> 
> ---
> 
> ### Inline magic byte literals make test isolation implicit rather than explicit
> 
> **`src/whitenoise/accounts_groups.rs`** (new tests, lines ~479, ~507, ~538, ~568, ~602)
> 
> Each new test picks a distinct byte for `GroupId::from_slice(&[N; 32])` — `120`, `121`, `122`, `123`, `145`. Choosing different literals is necessary to prevent cross-test interference when tests share a database (they don't here since each test calls `create_mock_whitenoise()`, which spins up a fresh instance). But the reader has to know that to trust the tests, and the byte values look arbitrary.
> 
> Since each test already creates its own `Whitenoise` instance, the `GroupId` byte values don't need to differ between tests at all — `[1; 32]` would work in all of them without collision. If you keep distinct values, a brief comment (e.g., `// arbitrary — unique per test to aid debugging`) would help readers understand that these values carry no semantic meaning.

erskingardner (approved):

PR #834, opened 2026-05-14
Perf/whitenoise init

### LINKED ISSUE #599: Investigate slow initial startup time (~15s)
## Context

Initial app startup currently feels too slow. The observed startup time is roughly 15 seconds before the app is ready.

This issue is to investigate where that time is being spent inside `whitenoise-rs` initialization and startup work.

## Details

Areas that look worth profiling first:

- `Whitenoise::initialize_whitenoise()` overall startup path
- startup cache sync work such as `sync_message_cache_on_startup()`
- relay / control-plane initialization
- scheduled task startup
- database migrations or other one-time initialization work
- any blocking network calls or waits happening before the UI can proceed

Questions to answer:

- what are the biggest contributors to the ~15 second startup time
- how much of that cost is local CPU / I/O vs network waiting
- which work is truly required before the app becomes usable
- which work can be deferred until after first render / first usable state
- whether there is a measurable difference between first startup, cold startup, and subsequent launches

## Outcome

Produce a breakdown of startup cost and identify concrete follow-up optimizations.

Possible outcomes:

- move non-critical startup work off the critical path
- defer sync or maintenance tasks until after initial readiness
- add instrumentation / timing logs around initialization phases
- split one-time setup from normal launch work
- open targeted follow-up issues once the hotspots are known

## Additional context

There is already app-side DX work around startup behavior in `marmot-protocol/whitenoise`, but this issue is specifically about the Rust core initialization path because that is likely where a large part of the delay lives.

jgmontoya:
Closes https://github.com/marmot-protocol/whitenoise-rs/issues/599

Moves the discovery plane initialization and the setup of subscriptions to be performed in the background and not block initialization.

## Summary by CodeRabbit

* **New Features**
  * WAN latency simulation for benchmarks and new startup-under-latency benchmark command producing JSON results.
  * Benchmark runner scripts and helpers to run scenarios with optional latency and timestamped outputs.

* **Bug Fixes / Reliability**
  * Ensures tracing is flushed on shutdown to avoid lost logs.

* **Chores**
  * Compose profile and service added for latency testing; build tooling recipes updated for latency-aware benchmarks.
* **Tests**
  * Added tests and helpers for configurable default relays.

[Review Change Stack](https://app.coderabbit.ai/change-stack/marmot-protocol/whitenoise-rs/pull/834)

### DIFF
--- a/dev/toxiproxy.json
+++ b/dev/toxiproxy.json
@@ -1,0 +1,14 @@
+[
+  {
+    "name": "nostr-rs-slow",
+    "listen": "0.0.0.0:18080",
+    "upstream": "nostr-rs-relay:8080",
+    "enabled": true
+  },
+  {
+    "name": "strfry-slow",
+    "listen": "0.0.0.0:17777",
+    "upstream": "strfry-nostr-relay:7777",
+    "enabled": true
+  }
+]
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -53,6 +53,26 @@ services:
       setup:
         condition: service_completed_successfully
 
+  # Network-latency simulator used by the WAN-latency benchmark recipe.
+  # Profile-gated so `just docker-up` / `docker-down` ignore it; only the
+  # `benchmark-startup-wan` recipe enables it via `--profile latency`.
+  toxiproxy:
+    image: ghcr.io/shopify/toxiproxy:2.12.0
+    profiles: ["latency"]
+    ports:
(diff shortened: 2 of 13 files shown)


dannym-arx (approved):
I like this, I think this is the solution to the flaky tests.

If they're still flaky after this... then we just gotta fix (or remove) the ones that are

mubarakcoded (approved):
This looks good. Nice!

erskingardner (approved):
🚢

PR #835, opened 2026-05-15
Reuse d_tag slot for kind:30443 publishes to enable NIP-33 replacement

erskingardner:
## Summary

Implements d_tag reuse for MLS key package (kind:30443) publishes so that successive key package rotations replace the previous event on relays per NIP-33, rather than accumulating as distinct addressable events.

The MDK library generates a fresh random `d` tag on every `create_key_package_for_event()` call, but its documentation explicitly states: "Callers SHOULD store and reuse this value when rotating the KeyPackage." Without reuse, relays accumulate stale key packages indefinitely instead of replacing them.

## Key Changes

- **New helper function `apply_d_tag_override()`**: Conditionally swaps the freshly-generated `d` tag from MDK with a previously-stored one, preserving all other event tags (MLS metadata, ciphersuite, extensions, etc.)

- **Database layer (`published_key_packages.rs`)**: Added `find_latest_d_tag()` method to retrieve the most recently recorded `d` tag for a given event kind, ordered by creation time with fallback to ID ordering

- **Repository wrapper**: Exposed `find_latest_d_tag()` through the `PublishedKeyPackagesRepo` interface for session-based access

- **Key package publish flow**: Modified `publish_key_package_for_account_with_signer()` to:
  - Query the database for any previously-stored `d` tag before publishing
  - Apply the override if found (gracefully falls back to fresh d_tag on query failure)
  - Pass canonical tags and d_tag to the relay publish and database record operations

- **Comprehensive test coverage**:
  - Unit tests for `apply_d_tag_override()` covering both stored and fresh d_tag paths
  - Unit tests for `find_latest_d_tag()` with empty table, multiple rows, and kind filtering
  - Integration test verifying that successive key package publishes reuse the same d_tag slot across multiple publish cycles

## Implementation Details

- The override is applied at the event-building stage, before relay publication and database recording, ensuring consistency across all downstream operations
- Error handling is defensive: if the d_tag lookup fails, the system logs a warning and falls back to the fresh MDK-generated d_tag rather than failing the publish
- The database query uses `ORDER BY created_at DESC, id DESC` to handle both normal ordering and edge cases where multiple rows share the same timestamp
- All non-d tags are preserved unchanged to maintain MLS validation and metadata integrity

https://claude.ai/code/session_01VPRo2i9bcezUHXHDqfQvHj

## Summary by CodeRabbit

* **New Features**
  * Canonical (30443) publishes now reuse a persisted replace identifier and support a monotonic publish timestamp so relays replace prior canonical events.
  * Publishing emits paired canonical+legacy events in a single operation; publish paths accept an optional custom timestamp.

* **Bug Fixes**
  * Canonical lifecycle tracking now reliably propagates failures; legacy tracking remains best-effort.

* **Tests**
  * Added unit and integration tests for identifier reuse, monotonic timestamps, override behavior, and canonical+legacy pairing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/marmot-protocol/whitenoise-rs/pull/835)

### DIFF
--- a/src/whitenoise/database/account/published_key_packages.rs
+++ b/src/whitenoise/database/account/published_key_packages.rs
@@ -39,11 +39,6 @@ impl PublishedKeyPackagesRepo {
     }
 
     /// Return the most recently inserted row for a given event kind, if any.
-    ///
-    /// Canonical-slot consumers read both `d_tag` and `created_at` from
-    /// this single row. Folding the lookup into one query avoids the
-    /// implication that the d-tag and the insert timestamp could come
-    /// from two different prior publishes.
     pub async fn find_latest_by_kind(
         &self,
         kind: nostr_sdk::Kind,
--- a/src/whitenoise/database/account/published_key_packages.rs
+++ b/src/whitenoise/database/account/published_key_packages.rs
@@ -38,6 +38,11 @@ impl PublishedKeyPackagesRepo {
         Ok(PublishedKeyPackage::find_by_hash_ref(&self.db, hash_ref).await?)
     }
 
+    /// Return the most recently recorded `d` tag for a given event kind, if any.
+    pub async fn find_latest_d_tag(&self, kind: nostr_sdk::Kind) -> Result<Option<String>> {
+        Ok(PublishedKeyPackage::find_latest_d_tag(&self.db, kind).await?)
+    }
+
     /// Mark a published key package as consumed (used by a Welcome).
     pub async fn mark_consumed(&self, event_id: &str) -> Result<bool> {
         Ok(PublishedKeyPackage::mark_consumed(&self.db, event_id).await?)
--- a/src/whitenoise/database/account/published_key_packages.rs
+++ b/src/whitenoise/database/account/published_key_packages.rs
@@ -43,6 +43,11 @@ impl PublishedKeyPackagesRepo {
         Ok(PublishedKeyPackage::find_latest_d_tag(&self.db, kind).await?)
     }
 
+    /// Return the maximum `created_at` recorded for a given event kind, if any.
+    pub async fn find_max_created_at(&self, kind: nostr_sdk::Kind) -> Result<Option<i64>> {
+        Ok(PublishedKeyPackage::find_max_created_at(&self.db, kind).await?)
+    }
+
     /// Mark a published key package as consumed (used by a Welcome).
     pub async fn mark_consumed(&self, event_id: &str) -> Result<bool> {
         Ok(PublishedKeyPackage::mark_consumed(&self.db, event_id).await?)
--- a/src/whitenoise/database/account/published_key_packages.rs
+++ b/src/whitenoise/database/account/published_key_packages.rs
@@ -38,14 +38,17 @@ impl PublishedKeyPackagesRepo {
         Ok(PublishedKeyPackage::find_by_hash_ref(&self.db, hash_ref).await?)
     }
 
-    /// Return the most recently recorded `d` tag for a given event kind, if any.
-    pub async fn find_latest_d_tag(&self, kind: nostr_sdk::Kind) -> Result<Option<String>> {
-        Ok(PublishedKeyPackage::find_latest_d_tag(&self.db, kind).await?)
-    }
-
-    /// Return the maximum `created_at` recorded for a given event kind, if any.
-    pub async fn find_max_created_at(&self, kind: nostr_sdk::Kind) -> Result<Option<i64>> {
-        Ok(PublishedKeyPackage::find_max_created_at(&self.db, kind).await?)
+    /// Return the most recently inserted row for a given event kind, if any.
+    ///
+    /// Canonical-slot consumers read both `d_tag` and `created_at` from
+    /// this single row. Folding the lookup into one query avoids the
+    /// implication that the d-tag and the insert timestamp could come
+    /// from two different prior publishes.
+    pub async fn find_latest_by_kind(
+        &self,
+        kind: nostr_sdk::Kind,
+    ) -> Result<Option<PublishedKeyPackage>> {
+        Ok(PublishedKeyPackage::find_latest_by_kind(&self.db, kind).await?)
     }
 
     /// Mark a published key package as consumed (used by a Welcome).
--- a/src/whitenoise/database/published_key_packages.rs
+++ b/src/whitenoise/database/published_key_packages.rs
@@ -107,22 +107,37 @@ impl PublishedKeyPackage {
         Ok(row)
     }
 
-    /// Returns the most recently recorded `d` tag for a given event kind, if any.
+    /// Returns the most recently inserted row for a given event kind, if any.
     ///
-    /// Used by the kind:30443 publish path to reuse a single addressable-event
-    /// slot per NIP-33. MDK generates a fresh random `d` tag inside
-    /// `create_key_package_for_event` every time it's called, and the MDK
-    /// docs explicitly say "Callers SHOULD store and reuse this value when
-    /// rotating the KeyPackage." Without reuse, every publish lands as a
-    /// distinct event on relays instead of replacing the previous one.
+    /// Single query that covers both NIP-33 canonical-slot needs:
+    /// - `d_tag` — reuse the previously-published addressable-slot
+    ///   identifier so the next publish replaces the prior canonical event
+    ///   on relays instead of landing in a fresh slot. MDK generates a
+    ///   fresh random `d` tag every call to `create_key_package_for_event`
+    ///   and the MDK docs say "Callers SHOULD store and reuse this value
+    ///   when rotating the KeyPackage."
+    /// - `created_at` (the row's SQLite `unixepoch()` insert timestamp,
+    ///   not the Nostr event's `created_at`) — use as a lower bound when
+    ///   computing a strictly-monotonic event `created_at` for the next
+    ///   publish. NIP-01 says relays keep the lowest-id event on a tie,
+    ///   so reusing the d-tag alone isn't enough under same-second
+    ///   publishes. The row insert timestamp is always ≥ the event's
+    ///   `created_at` (INSERT runs after the publish round-trip), which
+    ///   makes it a sound upper bound for the prior event time.
+    ///
+    /// Folding both lookups into one query saves a DB roundtrip and makes
+    /// it explicit at the call site that the d-tag and the insert time
+    /// come from the *same* prior publish — not two separate rows.
     #[perf_instrument("db::published_key_packages")]
-    pub(crate) async fn find_latest_d_tag(
+    pub(crate) async fn find_latest_by_kind(
         db: &AccountDatabase,
         kind: Kind,
-    ) -> Result<Option<String>, DatabaseError> {
-        let row: Option<(String,)> = sqlx::query_as(
-            "SELECT d_tag FROM published_key_packages
-             WHERE kind = ? AND d_tag IS NOT NULL
+    ) -> Result<Option<Self>, DatabaseError> {
+        let row = sqlx::query_as::<_, Self>(
+            "SELECT id, key_package_hash_ref, event_id, kind, d_tag,
+                    consumed_at, key_material_deleted, created_at
+             FROM published_key_packages
+             WHERE kind = ?
              ORDER BY created_at DESC, id DESC
              LIMIT 1",
         )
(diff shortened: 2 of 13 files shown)


src/whitenoise/session/key_packages.rs
```diff
@@ -98,20 +93,94 @@ impl<'a> KeyPackageOps<'a> {
...
+            .repos
+            .published_key_packages
+            .find_latest_d_tag(MLS_KEY_PACKAGE_KIND)
+            .await?;
+        let prev_canonical_max = self
+            .session
+            .repos
+            .published_key_packages
+            .find_max_created_at(MLS_KEY_PACKAGE_KIND)
+            .await?;
```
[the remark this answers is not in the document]
erskingardner: Declining this — it directly contradicts your earlier review on this same PR ([discussion_r3247043504](https://github.com/marmot-protocol/whitenoise-rs/pull/835#discussion_r3247043504)) on the parallel call site in `Whitenoise::publish_key_package_pair_to_relays`, which asked for the opposite ("Fail the publish when loading the canonical `d` tag fails. Falling back to a fresh MDK-generated `d` here breaks the exact invariant this PR is adding"). The reasoning was correct: a transient DB read that flips `stored_d_tag` to `None` causes the publish to land in a brand-new NIP-33 slot and leave the existing canonical event stale — re-introducing the bug.

The same logic applies to `prev_canonical_max`: if we fall back to the wall clock and the prior event happens to be in the same wall-clock second, the new event can lose the NIP-01 lowest-event-id tiebreaker.

Per-account SQLite read failures on a local file are rare, and the scheduler retries on its next tick — so fail-fast preserves the invariant without a real availability cost.

The companion comment about canonical *tracking* failures is valid and addressed in ff7c779 — that one bubbles `?` for the canonical write while keeping legacy tracking best-effort.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/whitenoise/database/published_key_packages.rs
```diff
@@ -107,6 +107,56 @@ impl PublishedKeyPackage {
...
+        kind: Kind,
+    ) -> Result<Option<i64>, DatabaseError> {
+        let row: Option<(Option<i64>,)> =
+            sqlx::query_as("SELECT MAX(created_at) FROM published_key_packages WHERE kind = ?")
+                .bind(i64::from(kind.as_u16()))
+                .fetch_optional(&db.inner.pool)
+                .await?;
+
+        Ok(row.and_then(|(v,)| v))
+    }
```
jgmontoya, line 145: I think this two queries are basically the same: get some information (d_tag, created_at) from the latest event of a given kind.

I think a better approach would be to replace both of this with a single method `get_latest_by_kind` or something that returns the whole object. Then we just take both the `d_tag` and `created_at` values from it at the point of calling. Doing it this way we'll save one db query.

src/whitenoise/key_packages.rs
```diff
@@ -256,6 +256,30 @@ pub(crate) fn validate_fetched_member_key_package(event: &Event, pk: &PublicKey)
...
+            if floor > now.as_secs() {
+                Timestamp::from_secs(floor)
+            } else {
+                now
+            }
+        }
+        _ => now,
+    }
+}
+
```
jgmontoya, line 285: I think this can be simplified a bit:

```rs
pub(crate) fn monotonic_canonical_created_at(prev_max: Option<i64>) -> Timestamp {
    let now = Timestamp::now();
    match prev_max {
        Some(prev) if prev >= 0 => {
            let floor = (prev as u64).saturating_add(1);
            Timestamp::from_secs(floor.max(now.as_secs()))
        }
        _ => now,
    }
}
```

src/whitenoise/key_packages.rs
```diff
@@ -349,7 +398,9 @@ impl Whitenoise {
         account: &Account,
         relays: &[Relay],
     ) -> Result<()> {
-        let key_package_data = self.encoded_key_package(account, relays).await?;
+        let (key_package_data, canonical_created_at) = self
+            .prepare_canonical_publish_inputs(account, relays)
+            .await?;
```
jgmontoya: My agent flagged the following:

> ```rust
> pub(crate) async fn create_key_package_and_background_publish(...) -> Result<()> {
>     let (key_package_data, canonical_created_at) =
>         self.prepare_canonical_publish_inputs(account, relays).await?;
>     ...
>     tokio::spawn(async move {
>         if let Err(e) = wn.publish_key_package_pair_to_relays(...).await {
> ```
> 
> Setup order in `src/whitenoise/accounts/setup.rs:96-104`:
> 
> 1. `insert_account_session(account)` (line 96) — session including `key_package_publish_lock` exists
> 2. `setup_subscriptions(account, relays)` (line 98) — giftwrap and other event handlers are wired
> 3. `setup_key_package(account, true, relays)` (line 104) → `create_key_package_and_background_publish` (line 196)
> 
> By the time the spawned task runs, a giftwrap delivery (`src/whitenoise/event_processor/event_handlers/handle_giftwrap.rs:430` calls `session.key_packages().publish()`) can fire and race the spawn. Both paths read `find_latest_d_tag → None` and `find_max_created_at → None`, MDK generates **two different fresh d-tags**, both publish, both `INSERT OR IGNORE` succeed (no unique constraint on `(kind, d_tag)` — only on `event_id`). Slot drift, which is the exact bug this branch fixes.
> 
> The race window is small (giftwrap arriving in the seconds between `setup_subscriptions` and the spawned task completing) but the entire point of adding the lock is to make it impossible by construction.
> 
> **Fix**: take the lock inside the spawned closure (or before spawning, with `OwnedMutexGuard` if you need to hand ownership to the spawned task). Or document explicitly why this path is exempt — the current comment ("intentionally left unlocked: it runs during fresh-account creation, before any other trigger can fire") doesn't survive scrutiny because subscriptions are already active.

src/whitenoise/database/published_key_packages.rs
jgmontoya, line 145: Actually, reading this again I think this is returning the created_at from the db, that is the time when this got inserted in the db, not the created_at of the nostr event. I guess we would need to query the processed events table (is this even in there? not sure) to get the `event_created_at`

src/whitenoise/key_packages.rs
```diff
@@ -397,11 +455,61 @@ impl Whitenoise {
...
+    /// inputs and just publishes + tracks.
     #[perf_instrument("key_packages")]
     async fn publish_key_package_pair_to_relays(
         &self,
         account: &Account,
         key_package_data: &KeyPackageEventData,
+        canonical_created_at: Timestamp,
         relay_urls: &[RelayUrl],
         signer: std::sync::Arc<dyn NostrSigner>,
     ) -> Result<()> {
```
jgmontoya, line 562: I'm finding it a bit confusing to have this method here and `KeyPackageOps::publish_pair` on `src/whitenoise/session/key_packages.rs:182-239`.

Why do we need both? When should we use each? It seems like the session scoped one should be the one to use and perhaps we should get rid of this one?

erskingardner: Both paths exist because of the external-signer login flow:

1. `Whitenoise::insert_external_signer(pubkey, signer)` — the signer goes into `Whitenoise.shared.external_signers` (a HashMap on shared state).
2. `Whitenoise::activate_account_without_publishing` → `AccountSession::from_account` — for external accounts (`!account.has_local_key()`), the session is created with `signer: None`.
3. `Whitenoise::publish_key_package_for_account_with_signer(account, signer)` — uses the explicit signer passed in.

If step 3 routed through `KeyPackageOps::publish_pair`, it would hit `require_signer()` which would return `SignerUnavailable` because the session's signer slot is still empty — the external signer lives in the `external_signers` HashMap, not on the session yet.

Real consolidation means changing `AccountSession::from_account` (or `insert_account_session`) to consult the `external_signers` HashMap and populate `session.signer` for external accounts too — a session-signer-lifecycle refactor that touches multiple flows (login_external_signer, register_external_signer, etc.). Worth doing but out of scope for this PR.

Keeping it in mind as a follow-up. Will leave a code comment on `publish_key_package_pair_to_relays` pointing at this thread.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/whitenoise/database/published_key_packages.rs
erskingardner: Done in `ccd3aef`. Replaced `find_latest_d_tag` + `find_max_created_at` with a single `find_latest_by_kind` that returns the whole `PublishedKeyPackage` row; both call sites (`Whitenoise::prepare_canonical_publish_inputs` and `KeyPackageOps::prepare_canonical_publish_inputs`) now take `d_tag` and `created_at` from that one row. One DB roundtrip instead of two, and it's now explicit that both values come from the same prior publish.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/whitenoise/key_packages.rs
erskingardner: Applied verbatim in `ccd3aef` — `floor.max(now.as_secs())` with the `prev as u64` cast guarded by `prev >= 0`. Existing unit tests (`test_monotonic_canonical_created_at_*`) still pass unchanged.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/whitenoise/database/published_key_packages.rs
erskingardner: Correct — `find_latest_by_kind` returns the row's `created_at`, which is the SQLite `unixepoch()` at INSERT time, not the Nostr event's `created_at`. I deliberately kept it that way rather than reaching into the processed-events table, because the monotonic invariant still holds:

1. `EventBuilder` sets `event.created_at = T1`.
2. The event is published to relays (network round-trip).
3. *Then* the row is inserted → `row.created_at = T2` via `unixepoch()`.

So `T2 ≥ T1` always — the row insert timestamp is a sound *upper bound* on the prior event's `created_at`. `monotonic_canonical_created_at` sets the next event's `created_at` to `max(now, T2 + 1)`, which is therefore strictly greater than the previous event's `T1`. The NIP-01 same-second tiebreaker is avoided.

Querying `processed_events` for the real `event_created_at` would be more semantically precise but adds a join/lookup for zero behavioral gain — the invariant we care about (strictly increasing canonical event `created_at`) is already guaranteed. I expanded the rustdoc on both `find_latest_by_kind` and `monotonic_canonical_created_at` in `ccd3aef` to spell this reasoning out so it's not a footgun for the next reader. If you'd rather track the true event timestamp anyway (e.g. for future precision), that'd be a `published_key_packages` schema/migration change I'd suggest as a follow-up rather than folding into this PR.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/whitenoise/database/account/published_key_packages.rs
```diff
@@ -38,6 +38,19 @@ impl PublishedKeyPackagesRepo {
         Ok(PublishedKeyPackage::find_by_hash_ref(&self.db, hash_ref).await?)
     }
 
+    /// Return the most recently inserted row for a given event kind, if any.
+    ///
+    /// Canonical-slot consumers read both `d_tag` and `created_at` from
+    /// this single row. Folding the lookup into one query avoids the
+    /// implication that the d-tag and the insert timestamp could come
+    /// from two different prior publishes.
```
jgmontoya: I would remove this comment

src/whitenoise/database/published_key_packages.rs
```diff
@@ -107,6 +107,47 @@ impl PublishedKeyPackage {
...
+    ///   computing a strictly-monotonic event `created_at` for the next
+    ///   publish. NIP-01 says relays keep the lowest-id event on a tie,
+    ///   so reusing the d-tag alone isn't enough under same-second
+    ///   publishes. The row insert timestamp is always ≥ the event's
+    ///   `created_at` (INSERT runs after the publish round-trip), which
+    ///   makes it a sound upper bound for the prior event time.
+    ///
+    /// Folding both lookups into one query saves a DB roundtrip and makes
+    /// it explicit at the call site that the d-tag and the insert time
+    /// come from the *same* prior publish — not two separate rows.
```
jgmontoya: I would remove this comment

src/whitenoise/database/published_key_packages.rs
erskingardner: Trimmed in `88df5e7`. The thin `PublishedKeyPackagesRepo::find_latest_by_kind` wrapper is now a one-line doc matching its siblings, and the "Folding both lookups into one query…" refactor-history paragraph is removed from the underlying `PublishedKeyPackage::find_latest_by_kind`. The behavioral docs there (what `d_tag`/`created_at` are used for, the row-insert vs event-time caveat) stay.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/whitenoise/key_packages.rs
```diff
@@ -322,86 +365,198 @@ impl Whitenoise {
...
+        // Single query: the d-tag and the insert timestamp both come from the
+        // most recent canonical row. Folding the two lookups avoids the
+        // implication that they could disagree.
+        let latest_canonical = session
+            .repos
+            .published_key_packages
+            .find_latest_by_kind(MLS_KEY_PACKAGE_KIND)
+            .await?;
+        let stored_d_tag = latest_canonical.as_ref().and_then(|r| r.d_tag.as_deref());
+        let prev_canonical_max = latest_canonical.as_ref().map(|r| r.created_at);
```
mubarakcoded, line 522: Small concern: this floor comes from the DB row `created_at`, which is the SQLite insert time rather than the canonical event `created_at`. If a publish bumps the event timestamp ahead of wall-clock time, a fast follow-up publish can still read the lower insert timestamp and produce the same kind:30443 `created_at` again. Should we persist/use the actual published event timestamp for this monotonic floor?

erskingardner: I'm pretty sure this is the same issue that @jgmontoya flagged in another comment above. Response here: https://github.com/marmot-protocol/whitenoise-rs/pull/835#discussion_r3255326383

I think this is fine the way it is.

mubarakcoded: I have seen it. Thanks

mubarakcoded (approved):
LGTM!

PR #840, opened 2026-05-17
fix: bound delete all data shutdown

mubarakcoded:
## Summary

This bounds the `Whitenoise::delete_all_data()` shutdown path so a destructive local reset cannot hang indefinitely while waiting on relay, scheduler, or background work.

The app-side delete flow was still able to spin for minutes because `delete_all_data()` first performed a graceful shutdown. If any scheduled/background task or relay-control shutdown path stalled, local data deletion never reached the filesystem wipe.

## What changed

- Replaces the unbounded delete shutdown path with a shared `DELETE_SHUTDOWN_GRACE` deadline.
- Gives event processing, scheduler tasks, background tasks, inbox deactivation, and relay-control shutdown a bounded grace window.
- Aborts scheduler/background task handles that do not stop within the shared deadline.
- Continues with local data deletion after best-effort network/session shutdown timeouts.
- Adds regression coverage for:
  - full `delete_all_data()` with a stuck registered background task
  - the lower-level abort-after-deadline shutdown helper

## Why

For delete-all-data, local wipe must be prioritized over perfect graceful network cleanup. A stuck network/task shutdown should not prevent the user from clearing local account data.

## Testing

- `cargo test test_delete_all_data_aborts_stuck_background_task`
- `cargo test test_delete_shutdown_aborts_task_after_shared_deadline`
- `cargo clippy -- -D warnings`

## Caveat

This PR only covers the `whitenoise-rs` side of the delete flow: making teardown bounded so local data deletion is not blocked by stuck background or relay work.

The Flutter-side lifecycle handling is covered in the companion app PR, which handles reinitializing after delete and recovering cleanly if the reset is interrupted.

## Summary by CodeRabbit

* **Bug Fixes**
  * Delete now uses a shared, deadline-based shutdown so deletion won’t hang; scheduled and background tasks are bounded, timed out with warnings, and aborted if they exceed the grace period.

* **Tests**
  * Added coverage verifying stuck background tasks are terminated during delete and that the shared-deadline logic prevents hangs.
  * Added a test drop-tracker utility to assert task teardown.

[Review Change Stack](https://app.coderabbit.ai/change-stack/marmot-protocol/whitenoise-rs/pull/840?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


src/whitenoise/mod.rs
```diff
@@ -4482,6 +4680,14 @@ mod tests {
     mod scheduler_lifecycle_tests {
         use super::*;
 
+        struct DropFlag(std::sync::Arc<std::sync::atomic::AtomicBool>);
```
erskingardner: The diff introduces `DropFlag` twice: once as a function-local struct inside `test_delete_all_data_aborts_stuck_background_task`, and once at module level in `scheduler_lifecycle_tests`. Since they are in different modules they compile fine, but the duplication is unnecessary. Move the struct to a shared test-utils location (the existing `pub mod test_utils` block or a sibling `mod test_helpers` within the top-level `#[cfg(test)]` block), and import it in both test modules. This is consistent with how other test fixtures (`create_mock_whitenoise`, `create_test_keys`, etc.) are shared.

src/whitenoise/mod.rs
```diff
@@ -1413,6 +1495,67 @@ impl Whitenoise {
         }
     }
 
+    async fn wait_for_pending_background_tasks_for_delete(&self, deadline: Instant) {
```
erskingardner, line 1569: The outer `loop` mirrors the original `wait_for_pending_background_tasks`, which is correct there because event processing is still live and tasks can re-enqueue themselves. Here, event processing has been shut down before this is called (via `shutdown_event_processing`), but **any background task that itself calls `spawn_background` will push a new handle into `self.background_handles` while the drained batch is being awaited**, causing the loop to iterate again with the new handle.

That is the expected behavior for the unbounded version — its doc comment explicitly calls this out. But in the delete path, `await_or_abort_tasks` will abort stuck tasks rather than waiting indefinitely, and the deadline tracks elapsed time globally, so the worst case here is bounded by the deadline plus one extra `await_or_abort_tasks` call per nested spawn generation. That is probably fine in practice, but it is subtly different from what the function name implies and the behavior is not documented. Add a comment explaining the loop rationale (same as the original's doc comment) so the next reader does not wonder why this isn't a simple `await_or_abort_tasks` call.

Additionally, once `remaining_delete_grace(deadline).is_zero()`, `await_or_abort_tasks` aborts every remaining handle immediately but the outer loop still needs a pass to discover the registry is empty. That is at most one extra mutex lock, not a correctness problem, but it reads awkwardly. Consider breaking early if `remaining_delete_grace(deadline).is_zero()` at the top of the loop.

src/whitenoise/mod.rs
```diff
@@ -1340,6 +1371,57 @@ impl Whitenoise {
...
+            .await;
+
+        tracing::info!(
+            target: "whitenoise::delete_all_data",
+            "Bounded shutdown before local data wipe complete"
+        );
+        Ok(())
+    }
+
+    async fn shutdown_scheduled_tasks_for_delete(&self, deadline: Instant) {
```
erskingardner, line 1475: should this be instrumented?

src/whitenoise/mod.rs
erskingardner, line 1569: should this be instrumented?

erskingardner (changes requested):

src/whitenoise/mod.rs
mubarakcoded: I added a comment explaining why the delete path keeps the same drain-loop shape. Also, I did not add an early break when the deadline is zero because that could leave a newly registered handle sitting in the registry. Instead, the next pass drains it, and `await_or_abort_tasks` aborts it immediately.

erskingardner (approved):

PR #841, opened 2026-05-19
Per-account gift-wrap decrypt throttle with deferral semantics

mubarakcoded:
## Summary

Adds a per-account token bucket gating NIP-44 decryption of inbound gift-wrap (`kind:1059`) events, with three-way result classification so each event class has the right outcome.

## Motivation

Inbound `kind:1059` events have no sender authentication discoverable before NIP-44 decryption — the recipient must perform secp256k1 ECDH and AEAD decrypt on every event addressed to them, and the KeyPackage reference that would identify the inviter sits inside the encrypted rumor. Without a budget, anyone who knows a recipient's pubkey can force unbounded per-event ECDH work on the recipient's device.

This change bounds that work at the client without changing the protocol.

## What changed

**Throttle.** `GiftwrapThrottle` is a per-account token bucket on `AccountSession` (capacity 60, refill 2/sec). Acquired immediately before `extract_rumor` and after the signer-availability check, so unavailable signers don't drain the budget.

**Single decrypt per event.** Sync-watermark advancement (`update_last_synced_max`) is now done inside `handle_giftwrap` using the rumor timestamp returned by the same `extract_rumor` call, removing the previous post-success second decrypt of the same event.

**Three-way outcome classification.** `extract_rumor` failures are mapped at the source:

| Class | Detection | Outer outcome |
|---|---|---|
| Throttled | bucket empty | `Err(GiftwrapThrottled)` → no track, no retry (relay replay redelivers after refill) |
| External-signer decrypt failed | `nip59::Error::Signer` with non-`Keys` backend | `Err(GiftwrapDeferred)` → no track, no retry (relay replay redelivers after signer recovers) |
| Terminal (malformed payload / local Keys decrypt failed) | any other `Err` | `Ok(())` + warn log → tracked, deduped against future redelivery |
| Success | — | `Ok(())` → tracked, watermark advanced |

The deferral cases deliberately avoid the in-process retry queue, which would re-acquire a token on each of up to 10 retries (`RetryInfo::max_attempts`).

## Tests

Five new unit tests in `giftwrap_throttle::tests` cover the bucket math (trickle, flood, refill cap, sustained-flood throughput, mutex-poison fail-open). Three new integration tests in `account_event_processor::tests` pin the contracts at the boundary:

- `test_throttled_giftwrap_is_not_recorded_as_processed` — exhausts the bucket, sends a valid `kind:1059`, asserts the event is NOT in `already_processed_account_event`.
- `test_undecryptable_giftwrap_is_terminal_not_retried` — sends garbage ciphertext with a local Keys signer, asserts the event IS marked processed AND exactly one token (not eleven) was consumed.
- `test_external_signer_decrypt_failure_is_deferred_not_terminal` — installs a `FailingExternalSigner` (`SignerBackend::Custom`, every `nip44_decrypt` fails), asserts the event is NOT marked processed AND exactly one token was consumed.

## Test plan

- [x] `just precommit-quick` green locally (verified)
- [ ] Manual sanity on physical device after merge: confirm no regression in normal welcome flow.

## Summary by CodeRabbit

* **New Features**
  * Added per-account rate limiting for encrypted message decryption to prevent resource exhaustion.

* **Bug Fixes**
  * Improved error handling to distinguish between permanent decryption failures and temporary signer unavailability.
  * Transient signer backend issues now trigger retry logic instead of marking messages as permanently failed.
  * Fixed synchronization state tracking for encrypted messages.

[Review Change Stack](https://app.coderabbit.ai/change-stack/marmot-protocol/whitenoise-rs/pull/841?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/whitenoise/error.rs
+++ b/src/whitenoise/error.rs
@@ -86,6 +86,12 @@ pub enum WhitenoiseError {
     #[error("Signer unavailable for account {0} — external signer not yet registered")]
     SignerUnavailable(PublicKey),
 
+    #[error("Inbound gift-wrap dropped: per-account decryption budget exhausted")]
+    GiftwrapThrottled,
+
+    #[error("Inbound gift-wrap deferred: signer backend transiently unavailable")]
+    GiftwrapDeferred,
+
     #[error("MDK error: {0}")]
     MdkCoreError(#[from] mdk_core::Error),
 
--- a/src/whitenoise/event_processor/account_event_processor.rs
+++ b/src/whitenoise/event_processor/account_event_processor.rs
@@ -143,57 +143,25 @@ impl Whitenoise {
                             );
                         }
                     }
-                    Kind::GiftWrap => {
-                        // Use rumor timestamp for advancement per NIP-59
-                        match self
-                            .extract_rumor_timestamp_for_advancement(&event, &session)
-                            .await
-                        {
-                            Ok(Some(rumor_timestamp)) => {
-                                let safe_timestamp = cap_timestamp_to_now(rumor_timestamp);
-                                let created_ms = (safe_timestamp.as_secs() as i64) * 1000;
-
-                                if let Err(e) = Account::update_last_synced_max(
-                                    &account.pubkey,
-                                    created_ms,
-                                    &self.shared.database,
-                                )
-                                .await
-                                {
-                                    tracing::warn!(
-                                        target: "whitenoise::event_processor::process_account_event",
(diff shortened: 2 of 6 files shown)


mubarakcoded:
Closing this based on the team discussion.
The throttle does cap client-side decrypt work, but I agree with the feedback that it does not really prevent or meaningfully mitigate the underlying giftwrap DoS shape, since the client still has to unwrap before it knows what is inside.

PR #843, opened 2026-05-20
Add foreground resume catch-up and snapshot refresh handling

erskingardner:
## Summary
- Added a new `Whitenoise::resume_after_background()` primitive for app foreground resume and notification tap handoff.
- The method refreshes subscriptions with bounded replay semantics, logs and skips non-catastrophic per-account failures, and replays current DB snapshots to active chat-list and message streams where possible.
- Added companion snapshot APIs for exact UI replacement: `fetch_chat_list_snapshot()` and `fetch_archived_chat_list_snapshot()`.
- Introduced `SnapshotRefresh` update triggers so Flutter can upsert refreshed rows instead of treating them as brand-new items.

## Testing
- Added focused unit coverage for idempotent resume, partial failure tolerance, snapshot rereads, and replay delivery to active chat-list/message subscribers.
- `just test-unit` passed.
- `just check` passed.
- `just coverage` was attempted, but it failed in the instrumented test harness after local relay containers became unreachable, producing broad relay-connection failures unrelated to the new foreground-resume path.

## Summary by CodeRabbit

* **New Features**
  * Foreground resume now refreshes subscriptions and replays chat-list snapshots to active subscribers.
  * Added snapshot fetch endpoints for active and archived chat lists.
  * Ability to enumerate currently-active subscription keys/pubkeys.

* **Improvements**
  * Preserve existing account state during same-spec subscription updates to avoid unnecessary removals.
  * Option to refresh account subscriptions without full deactivation.

* **Tests**
  * Added tests covering resume, snapshot replay, subscription enumeration, and related edge cases.

[Review Change Stack](https://app.coderabbit.ai/change-stack/marmot-protocol/whitenoise-rs/pull/843?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/whitenoise/foreground_catchup.rs
+++ b/src/whitenoise/foreground_catchup.rs
@@ -44,7 +44,7 @@ impl Whitenoise {
     #[perf_instrument("whitenoise")]
     pub async fn resume_after_background(&self) -> Result<()> {
         let accounts = self.force_foreground_subscription_catch_up().await?;
-        self.replay_foreground_stream_snapshots(&accounts).await;
+        self.replay_chat_list_stream_snapshots(&accounts).await;
         Ok(())
     }
 
@@ -77,9 +77,14 @@ impl Whitenoise {
     }
 
     fn foreground_catch_up_should_abort(error: &WhitenoiseError) -> bool {
+        // Relay/network/subscription errors are treated as transient catch-up
+        // failures so foreground resume can still refresh other accounts and
+        // replay local DB snapshots. Process-wide storage/config/init failures
+        // mean the app cannot reliably read shared state, so return them.
         matches!(
             error,
-            WhitenoiseError::Database(_)
+            WhitenoiseError::Initialization
+                | WhitenoiseError::Database(_)
                 | WhitenoiseError::SqlxError(_)
                 | WhitenoiseError::MdkSqliteStorage(_)
                 | WhitenoiseError::Filesystem(_)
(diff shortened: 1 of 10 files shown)


erskingardner:
@coderabbitai full review

erskingardner:
@claude please review this PR

src/whitenoise/foreground_catchup.rs
```diff
@@ -0,0 +1,381 @@
...
+            tracing::warn!(
+                target: "whitenoise::foreground_catchup",
+                "Foreground discovery subscription catch-up failed, continuing with accounts: {}",
+                error
+            );
+        }
+
+        let accounts = Account::all(&self.shared.database).await?;
+        for account in &accounts {
+            if let Err(error) = self.refresh_account_subscriptions(account).await {
```
mubarakcoded: I think this best-effort path can still cause a healthy account to become worse. `refresh_account_subscriptions()` tears down the current inbox/group subscriptions before activating replacements, so a transient activation failure during foreground resume leaves that account unsubscribed, even though we only log and continue here. Can we use a foreground-specific refresh that keeps the old subscription state until the replacement is active, or otherwise restore/retry before treating this as a skipped account?

src/whitenoise/foreground_catchup.rs
```diff
@@ -0,0 +1,381 @@
...
+                            error
+                        );
+                    }
+                }
+            }
+        }
+    }
+
+    async fn replay_message_stream_snapshots(&self, accounts: &[Account]) {
+        for group_id in self.shared.message_stream_manager.subscribed_group_ids() {
```
mubarakcoded: This replay is keyed only by `group_id`, but the snapshot read is account-specific. If two local accounts are in the same group, this picks the first account that can read the group and broadcasts that account's message view to every subscriber for the group. That can ignore the subscribed account's `chat_cleared_at`/per-account projection. Can the message stream subscribers be keyed by `(account_pubkey, group_id)`, or can we track the subscribing account for each replay?

PR #845, opened 2026-05-20
[codex] Add one-time key package relay cleanup

erskingardner:
## Summary

- add a one-time per-account key package relay cleanup marker
- wait through the invite quiet period, delete relay-side legacy/canonical key package events, then republish one legacy and one kind:30443 package
- broaden cleanup relay scope to key-package relays, NIP-65 relays, and default account relays while tolerating unreachable relays via the existing ephemeral relay plane

## Validation

- cargo test --features cli --lib whitenoise::session::key_packages::tests
- just precommit

## Summary by CodeRabbit

* **New Features**
  * Automatic key-package relay cleanup after startup and on resume with grace windows, retries, verification, and idempotent completion marking.
  * New check to detect recently consumed key packages to defer cleanup when activity is recent.

* **Chores**
  * Per-account tracking for one-time maintenance task completion and accompanying database migration.

* **Tests**
  * Async tests for scheduling, deferral, task-marking idempotency, migration, and consumption/cleanup behaviors.

[Review Change Stack](https://app.coderabbit.ai/change-stack/marmot-protocol/whitenoise-rs/pull/845?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/whitenoise/database/account/maintenance_tasks.rs
+++ b/src/whitenoise/database/account/maintenance_tasks.rs
@@ -33,14 +33,11 @@ impl MaintenanceTasksRepo {
     }
 
     /// Mark `name` complete for this account.
-    ///
-    /// This is intentionally an upsert so retrying the same successful
-    /// maintenance task remains idempotent.
     pub async fn mark_completed(&self, name: &str) -> Result<()> {
         sqlx::query(
             "INSERT INTO account_maintenance_tasks (name, completed_at)
              VALUES (?, unixepoch())
-             ON CONFLICT(name) DO UPDATE SET completed_at = excluded.completed_at",
+             ON CONFLICT(name) DO NOTHING",
         )
         .bind(name)
         .execute(&self.db.inner.pool)
@@ -79,7 +76,8 @@ mod tests {
             .unwrap();
         sqlx::query(
             "CREATE TABLE account_maintenance_tasks (
-                name TEXT PRIMARY KEY,
+                id INTEGER PRIMARY KEY AUTOINCREMENT,
+                name TEXT NOT NULL UNIQUE,
                 completed_at INTEGER NOT NULL
             )",
         )
@@ -104,6 +102,18 @@ mod tests {
         let (repo, _dir) = setup().await;
 
         repo.mark_completed("cleanup").await.unwrap();
+        sqlx::query(
+            "UPDATE account_maintenance_tasks SET completed_at = 123 WHERE name = 'cleanup'",
+        )
+        .execute(&repo.db.inner.pool)
+        .await
+        .unwrap();
+        let first_completed_at: i64 = sqlx::query_scalar(
+            "SELECT completed_at FROM account_maintenance_tasks WHERE name = 'cleanup'",
+        )
+        .fetch_one(&repo.db.inner.pool)
+        .await
+        .unwrap();
         repo.mark_completed("cleanup").await.unwrap();
 
         let completed_count: i64 = sqlx::query_scalar(
@@ -113,5 +123,12 @@ mod tests {
         .await
         .unwrap();
         assert_eq!(completed_count, 1);
+        let second_completed_at: i64 = sqlx::query_scalar(
+            "SELECT completed_at FROM account_maintenance_tasks WHERE name = 'cleanup'",
+        )
+        .fetch_one(&repo.db.inner.pool)
+        .await
+        .unwrap();
+        assert_eq!(second_completed_at, first_completed_at);
     }
 }
--- a/src/whitenoise/database/account/maintenance_tasks.rs
+++ b/src/whitenoise/database/account/maintenance_tasks.rs
@@ -1,0 +1,113 @@
+//! Per-account repository for one-time maintenance task markers.
+
+use std::sync::Arc;
+
+use crate::whitenoise::database::DatabaseError;
+use crate::whitenoise::database::account_db::AccountDatabase;
+use crate::whitenoise::error::Result;
+
+/// Repository for account-scoped maintenance task completion markers.
+#[derive(Clone, Debug)]
+pub struct MaintenanceTasksRepo {
+    db: Arc<AccountDatabase>,
+}
+
+impl MaintenanceTasksRepo {
+    pub(crate) fn new(db: Arc<AccountDatabase>) -> Self {
+        Self { db }
+    }
+
+    /// Returns true once `name` has completed successfully for this account.
+    pub async fn is_completed(&self, name: &str) -> Result<bool> {
+        let completed: i64 = sqlx::query_scalar(
+            "SELECT EXISTS(
+                SELECT 1 FROM account_maintenance_tasks WHERE name = ?
+             )",
+        )
+        .bind(name)
+        .fetch_one(&self.db.inner.pool)
+        .await
+        .map_err(DatabaseError::Sqlx)?;
+
+        Ok(completed != 0)
+    }
+
+    /// Mark `name` complete for this account.
+    ///
+    /// This is intentionally an upsert so retrying the same successful
+    /// maintenance task remains idempotent.
+    pub async fn mark_completed(&self, name: &str) -> Result<()> {
+        sqlx::query(
+            "INSERT INTO account_maintenance_tasks (name, completed_at)
+             VALUES (?, unixepoch())
+             ON CONFLICT(name) DO UPDATE SET completed_at = excluded.completed_at",
+        )
+        .bind(name)
+        .execute(&self.db.inner.pool)
+        .await
+        .map_err(DatabaseError::Sqlx)?;
+
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use std::sync::Arc;
+
+    use nostr_sdk::Keys;
+    use tempfile::TempDir;
+
+    use super::*;
+
+    async fn setup() -> (MaintenanceTasksRepo, TempDir) {
+        let dir = TempDir::new().unwrap();
+        let pubkey = Keys::generate().public_key();
+        let db = Arc::new(
+            AccountDatabase::new(pubkey, dir.path().join("acct.db"))
+                .await
+                .unwrap(),
+        );
+
+        sqlx::query("DROP TABLE IF EXISTS account_maintenance_tasks")
+            .execute(&db.inner.pool)
+            .await
+            .unwrap();
+        sqlx::query(
+            "CREATE TABLE account_maintenance_tasks (
+                name TEXT PRIMARY KEY,
+                completed_at INTEGER NOT NULL
+            )",
+        )
+        .execute(&db.inner.pool)
+        .await
+        .unwrap();
+
+        (MaintenanceTasksRepo::new(db), dir)
+    }
+
+    #[tokio::test]
+    async fn task_is_incomplete_until_marked() {
+        let (repo, _dir) = setup().await;
+
+        assert!(!repo.is_completed("cleanup").await.unwrap());
+        repo.mark_completed("cleanup").await.unwrap();
+        assert!(repo.is_completed("cleanup").await.unwrap());
+    }
+
+    #[tokio::test]
+    async fn mark_completed_is_idempotent() {
+        let (repo, _dir) = setup().await;
+
+        repo.mark_completed("cleanup").await.unwrap();
+        repo.mark_completed("cleanup").await.unwrap();
+
+        let completed_count: i64 = sqlx::query_scalar(
+            "SELECT COUNT(*) FROM account_maintenance_tasks WHERE name = 'cleanup'",
+        )
+        .fetch_one(&repo.db.inner.pool)
+        .await
+        .unwrap();
+        assert_eq!(completed_count, 1);
+    }
+}
--- a/src/whitenoise/database/account/maintenance_tasks.rs
+++ b/src/whitenoise/database/account/maintenance_tasks.rs
@@ -69,6 +69,10 @@ mod tests {
                 .unwrap(),
         );
 
+        // Matches the project-wide account-DB test pattern: stamp the schema
+        // directly because `AccountDatabase::new` uses `open_without_migrations`.
+        // Keep this definition in sync with `fresh_account_schema.sql` until a
+        // shared `setup_account_db_with_migrations` helper exists.
         sqlx::query("DROP TABLE IF EXISTS account_maintenance_tasks")
             .execute(&db.inner.pool)
             .await
--- a/src/whitenoise/database/rust_migrations/fresh_account_schema.sql
+++ b/src/whitenoise/database/rust_migrations/fresh_account_schema.sql
@@ -49,6 +49,11 @@ CREATE INDEX IF NOT EXISTS idx_published_kp_event_id
 CREATE INDEX IF NOT EXISTS idx_published_kp_cleanup
     ON published_key_packages(consumed_at, key_material_deleted);
 
+CREATE TABLE IF NOT EXISTS account_maintenance_tasks (
+    name          TEXT PRIMARY KEY,
+    completed_at  INTEGER NOT NULL
+);
+
 CREATE TABLE IF NOT EXISTS published_events (
     id          INTEGER PRIMARY KEY AUTOINCREMENT,
     event_id    TEXT NOT NULL UNIQUE
--- a/src/whitenoise/database/rust_migrations/fresh_account_schema.sql
+++ b/src/whitenoise/database/rust_migrations/fresh_account_schema.sql
@@ -50,7 +50,8 @@ CREATE INDEX IF NOT EXISTS idx_published_kp_cleanup
     ON published_key_packages(consumed_at, key_material_deleted);
 
 CREATE TABLE IF NOT EXISTS account_maintenance_tasks (
-    name          TEXT PRIMARY KEY,
+    id            INTEGER PRIMARY KEY AUTOINCREMENT,
+    name          TEXT NOT NULL UNIQUE,
     completed_at  INTEGER NOT NULL
 );
 
--- a/src/whitenoise/foreground_catchup.rs
+++ b/src/whitenoise/foreground_catchup.rs
@@ -1,12 +1,8 @@
 use std::collections::HashSet;
-#[cfg(not(test))]
 use std::time::Duration;
 
 use crate::perf_instrument;
-#[cfg(not(test))]
-use crate::whitenoise::session::key_packages::{
-    KEY_PACKAGE_RELAY_CLEANUP_INVITE_GRACE_SECS, KeyPackageRelayCleanupOutcome,
-};
+use crate::whitenoise::session::key_packages::KeyPackageRelayCleanupOutcome;
 use crate::whitenoise::{
     Whitenoise,
     accounts::Account,
@@ -15,9 +11,11 @@ use crate::whitenoise::{
 };
 
 #[cfg(not(test))]
-const KEY_PACKAGE_RELAY_CLEANUP_DELAY: Duration =
-    Duration::from_secs(KEY_PACKAGE_RELAY_CLEANUP_INVITE_GRACE_SECS);
-#[cfg(not(test))]
+const KEY_PACKAGE_RELAY_CLEANUP_DELAY: Duration = Duration::from_secs(
+    crate::whitenoise::session::key_packages::KEY_PACKAGE_RELAY_CLEANUP_INVITE_GRACE_SECS,
+);
+#[cfg(test)]
+const KEY_PACKAGE_RELAY_CLEANUP_DELAY: Duration = Duration::from_millis(50);
 const KEY_PACKAGE_RELAY_CLEANUP_MAX_ATTEMPTS: usize = 3;
 
 impl Whitenoise {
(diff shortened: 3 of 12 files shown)


erskingardner:
@claude review
@coderabbitai full review

src/whitenoise/session/key_packages.rs
```diff
@@ -243,12 +263,159 @@ impl<'a> KeyPackageOps<'a> {
...
+                canonical,
+                legacy,
+                final_events.len()
+            )));
+        }
+
+        self.session
+            .repos
+            .maintenance_tasks
+            .mark_completed(KEY_PACKAGE_RELAY_CLEANUP_TASK)
```
mubarakcoded, line 403: Before marking this one-time cleanup completed, can we make the completion condition stricter? Two cases seem worth guarding: if any cleanup relay was unreachable, we may have only cleaned a subset, and if a Welcome marked a key package consumed while this relay work was running, the queued rotation can publish another pair afterward. In either case, marking completed here means we won't retry later.

ghost:
Ready to review this PR? Stage has broken it down into 4 individual chapters for you:

| | Title |
|---|-------|
| 1 | [Define maintenance task schema and migrations](https://stagereview.app/marmot-protocol/whitenoise-rs/pull/845/chapters/1) |
| 2 | [Implement maintenance task and key package repositories](https://stagereview.app/marmot-protocol/whitenoise-rs/pull/845/chapters/2) |
| 3 | [Add key package relay cleanup logic](https://stagereview.app/marmot-protocol/whitenoise-rs/pull/845/chapters/3) |
| 4 | [Schedule cleanup on startup and resume](https://stagereview.app/marmot-protocol/whitenoise-rs/pull/845/chapters/4) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-rs/pull/845">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 930b207a3bd2e4a2a1cbc6c80b7384ca0550d481 on May 21, 2026 8:08am UTC.</sub>

src/whitenoise/session/key_packages.rs
erskingardner: my agent tried to pull this one me too. i don't think we can reliably guarantee this because people have dead relays in their list. those people would never ever get this migration done as best we can. I think this will work just fine for most relays and most users so I'm happier to have the mostly happy path rather than try and optimise for an edge case.

mubarakcoded: I agree, we shouldn't block completion forever on dead relays.

src/whitenoise/database/rust_migrations/fresh_account_schema.sql
```diff
@@ -49,6 +49,11 @@ CREATE INDEX IF NOT EXISTS idx_published_kp_event_id
 CREATE INDEX IF NOT EXISTS idx_published_kp_cleanup
     ON published_key_packages(consumed_at, key_material_deleted);
 
+CREATE TABLE IF NOT EXISTS account_maintenance_tasks (
+    name          TEXT PRIMARY KEY,
+    completed_at  INTEGER NOT NULL
+);
```
jgmontoya, line 56: I think we should keep using id integer with autoincrement as primary key

src/whitenoise/session/key_packages.rs
```diff
@@ -243,12 +263,159 @@ impl<'a> KeyPackageOps<'a> {
...
+        }
+
+        let publish_relays = self.prepare_relays().await?;
+        if publish_relays.is_empty() {
+            return Err(WhitenoiseError::AccountMissingKeyPackageRelays);
+        }
+        let cleanup_relay_urls = self.cleanup_relay_urls().await?;
+        if cleanup_relay_urls.is_empty() {
+            return Err(WhitenoiseError::AccountMissingKeyPackageRelays);
+        }
```
jgmontoya: I don't think `cleanup_relay_urls` can ever be empty given the previous guard making sure that publish_relays is not empty.

src/whitenoise/session/key_packages.rs
```diff
@@ -584,6 +774,38 @@ impl<'a> KeyPackageOps<'a> {
...
     }
 
+    async fn cleanup_relay_urls(&self) -> Result<Vec<RelayUrl>> {
+        let user =
+            User::find_by_pubkey(&self.session.account_pubkey, &self.session.shared.database)
+                .await
+                .map_err(|_| WhitenoiseError::AccountNotFound)?;
+        let key_package_relays = user
+            .relays(RelayType::KeyPackage, &self.session.shared.database)
+            .await?;
```
jgmontoya, line 840: We had already queried for these things (on the `prepare_relays` call) and we're doing it again here. I think we can structure this in a more efficient way.

src/whitenoise/session/key_packages.rs
```diff
@@ -243,12 +263,159 @@ impl<'a> KeyPackageOps<'a> {
...
+
+        if self
+            .session
+            .repos
+            .maintenance_tasks
+            .is_completed(KEY_PACKAGE_RELAY_CLEANUP_TASK)
+            .await?
+        {
+            return Ok(KeyPackageRelayCleanupOutcome::AlreadyCompleted);
+        }
```
jgmontoya, line 311: Can this actually happen given that we already have a guard in lines 279-284?

src/whitenoise/session/key_packages.rs
```diff
@@ -722,12 +944,73 @@ impl<'a> KeyPackageOps<'a> {
...
 
+fn push_unique_relay_url(
+    relay_urls: &mut Vec<RelayUrl>,
+    seen: &mut HashSet<String>,
+    relay_url: RelayUrl,
+) {
+    if seen.insert(relay_url.as_str().to_owned()) {
+        relay_urls.push(relay_url);
+    }
+}
```
jgmontoya: This method seems... useless tbh

src/whitenoise/session/key_packages.rs
```diff
@@ -722,12 +944,73 @@ impl<'a> KeyPackageOps<'a> {
...
+    }
+    for relay_url in nip65_relays {
+        push_unique_relay_url(&mut relay_urls, &mut seen, relay_url);
+    }
+    for relay_url in default_relays {
+        push_unique_relay_url(&mut relay_urls, &mut seen, relay_url);
+    }
+
+    relay_urls
+}
```
jgmontoya, line 1022: I don't understand why we're doing this in what seems like a convoluted way. I think we should just push all to HashSet and then convert to Vec. I don't see any reason why the particular order (kp, then nip65, then default) would matter for this.

If we wanted to keep the order anyway, I think this could be a cleaner way:

```rs
  let mut seen = HashSet::new();
  key_package_relays.into_iter()
      .chain(nip65_relays)
      .chain(default_relays)
      .filter(|url| seen.insert(url.clone()))
      .collect()
```

src/whitenoise/session/key_packages.rs
```diff
@@ -722,12 +944,73 @@ impl<'a> KeyPackageOps<'a> {
...
+    let canonical = events
+        .iter()
+        .filter(|event| event.kind == MLS_KEY_PACKAGE_KIND)
+        .count();
+    let legacy = events
+        .iter()
+        .filter(|event| event.kind == MLS_KEY_PACKAGE_KIND_LEGACY)
+        .count();
+    (canonical, legacy)
+}
```
jgmontoya: This can be done in a single iteration instead of going twice over all the events:

```rs
fn key_package_pair_counts(events: &[Event]) -> (usize, usize) {
    let mut canonical = 0;
    let mut legacy = 0;

    for event in events {
        match event.kind {
            MLS_KEY_PACKAGE_KIND => canonical += 1,
            MLS_KEY_PACKAGE_KIND_LEGACY => legacy += 1,
            _ => {}
        }
    }

    (canonical, legacy)
}
```

src/whitenoise/session/key_packages.rs
```diff
@@ -243,12 +263,159 @@ impl<'a> KeyPackageOps<'a> {
...
+    async fn fetch_all_from_reachable_cleanup_relays(
+        &self,
+        relay_urls: &[RelayUrl],
+    ) -> Result<Vec<Event>> {
+        // This naming wrapper keeps cleanup call sites explicit. Partial-connect
+        // handling lives inside the ephemeral fetch path; cleanup should operate
+        // on relays that are reachable now without one disconnected relay
+        // blocking the rest.
+        self.fetch_all_from_relay_urls(relay_urls).await
+    }
```
jgmontoya: I would remove this method/wrapper. I don't think it's actually helping in any way as it still depends on being handed the proper relays to use, the responsibility is on the caller.

src/whitenoise/session/key_packages.rs
```diff
@@ -26,6 +26,26 @@ const MAX_PUBLISH_ATTEMPTS: u32 = 3;
...
+/// One-time per-account marker for the release cleanup that purges legacy
+/// relay-side key packages and republishes one fresh pair.
+pub(crate) const KEY_PACKAGE_RELAY_CLEANUP_TASK: &str = "key_package_relay_cleanup_v1";
+
+/// Quiet period used before the one-time relay cleanup runs after foreground catch-up.
+pub(crate) const KEY_PACKAGE_RELAY_CLEANUP_INVITE_GRACE_SECS: u64 = 30;
+
+/// Quiet period after a key package is consumed before relay cleanup may start.
+pub(crate) const KEY_PACKAGE_RELAY_CLEANUP_QUIET_PERIOD_SECS: u64 =
+    KEY_PACKAGE_RELAY_CLEANUP_INVITE_GRACE_SECS;
```
jgmontoya: Do we need two constants? Is there a reason these might diverge? If so I think we should just set the value explicitly (30) instead of implying they're always tight together.

src/whitenoise/session/key_packages.rs
```diff
@@ -406,15 +573,38 @@ impl<'a> KeyPackageOps<'a> {
...
+        relay_urls: &[RelayUrl],
+    ) -> Result<usize> {
+        if key_package_events.is_empty() {
+            return Ok(0);
+        }
+        if relay_urls.is_empty() {
+            return Err(WhitenoiseError::AccountMissingKeyPackageRelays);
+        }
+
+        let original_count = key_package_events.len();
```
jgmontoya: I don't think we need this, we can just use `original_ids.len()`

src/whitenoise/session/key_packages.rs
```diff
@@ -722,12 +944,73 @@ impl<'a> KeyPackageOps<'a> {
...
+        .iter()
+        .filter(|event| event.kind == MLS_KEY_PACKAGE_KIND_LEGACY)
+        .count();
+    (canonical, legacy)
+}
+
+fn has_exact_key_package_pair(events: &[Event]) -> bool {
+    let (canonical, legacy) = key_package_pair_counts(events);
+    canonical == 1 && legacy == 1 && events.len() == 2
+}
```
jgmontoya: Just a mental note: this will have to change after we drop the legacy key packages and also when we do multi-device.

src/whitenoise/foreground_catchup.rs
```diff
@@ -45,9 +55,113 @@ impl Whitenoise {
...
         self.replay_chat_list_stream_snapshots(&accounts).await;
+        self.schedule_key_package_relay_cleanup_after_grace("foreground")
+            .await;
         Ok(())
     }
 
+    pub(crate) async fn schedule_key_package_relay_cleanup_after_grace(
+        &self,
+        trigger: &'static str,
+    ) {
```
jgmontoya: Why do we need the trigger parameter? I'm not seeing how it's useful.

src/whitenoise/foreground_catchup.rs
```diff
@@ -45,9 +55,113 @@ impl Whitenoise {
...
+                        tracing::warn!(
+                            target: "whitenoise::foreground_catchup",
+                            account = %session.account_pubkey.to_hex(),
+                            trigger,
+                            "One-time key package relay cleanup failed: {}",
+                            error
+                        );
+                    }
+                }
+            }
```
jgmontoya: I think this could be extracted to a method, something like

```rs
async fn cleanup_sweep_once(&self, attempt: usize) -> bool {
  let mut any_deferred = false;
  for session in self.account_manager.sessions_iter() {
    match session.key_packages().cleanup_relay_key_packages_once().await {
      Ok(AlreadyCompleted) => { /* debug log */ }
      Ok(DeferredRecentConsumedKeyPackage) => { any_deferred = true; /* info log */ }
      Ok(DeletedAndPublished { deleted }) => { /* info log */ }
      Err(error) => { /* warn log */ }
    }
  }
  any_deferred
}
```

And then this becomes:

```rs
  for attempt in 1..=KEY_PACKAGE_RELAY_CLEANUP_MAX_ATTEMPTS {
      if self.wait_through_grace_or_shutdown(&mut shutdown_rx).await.is_break() {
          return;
      }
      if !self.cleanup_sweep_once(attempt).await {
          return;
      }
  }
  tracing::warn!(/* remained deferred after all attempts */);
```

jgmontoya (commented):
Leaving some comments/questions/suggestions

src/whitenoise/session/key_packages.rs
```diff
@@ -243,12 +263,159 @@ impl<'a> KeyPackageOps<'a> {
...
+            return Err(WhitenoiseError::KeyPackagePublishFailed(format!(
+                "one-time key package cleanup verification expected exactly one kind:{} and \
+                 one kind:{} event, found {} canonical, {} legacy, {} total",
+                MLS_KEY_PACKAGE_KIND.as_u16(),
+                MLS_KEY_PACKAGE_KIND_LEGACY.as_u16(),
+                canonical,
+                legacy,
+                final_events.len()
+            )));
+        }
```
jgmontoya: From my agent:


> ### 1. Strict final verification can cause permanent delete/republish thrash
> 
> `src/whitenoise/session/key_packages.rs:371-385` — after publishing the fresh
> pair, the job sleeps 500 ms and then requires `has_exact_key_package_pair`:
> *exactly* one canonical, one legacy, total 2. If that check fails the job
> returns `Err`, `mark_completed` is never called, and the next foreground/startup
> re-runs the **whole** cleanup — deleting the pair it just published and
> republishing.
> 
> The verification fetch only sees *reachable* relays (`fetch_all_from_reachable_cleanup_relays`),
> and 500 ms is optimistic for relay round-trip + indexing on a slow or
> eventually-consistent relay. Two realistic ways this thrashes forever:
> 
> - The new pair simply isn't visible within 500 ms → `total != 2` → fail.
> - A relay that was unreachable during the delete phase (so its stale events
>   weren't seen by the post-delete `remaining` check at line 348-357) becomes
>   reachable during verification → `total > 2` → fail forever.
> 
> Every thrash also re-opens the zero-key-package window (see #2). The asymmetry
> matters: a false *failure* costs a permanent delete/republish loop across app
> sessions; a false *success* only leaves the account with extra key packages —
> the benign condition this job exists to clean up.
> 
> Suggestion: have `publish_pair` return the two `EventId`s it published, and make
> verification assert *those two are present* rather than *nothing else exists*.
> That is the invariant you actually care about and it cannot be tripped by relay
> lag or a resurfacing stale relay. Alternatively, downgrade a verification
> mismatch from `Err` to a `warn!` and still `mark_completed`, since delete and
> publish both already succeeded by that point.

src/whitenoise/session/key_packages.rs
```diff
@@ -243,12 +263,159 @@ impl<'a> KeyPackageOps<'a> {
...
+                MAX_DELETE_ROUNDS
+            )));
+        }
+
+        let (key_package_data, canonical_created_at) = self
+            .prepare_canonical_publish_inputs(&publish_relays)
+            .await?;
+        let publish_relay_urls = Relay::urls(&publish_relays);
+        self.publish_pair(&key_package_data, canonical_created_at, &publish_relay_urls)
+            .await?;
```
jgmontoya: From my agent:

> ### 2. Delete-then-publish leaves a window with zero key packages on relays
> 
> `src/whitenoise/session/key_packages.rs:321-364` — the job deletes every
> relay-side key package event, runs the post-delete emptiness check, sleeps
> 500 ms, and only then publishes the fresh pair. For the several seconds this
> spans, the account has **no** key package on any relay, so any peer trying to
> invite it to a group during that window fails to find one.
> 
> The 30 s invite grace and the consumed-key-package deferral reduce the odds,
> but the window is real and re-opens on every thrash from #1.
> 
> Consider inverting the order: publish the fresh pair *first*, then delete every
> key package event *except* the two just-published IDs. The account is never
> without a key package. This is more code (you must thread the new event IDs
> into the delete filter, and the NIP-33 d-tag reuse already replaces the
> same-slot canonical for free), so weigh it against the fact that this is a
> one-time job behind a 30 s grace. But for a security messenger, "briefly
> un-invitable" is worth a deliberate decision rather than an accident of
> ordering.

src/whitenoise/session/key_packages.rs
```diff
@@ -243,12 +263,159 @@ impl<'a> KeyPackageOps<'a> {
...
+        self.session
+            .repos
+            .maintenance_tasks
+            .mark_completed(KEY_PACKAGE_RELAY_CLEANUP_TASK)
+            .await?;
+
+        Ok(KeyPackageRelayCleanupOutcome::DeletedAndPublished {
+            deleted: total_deleted,
+        })
+    }
```
jgmontoya, line 409: From my agent:

> ### 3. Cleanup permanently orphans local MLS key material for stale key packages
> 
> `src/whitenoise/session/key_packages.rs:269-396` — the job deletes relay events
> with `delete_mls_stored_keys = false` and never touches the
> `published_key_packages` DB rows. The doc comment at lines 271-274 correctly
> justifies keeping local material (a not-yet-processed Welcome may still need
> it), but it does not acknowledge the consequence: for genuinely stale key
> packages that were *never consumed and never will be* (their relay event is now
> gone, so no new inviter can reach them), the local MLS material and the DB row
> leak forever. `find_eligible_for_cleanup` only reclaims **consumed** packages,
> so nothing ever collects these.
> 
> The leak is small and bounded (one-time, proportional to how many extra key
> packages the account accumulated), so this is not a blocker — but it should be
> a *deliberate* decision. Either extend the doc comment to state the leak is
> accepted, or file a follow-up to reclaim the orphans. Silent leaks age badly.

src/whitenoise/database/account/maintenance_tasks.rs
```diff
@@ -0,0 +1,117 @@
...
+    /// Mark `name` complete for this account.
+    ///
+    /// This is intentionally an upsert so retrying the same successful
+    /// maintenance task remains idempotent.
+    pub async fn mark_completed(&self, name: &str) -> Result<()> {
+        sqlx::query(
+            "INSERT INTO account_maintenance_tasks (name, completed_at)
+             VALUES (?, unixepoch())
+             ON CONFLICT(name) DO UPDATE SET completed_at = excluded.completed_at",
+        )
```
jgmontoya, line 41: From my agent:

> ### 6. `mark_completed` upsert writes a column nobody reads
> 
> `src/whitenoise/database/account/maintenance_tasks.rs:41-57` — `mark_completed`
> uses `ON CONFLICT(name) DO UPDATE SET completed_at = excluded.completed_at`, and
> `is_completed` only checks row existence (`SELECT EXISTS`). Nothing in the
> codebase ever reads `completed_at`. So the `DO UPDATE` is dead logic and, worse,
> it overwrites the original completion time with the *last retry* time — the less
> useful value.
> 
> If `completed_at` is kept for debugging/observability, use
> `ON CONFLICT DO NOTHING` so it records *first* completion. If it is truly
> unused, `INSERT OR IGNORE` (or drop the column) is simpler. Storing a timestamp
> no reader consumes is speculative — prefer the simpler form.

src/whitenoise/session/key_packages.rs
```diff
@@ -243,12 +263,159 @@ impl<'a> KeyPackageOps<'a> {
...
+            let deleted = self
+                .delete_batch_from_relay_urls(key_package_events, false, 1, &cleanup_relay_urls)
+                .await?;
+            total_deleted += deleted;
+
+            if deleted == 0 {
+                let remaining = self
+                    .fetch_all_from_reachable_cleanup_relays(&cleanup_relay_urls)
+                    .await?;
+                return Err(WhitenoiseError::KeyPackagePublishFailed(format!(
```
jgmontoya: From my agent:

> ### 7. `KeyPackagePublishFailed` used for deletion failures
> 
> `src/whitenoise/session/key_packages.rs:339` and `:352` — both error sites
> report a *deletion* failure ("could not delete relay event(s)", "left N relay
> event(s) after delete rounds") but wrap it in `WhitenoiseError::KeyPackagePublishFailed`.
> The message text is clear, but the variant misleads anyone matching on error
> types or scanning logs. There is no delete-specific variant today; consider
> adding one, or at least leave a comment noting the deliberate reuse.
>

src/whitenoise/session/key_packages.rs
```diff
@@ -807,4 +1090,363 @@ mod tests {
...
+        );
+        assert_eq!(KEY_PACKAGE_RELAY_CLEANUP_INVITE_GRACE_SECS, 30);
+        assert_eq!(KEY_PACKAGE_RELAY_CLEANUP_QUIET_PERIOD_SECS, 30);
+        assert_eq!(KEY_PACKAGE_RELAY_CLEANUP_VERIFY_DELAY_MS, 500);
+        let outcomes = [
+            KeyPackageRelayCleanupOutcome::AlreadyCompleted,
+            KeyPackageRelayCleanupOutcome::DeferredRecentConsumedKeyPackage,
+            KeyPackageRelayCleanupOutcome::DeletedAndPublished { deleted: 2 },
+        ];
+        assert_eq!(outcomes.len(), 3);
```
jgmontoya, line 1444: What's this testing exactly?

jgmontoya (commented):
Some additional comments, mostly from my agent

jgmontoya (changes requested):

src/whitenoise/session/key_packages.rs
erskingardner: Yes it can. This handles the edge case of another task completing while this task waits for the publish lock.

src/whitenoise/session/key_packages.rs
erskingardner: I think this is ok - it's a very small amount of data and only helps in the case of a user receiving an invite using one of the older key packages.

src/whitenoise/foreground_catchup.rs
erskingardner: it's a minor thing for logs to distinguish between start up and foreground. I don't care though. we can remove.

mubarakcoded (approved):
LGTM!

dannym-arx (approved):

